PR #13, opened 2021-11-09
chore(build): replace esbuild with rollup

YonatanKra:


rollup.config.ts
```diff
@@ -0,0 +1,35 @@
...
+        });
+    return testFolders;
+}
+
+const components = getFoldersInAFolder('./src/components');
+const input = components.reduce((inputObject, componentName) => {
+    inputObject[`components/${componentName}/${componentName}`] = path.join(process.cwd(), `src/components/${componentName}/${componentName}.ts`);
+    return inputObject;
+}, {});
+
```
yinonov, line 26: should we?
```suggestion

/**
 * @type {import('rollup').RollupOptions}
 */
```

yinonov (changes requested):
should rollup.config.js be .git ignored?

rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+var components = getFoldersInAFolder('./src/components');
+console.log(components);
+// export default {
+//     input: 'src/index.js',
+//     output: {
+//         dir: 'output',
+//         format: 'cjs'
+//     },
+//     plugins: [typescript()]
+// };
```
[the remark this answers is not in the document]
YonatanKra: Removed :)

rollup.config.ts
YonatanKra: What for?

package.json
```diff
@@ -7,7 +7,7 @@
     "prestart": "tsc -b",
     "start": "tsc --watch --preserveWatchOutput & web-dev-server --watch",
     "build:typescript": "tsc -b",
-    "build": "node ./scripts/build.js",
+    "build": "rollup -c",
```
yinonov, line 10: I kind of like the separation of scripts to files

rollup.config.ts
yinonov: https://rollupjs.org/guide/en/#config-intellisense
maybe not needed in the ts case?

YonatanKra: Not needed. That's why I'm using TS anywhere possible.

package.json
YonatanKra: In the future, when we will have a good reason to do so, and the capacity to maintain scripts, I am all for it. This is why I postponed with Nx cause building the `generators` and `executors` is a big task.

rollup.config.ts
YonatanKra: This file is no longer in the repo BTW.

yinonov (approved):

yinonov (approved):

PR #17, opened 2021-11-09
chore(workspace): tsconfig default hints

yinonov:


YonatanKra:
Our `tsconfig` was bloated as it is. Can you plz remove all the comments so we'll be able to compare oranges to oranges?
And what is the goal of this change?

yinonov:
> Our `tsconfig` was bloated as it is. Can you plz remove all the comments so we'll be able to compare oranges to oranges? And what is the goal of this change?

I initialized the default tsconfig to keep the clear source descriptions for each property.
this could help better understand configuration in real time

YonatanKra (changes requested):
I suggest waiting until we have our build system in place and only then tweak this one with such a massive change. Otherwise, we'll be chasing our tail...

PR #25, opened 2021-11-11
build(workspace): add scss to lit css plugin

yinonov:


YonatanKra (commented):
How do I get only css and not li?

yinonov:
> How do I get only css and not lit-css?

Not sure yet. Better separate to a different task

YonatanKra (approved):

PR #28, opened 2021-11-12
build(theme): render styles script added

yinonov:


YonatanKra:
Use rollup to build the themes styles both for tests and for production

PR #32, opened 2021-11-14
chore(CI/CD): add test to pull request

rinaok:


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,70 @@
...
+            - browser: SafariNative
+              browser_name: safari
+              report_coverage: false
+      steps:
+        - uses: actions/checkout@v2
+        - uses: actions/setup-node@v2
+          with:
+            node-version: '16'
+            cache: 'npm'
+        - run: npm install
```
YonatanKra, line 58: I think we can use cache here for `npm install`. What do you think?

rinaok: Would it  be better to change it to `npm ci`?

PR #68, opened 2021-12-02
chore(ui-tests): extract test use cases from readme

YonatanKra:


package.json
```diff
@@ -75,6 +75,7 @@
     "eslint-plugin-lit": "^1.6.1",
     "eslint-plugin-lit-a11y": "^2.2.0",
     "eslint-plugin-wc": "^1.3.2",
+    "extract-gfm": "^0.1.0",
```
yinonov, line 78: not sure if explored already but this lib's last publish is 7 years old.
this seem more recent https://github.com/earldouglas/codedown
(your call)

src/components/badge/ui.test.ts
```diff
@@ -34,17 +36,11 @@ async function loadTemplate({
...
-		'cta',
-		'success',
-		'alert',
-		'warning',
-		'info',
-		'announcement',
-	];
-
-	const template = connotations.reduce((htmlString, connotationValue) => `${htmlString}<div style="margin: 5px"><vwc-badge text="Badge" connotation="${connotationValue}"></vwc-badge></div>`, '');
+	const template = extractHTMLBlocksFromReadme(path.join(__dirname, 'README.md')).reduce((accumulator: string, block: string) => {
```
yinonov, line 39: funny that the eslint threw on reduce's acc

yinonov: filters to only fetch html snippets?

yinonov (changes requested):

src/components/badge/ui.test.ts
YonatanKra: It's more funny since I put a rule to overule this throw... didn't give it too much effort though. Maybe you can come up with something (look at the eslint change).
I take only `html`.  Do you have a use case for `JS`? That's a bit more complex I think and not cost effective...

yinonov: you could reduce over a nodeList which is more explicit.
there's [this](https://stackoverflow.com/a/66902167/3027703) solution but it's too TAFUR for the specific use-case.
alternatively just do
```js
const template = extractHTMLBlocksFromReadme(path.join(__dirname, 'README.md'))
		.reduce((accumulator: string, block: string) => `${accumulator} <div style="margin: 5px;">${block}</div>`, '');
```

yinonov (changes requested):

package.json
YonatanKra: `codedown` doesn't have exposed API.  I could make a PR on their repository (or fork theirs and take over it).

.eslintrc
```diff
@@ -58,6 +58,13 @@
     "no-var": "error",
     "no-floating-decimal": "error",
     "prefer-const": "error",
+    "no-param-reassign": [
+      "error",
+      {
+        "props": true,
+        "ignorePropertyModificationsFor": ["accumulator"]
+      }
+    ],
```
yinonov, line 67: no longer needed right?

yinonov (changes requested):

.eslintrc
YonatanKra: It actually didn't work, so can remove

yinonov (approved):

PR #71, opened 2021-12-04
build(typescript): turn on isolated modules flag

yinonov:
explainer
https://www.typescriptlang.org/tsconfig#isolatedModules


YonatanKra (approved):
I approve, but the `lint-pr` workflow is stuck.

PR #97, opened 2021-12-31
ci(workspace): add husky hook

yinonov:
optimize CI by
- [ ] PR lint diff files against target branch
- [ ] pre-commit lint staged

is full lint (all files) ever required?

I agree with the following
> This “we shall lint everything” concept is hardcore for developers, but not really agile for business. As my scrum master once told me: it is very likely this section will be refactor or removed before we really need to maintain it; we may end up waste time to correct the lint errors in this section for nothing. This is not align with business value.

https://seeliang.medium.com/how-to-update-lint-without-lint-the-whole-project-841150aa59b0#:~:text=This%20%E2%80%9Cwe%20shall,with%20business%20value.

Pit falls -
- a scenario where lint rules are updated will not catch unchanged files that don't meet the new code standard
- following this scenario can lead to authors touch files and get lint errors thrown on, as far as they know, a **pristine code**


yinonov:
following a discussion on [node-tooling](https://devtoolscommunity.herokuapp.com/) here's a [great answer](https://node-tooling.slack.com/archives/C09PUBXB8/p1641188267064800)

> lint checks can NOT just run on changed files; eslint-plugin-import checks imports across files, and tsc is also a “lint check” that runs across files
> 
> iow, the ONLY way you can safely run on a subset is if you have 100% certainty of the dependency graph, and you test all changed things, PLUS all things that depend on them
> 
> and since that’s very difficult to know without applying a lot of constraints to the codebase, and since the cost of running checks on the entire codebase is minimal (for example, for eslint, use jest-runner-eslint - airbnb dropped lint time from 22 minutes to 2 by adding that) it’s just not worth the complexity
> 
> focusing on ideological wastefulness is actively harmful. focus on actual problems instead.

yinonov:
@YonatanKra familiar with _jest-runner-eslint_?

yinonov:
Husky fails to install on docker image with Node 16. see nodejs/docker-node#1598

PR #104, opened 2022-01-04
docs(icon): update icon

rinaok:


src/components/icon/icon.template.ts
```diff
@@ -8,7 +8,7 @@ import type { Icon } from './icon.base';
 const getClasses = ({
 	connotation, size,
 }: Icon) => classNames(
-	'control',
+	'icon',
```
yinonov, line 11: control is the base component selector convention in fast. can we consider adopting it?

yinonov (commented):
@rinaok mind restoring connotation? this component was rebuilt from scratch

PR #107, opened 2022-01-06
chore(button): migrate to vivid next

yinonov:
BREAKING CHANGE: removed connotations - announcement, info

BREAKING CHANGE: single affix icon support only

- scaffolding
- tests
- if no label but has icon should set class to apply icon-only styles
  - state applies class
  - add style according to class
- disabled style


src/components/button/README.md
```diff
@@ -0,0 +1,110 @@
...
+### Outlined button with connotation
+
+```html preview
+<vwc-button appearance="outlined" label='primary' connotation='primary'></vwc-button>
+<vwc-button appearance="outlined" label='cta' connotation='cta'></vwc-button>
+<vwc-button appearance="outlined" label='success' connotation='success'></vwc-button>
+<vwc-button appearance="outlined" label='alert' connotation='alert'></vwc-button>
+<vwc-button appearance="outlined" label='announcement' connotation='announcement'></vwc-button>
+<vwc-button appearance="outlined" label='info' connotation='info'></vwc-button>
+```
```
rachelbt, line 110:
```suggestion
```## States
### Disabled
```html preview
<vwc-button label='ghost' appearance='ghost' disabled></vwc-button>
<vwc-button label='filled' appearance='filled' disabled></vwc-button>
<vwc-button label='outlined' appearance='outlined' disabled></vwc-button>
```
```

File: src/components/button/button.template.ts [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+
+const getClasses = ({
+	connotation, appearance, shape, size, iconTrailing,
+}: Button) => classNames(
+	'control',
+	['icon-trailing', iconTrailing],
+	[`connotation-${connotation}`, Boolean(connotation)],
+	[`appearance-${appearance}`, Boolean(appearance)],
+	[`shape-${shape}`, Boolean(shape)],
+	[`size-${size}`, Boolean(size)],
```
rachelbt, line 15: disabled class is needed and style as well :)

yinonov:
@jshenkman should we support loading "state"?

yinonov:
@jshenkman are `tonal` buttons in consideration? should they replace `ghost`? maybe `tonal` only apply in **toggle-buttons**?

yinonov:
tests fail when running both _badge_ and _button_.
tests pass when removing `badge.spec.ts`.
@YonatanKra anything u can think of?

yinonov:
@jshenkman FYI
meetings sign in button loading state

https://user-images.githubusercontent.com/15140652/149282886-2526ff80-1816-4f92-aa4d-df8f450e3d21.mov

src/components/button/button.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+	});
+
+	describe('size', () => {
+		it('sets correct internal size style', async () => {
+			const size = 'small';
+			(element as any).size = size;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot.querySelector(`.control.size-${size}`);
+			expect(control).toBeInstanceOf(Element);
```
YonatanKra, line 103: What does this test mean?

src/components/button/button.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+		});
+	});
+
+	describe('icon-only', () => {
+		it('sets correct internal icon-only style', async () => {
+			element.icon = 'home';
+			await elementUpdated(element);
+
+			const control = element.shadowRoot.querySelector('.control.icon-only');
```
YonatanKra, line 112: What's the other alternative? If I don't set an icon, what happens then?

src/components/button/button.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+type ButtonSize = Extract<Size, Size.BaseSmall | Size.Base | Size.BaseLarge>;
+
+/**
+ * Base class for button
+ *
+ * @public
+ */
+export class Button extends FoundationButton {
+	/**
```
YonatanKra, line 49: As a side note - these comments in the code are driving me nuts!

YonatanKra (changes requested):
See the comments

src/components/button/button.spec.ts
yinonov: it won't set `icon-only` class on `control`

src/components/button/button.ts
yinonov: JSDoc / TSDoc?

YonatanKra: I know - I don't like them :)
Not critical. Just leaving on the table :)

src/components/button/button.spec.ts
YonatanKra: Might need to check before and after.

YonatanKra:
```
const hasIconOnlyBefore = checkIconOnly(element);
element.icon = 'home';
const hasIconOnlyAfter = checkIconOnly(element);
expect(hasIconOnlyBefore).toEqual(false);
expect(hasIconOnlyAfter).toEqual(true);
```

src/components/button/button.spec.ts
yinonov: don't we prefer repeat the same structure?
won't it be incoherent to have unique naming for each test?
it's the same pattern

src/components/button/button.ts
yinonov: "this is all over the town"

yinonov:
icon figure element does not respond well to
```css
display: flex;
width: 1em;
height: 1em;
```
in safari, flex doesn’t set the flow to be blocked (in opposed to chrome / firefox)
it results in a “non-sized” element

if set to **block** instead of **flex**, the `<figure>` element is, again!, not vertically aligned to exact middle

image

YonatanKra (approved):

PR #108, opened 2022-01-06
refactor(interface): rename layout to appearance

yinonov:
BREAKING CHANGE: layout attribute is no longer supported. appearance should be used instead


src/core/foundation/enums.ts
```diff
@@ -21,7 +21,7 @@ export enum Shape {
 }
 
 /* eslint-disable no-shadow */
-export enum Layout {
+export enum Appearance {
```
YonatanKra, line 24: Didn't we say we are moving away from `enum`?

yinonov: yeah but it's not relevant to this scope

YonatanKra (approved):

PR #114, opened 2022-01-09
refactor(workspace): entry points changed to index

yinonov:


YonatanKra:
What is this change for?

yinonov:
> What is this change for?

I changed the folders structure
instead of
```js
src/components/button/button.ts
```
to
```js
src/components/button/index.ts
```

YonatanKra (approved):

PR #116, opened 2022-01-09
style(workspace): simplify eslint configuration

yinonov:
some plugins already apply on
 https://github.com/microsoft/fast/blob/5fce1b50f35898550c71659d13d3fa2c0bf1e075/packages/utilities/fast-eslint-rules/.eslintrc.js


.eslintrc
```diff
@@ -1,31 +1,18 @@
...
   },
-  "plugins": ["wc", "jsdoc"],
+  "plugins": ["@typescript-eslint", "import", "wc", "jsdoc"],
   "extends": [
-    "airbnb-base",
-    "airbnb-typescript/base",
-    "plugin:compat/recommended",
+    "@microsoft/eslint-config-fast-dna",
     "plugin:wc/recommended",
     "plugin:wc/best-practice",
```
YonatanKra, line 12: If we are taking `fast`'s eslint rules, do we really need these?

yinonov: fast doesn't include it.
it's kind of maintained, but I'm not sure it's needed that much

YonatanKra (commented):
Got a conflict with `main`

YonatanKra (approved):

PR #129, opened 2022-01-23
refactor(workspace): integrate nrwl nx

yinonov:
- [x] explicitly rename themes package to styles so that it’s clear it only stylelint and not eslint as well
  - sensible readable folder structure. if a folder is redundant - remove
    - [ ] no lib folder - see https://github.com/nrwl/nx/issues/560#issuecomment-395989584 or https://github.com/nrwl/nx/issues/8156#issuecomment-998440950
    - [x] move foundation enums to types folder?
    - [x] types folder can nest in components root?
    - [x] extract tokens folder to external tokens package (in design tokens repo)
    - [ ] normalize common folder so that all nested folders design-system, sass, test-utils reside next to all component.
    > **YONATAN**: This is not good. The components should reside in their own folders and use common stuff from outside the folder (notice the folder is used for the build scheme -it runs on the components folder and grabs all the components for the build).
    > **Yinon**: `design-system` is needed post build. it's the custom elements registering function
- [x] Need to fix the ui-tests
- [x] Refactoring tasks
  - [x] Need to extract the common to their own lib (like I did to themes right now theme and fonts are inside themes but can be changed easily)
- [ ] Linting
  - [x] Fix the linting for components 
    - [x] Setup the linting rules
    - [x] Make sure all linting passes
  - [x] Setup stylelinting for components
  - [x] Fix the linting for themes 
  - [ ] Setup lintings for docs
    - [ ] Setup style linting
    - [ ] Setup njk md js and json linting?
- [ ] Fix the github actions to work with Nx commands
  - [x] Use npm run nx instead of the current npm scripts
  - [ ] Use affected to publish only what’s need publishing
  - [x] Unit tests can be super simplified since moved to Jest
  - [ ] For some reason, the index.d.ts of the components is created inside dist/libs/libs/components and not dist/libs/components like the rest of the files
- [ ] Need to copy the `package.json`, `README.md`, `license` from libs/components to its dist folder and update it to have all needed values (keywords, license etc.) - probably using rollup

<details open>
  <summary>
    Non blocking tasks
  </summary>

  - [ ] Publish
    - [ ] Publish the components package as vivid
    - [ ] Publish affected publishable libraries (add a publish executor to the relevant projects’ project.json )
  - [ ] Components watch mode through playwright?
  - [ ] Generators
    - [ ] Create a component generator
    - [ ] Create a component page in the docs

</details>


docs/ui-tests/readme.md
```diff
@@ -18,7 +18,7 @@ Our docker image runs the tests on the 3 browser on the Linux OS.
 
 Running the tests can be done locally by running:
 
-`yarn http-server -s & yarn playwright test`
+`npx http-server -s & npx playwright test`
```
yinonov, line 21: now it's pnpm?

PR #138, opened 2022-01-28
test(workspace): integrate husky and lint-staged

yinonov:


yinonov:
@YonatanKra docker prohibits the creation of a folder when running the husky related prepare script
we dont actually need it in the docker run but it ignores the following flag
image

What is the docker container node version? 16 as defined by the action?

see
https://github.com/npm/cli/issues/2434#issuecomment-756445805

PR #141, opened 2022-01-30
docs(button): example toggle button practice

yinonov:


rinaok (commented):
Why is the toggle button under usage examples and not components?

rinaok (changes requested):
.

rinaok (approved):

rachelbt:
What about the "on" state? 
image
image
image

In only text button - if the text doesn't change - how can a user know that the button is toggled? Are we leaving it to our consumers to decide and switch between appearances?

yinonov:
> What about the "on" state? image image image
> 
> In only text button - if the text doesn't change - how can a user know that the button is toggled? Are we leaving it to our consumers to decide and switch between appearances?

yes, it's for consumers to decide. the point here is to reflect how easy it is to tweak a button to be toggle-able.

1 of the issues I experience here is that the _toggled-to_ icon isn't preloaded. thus, firstly shows the preloading spinner

YonatanKra:
> > What about the "on" state? image image image
> > In only text button - if the text doesn't change - how can a user know that the button is toggled? Are we leaving it to our consumers to decide and switch between appearances?
> 
> yes, it's for consumers to decide. the point here is to reflect how easy it is to tweak a button to be toggle-able.
> 
> 1 of the issues I experience here is that the _toggled-to_ icon isn't preloaded. thus, firstly shows the preloading spinner

Won't it always be a problem? I mean, we can't control how fast a user downloads an icon...

docs/usage-examples/toggle-button.md
```diff
@@ -0,0 +1,49 @@
...
+  id="button"
+  connotation='cta'
+  shape='pill'
+  label='not pressed'
+  aria-label="microphone button">
+</vwc-button>
+
+<script>
+  const toggleWithLabel = ({target}) => {
+    target.ariaPressed = !target.ariaPressed;
```
YonatanKra, line 16: This is actually one of the values of `vivid` - a developer would hardly think of changing the aria pressed value...

yinonov: ure right, but does this tiny "program" justify a dedicated component?

yinonov:
> > > What about the "on" state? image image image
> > > In only text button - if the text doesn't change - how can a user know that the button is toggled? Are we leaving it to our consumers to decide and switch between appearances?
> > 
> > 
> > yes, it's for consumers to decide. the point here is to reflect how easy it is to tweak a button to be toggle-able.
> > 1 of the issues I experience here is that the _toggled-to_ icon isn't preloaded. thus, firstly shows the preloading spinner
> 
> Won't it always be a problem? I mean, we can't control how fast a user downloads an icon...

I think this might introduce a lazy/eager loading mechanism for icon component. maybe a task of its own?

YonatanKra (approved):

PR #176, opened 2022-02-04
chore(components): copy using the project.json definition

YonatanKra:
Now when building it also copies the readme and the styles (into the `styles` folder inside `dist/components`).

### DIFF
--- a/libs/components/project.json
+++ b/libs/components/project.json
@@ -33,7 +33,23 @@
         "tsConfig": "libs/components/tsconfig.lib.json",
         "project": "libs/components/package.json",
         "rollupConfig": "libs/components/rollup.config.prod.ts",
-        "assets": [],
+        "assets": [
+          {
+            "input": "./libs/components/**",
+            "glob": "*.md",
+            "output": "."
+          },
+          {
+            "input": "./dist/libs/styles/themes/**",
+            "glob": "*.css",
+            "output": "./styles/themes"
+          },
+          {
+            "input": "./dist/libs/styles/fonts/**",
+            "glob": "*.css",
+            "output": "./styles/fonts"
+          }
+        ],
         "buildableProjectDepsInPackageJsonType": "dependencies",
         "updateBuildableProjectDepsInPackageJson": true
       }
@@ -48,7 +64,6 @@
         "project": "libs/components/package.json",
         "rollupConfig": "libs/components/rollup.config.prod.ts",
         "watch": true,
-        "assets": [],
         "buildableProjectDepsInPackageJsonType": "dependencies",
         "updateBuildableProjectDepsInPackageJson": true
       }
--- a/libs/components/rollup.config.base.ts
+++ b/libs/components/rollup.config.base.ts
@@ -1,13 +1,23 @@
+const path = require('path');
 const typescript = require('@rollup/plugin-typescript');
 const nodeResolve = require('@rollup/plugin-node-resolve').default;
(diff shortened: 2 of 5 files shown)


libs/components/project.json
```diff
@@ -33,7 +33,23 @@
         "tsConfig": "libs/components/tsconfig.lib.json",
         "project": "libs/components/package.json",
         "rollupConfig": "libs/components/rollup.config.prod.ts",
-        "assets": [],
+        "assets": [
+          {
+            "input": "./libs/components/**",
+            "glob": "*.md",
+            "output": "."
+          },
```
yinonov, line 41: no need. we'll create a new dedicated readme

libs/components/rollup.config.base.ts
```diff
@@ -1,23 +1,38 @@
...
 		typescript({
 			noEmitOnError: true,
 			tsconfig: './libs/components/tsconfig.lib.json',
 		}),
 		litCss({
 			specifier: '@microsoft/fast-element',
 			include: ['/**/*.scss'],
-			transform: (data, { filePath }) => {
+			transform: (data, {filePath}) => {
 				console.log(filePath);
```
yinonov, line 29:
```suggestion
```

yinonov (changes requested):

libs/components/project.json
```diff
@@ -33,9 +33,27 @@
         "tsConfig": "libs/components/tsconfig.lib.json",
         "project": "libs/components/package.json",
         "rollupConfig": "libs/components/rollup.config.prod.ts",
-        "assets": [],
+        "assets": [
+          {
+            "input": ".",
+            "glob": "README.md",
```
yinonov, line 39: We don't need this. Let's create a dedicated readme for the components lib

yinonov (changes requested):

libs/components/project.json
```diff
@@ -48,7 +66,6 @@
         "project": "libs/components/package.json",
         "rollupConfig": "libs/components/rollup.config.prod.ts",
         "watch": true,
-        "assets": [],
```
yinonov, line 51: Maybe we want the assets here as well to mimic the same usage as consumers.
Right now docs fetch files from internal repo paths

yinonov (changes requested):

libs/components/project.json
YonatanKra: Np - so I'll just copy the current empty Readme then

libs/components/project.json
YonatanKra: Yea, good idea. I have another issue to take care of `serve` - do we want this to be settled here or then?

YonatanKra: Ok - done. This took like 2 seconds anyway

libs/components/src/lib/anchor/index.ts
```diff
@@ -1,5 +1,5 @@
 import { Anchor, anchorTemplate as template } from '@microsoft/fast-foundation';
-import { designSystem } from '../design-system';
+import { designSystem } from '../../shared/design-system';
```
yinonov, line 2: We need this at run-time

yinonov (changes requested):

libs/components/src/lib/anchor/index.ts
YonatanKra: It exists in runtime. Why not?

yinonov (approved):

PR #178, opened 2022-02-06
refactor(docs): integrate vivid components

### LINKED ISSUE #177: [docs] integrate app template components
Our docs should exemplify how vivid is integrated from the ground up.

All app template scaffolding should include vivid

- top app bar
- side drawer
- layout

yinonov:
Closes #177


rachelbt (dismissed):
Nice!!!
The only glitch I see is the `wvc-layout` that has margin: 24px that causes the main part unnecessary `margin-top`. It's a bit overkill to my opinion to use the layout only for margin.
image

yinonov:
> Nice!!!
> 
> The only glitch I see is the `wvc-layout` that has margin: 24px that causes the main part unnecessary `margin-top`. It's a bit overkill to my opinion to use the layout only for margin.
> 
> image
> 
> 

That's the only way to have the margin set by a token. The alternative is manual style. Any other idea?

rachelbt:
> > Nice!!!
> > The only glitch I see is the `wvc-layout` that has margin: 24px that causes the main part unnecessary `margin-top`. It's a bit overkill to my opinion to use the layout only for margin.
> > image
> 
> That's the only way to have the margin set by a token. The alternative is manual style. Any other idea?

maybe we need to consider exposing tokens for padding / margin and not only colors, or maybe even create global classes for that such as - `padding-base` that is `padding: 16px;` etc. (like Tailwind - but only for the good sides of it ;) )

yinonov:
> > > Nice!!!
> 
> > > The only glitch I see is the `wvc-layout` that has margin: 24px that causes the main part unnecessary `margin-top`. It's a bit overkill to my opinion to use the layout only for margin.
> 
> > > image
> 
> > 
> 
> > That's the only way to have the margin set by a token. The alternative is manual style. Any other idea?
> 
> 
> 
> maybe we need to consider exposing tokens for padding / margin and not only colors, or maybe even create global classes for that such as - `padding-base` that is `padding: 16px;` etc. (like Tailwind - but only for the good sides of it ;) )

Such tokens are in our current Q roadmap

rinaok (approved):

PR #179, opened 2022-02-07
docs(tokens): add color docs

rinaok:
Screen Shot 2022-02-07 at 16 54 07


apps/docs/_includes/navigation.njk
```diff
@@ -1,10 +1,25 @@
...
-    {%  for item in components %}
-    <li>
-        <a href="{{ ('/components/' + item.title) | url }}">{{ item.title }}</a>
-    </li>
-    {% endfor %}
-</ul>
-{% endblock %}
+    <a href="{{ '/' | url}}">Home</a>
+    <ul class="navigation">
+  <li>
```
yinonov, line 4: Maybe use layout?

yinonov (changes requested):

apps/docs/_includes/navigation.njk
rinaok: The navigation should use the side-drawer

PR #180, opened 2022-02-07
docs(layout): add css custom properties table

rinaok:


rinaok:
@yinonov @rachelbt We should consider adding some common style to all the tables in the readme files, cause now it doesn't look good.

rachelbt (approved):

PR #182, opened 2022-02-08
test(layout): add test coverage

rinaok:


yinonov (commented):
.

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+			expect(element).toBeInstanceOf(Layout);
+			expect(element.gutters).toBeUndefined();
+			expect(element.columnBasis).toBeUndefined();
+			expect(element.columnSpacing).toBeUndefined();
+			expect(element.autoSizing).toBeUndefined();
+		});
+	});
+
+	describe('gutters', () => {
+		it('sets correct internal gutters style', async () => {
```
yinonov, line 29: Can u reword using 'should'?

yinonov (changes requested):

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+	});
+
+	describe('gutters', () => {
+		it('should set correct internal gutters style', async () => {
+			const gutters = Size.BaseLarge;
+			(element as any).gutters = gutters;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.gutters-${gutters}`);
+			expect(control)
```
YonatanKra, line 35: Should I expect this to fail if I do not define `gutters`?

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+
+	describe('column-basis', () => {
+		it('should set correct internal column-basis style', async () => {
+			const columnBasis = Size.BaseLarge;
+			(element as any).columnBasis = columnBasis;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.column-basis-${columnBasis}`);
+			expect(control)
```
YonatanKra, line 48: Same question regarding failure here

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+			expect(element).toBeInstanceOf(Layout);
+			expect(element.gutters).toBeUndefined();
+			expect(element.columnBasis).toBeUndefined();
+			expect(element.columnSpacing).toBeUndefined();
+			expect(element.autoSizing).toBeUndefined();
+		});
+	});
+
+	describe('gutters', () => {
+		it('should set correct internal gutters style', async () => {
```
YonatanKra, line 29: Maybe class instead of style?
```suggestion
		it('should set correct internal gutters class', async () => {
```

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+			const gutters = Size.BaseLarge;
+			(element as any).gutters = gutters;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.gutters-${gutters}`);
+			expect(control)
+				.toBeInstanceOf(Element);
+		});
```
YonatanKra, line 37: Should I expect this to fail if I do not define `gutters`?
We'd might want to check "before and after".
```suggestion
			const gutters = Size.BaseLarge;
			const control = getControlElement();
			const classExistsWithoutGutters = control.classes.hasClass(`control.gutters-${gutters}`);
			(element as any).gutters = gutters;
			await elementUpdated(element);

			expect(classExistsWithoutGutters).toEqual(false);
			expect(control.classes.hasClass(`control.gutters-${gutters}`)).toEqual(true);
		});
```
You can even refactor for better readability by extracting to meaningfully named functions.

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+	});
+
+
+	describe('column-basis', () => {
+		it('should set correct internal column-basis style', async () => {
```
YonatanKra, line 42: Same as above both for `style => class` as well as the test itself.

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.column-basis-${columnBasis}`);
+			expect(control)
+				.toBeInstanceOf(Element);
+		});
+	});
+
+	describe('column-spacing', () => {
+		it('should set correct internal column-spacing style', async () => {
```
YonatanKra, line 54: Same as above

libs/components/src/lib/layout/layout.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+			const control = element.shadowRoot?.querySelector(`.control.column-spacing-${columnSpacing}`);
+			expect(control)
+				.toBeInstanceOf(Element);
+		});
+	});
+
+
+	describe('auto-sizing', () => {
+		it('should set correct internal auto-sizing style', async () => {
```
YonatanKra, line 67: Same as above

YonatanKra (changes requested):
See remarks in the tests files

yinonov (commented):
it seems like the width cross browsers' viewport is different

rinaok:
> is the width cross browsers' viewport different? 🤔

@yinonov @YonatanKra The UI test in Firefox looks different.

YonatanKra:
> > is the width cross browsers' viewport different? 🤔
> 
> @yinonov @YonatanKra The UI test in Firefox looks different. 

Yes it is different in each browser

yinonov (approved):

PR #186, opened 2022-02-09
chore(docs): setup serve mode for docs

### LINKED ISSUE #171: Setup a `serve` mode for docs
Should also build the dependencies and watch them

YonatanKra:
Just run `npm run nx serve docs` to run the docs in watch mode with all its current deps.
Note that this is a kind of `hack` outside Nx until we finalize the mechanism for styles. But should be good for development cycle right now.
closes #171


apps/docs/.eleventy.js
```diff
@@ -13,23 +13,25 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addPlugin(syntaxHighlight);
 
   eleventyConfig.addPassthroughCopy({
-    "dist/libs/components/styles/fonts/*.css": "assets/styles/fonts",
-    "dist/libs/components/styles/themes/*.css": "assets/styles/themes",
+    "dist/libs/styles/fonts/*.css": "assets/styles/fonts",
```
yinonov, line 16: Why revert?

YonatanKra: Cause I need to fix the components' `serve` for it to work this way.
So it's currently a hack until I fix it.

yinonov (approved):

PR #193, opened 2022-02-13
chore(components): change `lib` folder to `components`

### LINKED ISSUE #144: sensible readable folder structure. if a folder is redundant - remove
Currently, the components are in the `src/lib` folder. This is so because there are 3 more folders that are not components in the library (`shared`, `types` and `visual-tests`).

The task is two fold:

- [ ] Take out the `shared`, `types` and `visual-tests` outside the library
- [ ] Put the set the `lib` contents as first children of `src`

So the components library will hold only the components without the additional non-component folders.

@yinonov - this is in continuation to our dialog in the PR. Please verify it is a good direction for us.

YonatanKra:
Closes #144 ?


libs/components/tsconfig.lib.json
```diff
@@ -1,7 +1,9 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "rootDirs": ["src", "src/lib"],
+    "rootDirs": ["src",
+      "src/components"
```
yinonov, line 5: any chance of skipping the 2nd level _components_? it just switching 1 redundant folder with another

yinonov (changes requested):

libs/components/tsconfig.lib.json
YonatanKra: @yinonov - this folder is not redundant. It separates the components from other things.
We are going to have lots of components, and we'd like the `src` to remain relatively clean.

PR #194, opened 2022-02-13
test(components): lint rule prevent dom-mutating-attribute

yinonov:


yinonov:
@YonatanKra mind reviewing why workspace lint rule isn't resolved by default?
reproduce by

- `npx nx g @nrwl/linter:workspace-rule my-custom-rule`
- add `@nrwl/nx` to plugins
- add `@nrwl/nx/workspace/my-custom-rule` to rules
- run `npx nx lint components`

yinonov:
fails due nrwl/nx#9395

yinonov:
@rinaok can you please set the elevation's `dp` default value in css rather than in its class member

yinonov:
> @rinaok can you please set the elevation's `dp` default value in css rather than in its class member

@rinaok please review Vonage/vivid-3#268

PR #196, opened 2022-02-14
refactor(components): redefine sizes interface

rachelbt:
fixes the sizing interface.
wording will be set according to where it is applied.
not always the same values but should makes more sense this way

currently in enums -
```ts
export enum BlockSize {
	Condensed = 'condensed',
	Normal = 'normal',
	Extended = 'extended',
}

export enum Size {
	Small = 'sm',
	Medium = 'md',
	Large = 'lg',
}
```

there’s the generic Size and the Block (height) which are interpreted differently

the reason for the difference in block is that this way imply the sizing of `normal` to be a standard while other values are derivative of it.


libs/components/src/lib/button/partials/variables.scss
```diff
@@ -1,2 +1,3 @@
 $icon-gap: --button-icon-gap;
+$stacked-icon-gap: --stacked-button-icon-gap;
```
yinonov, line 2: Can we adopt the private access convention of variable prefixed with underline?

yinonov (changes requested):

libs/components/src/lib/button/README.md
```diff
@@ -76,12 +76,22 @@ Set the `appearance` attribute to change the button's appearance.
 <vwc-button label='outlined' appearance='outlined'></vwc-button>
 ```
 
+## Stacked
+Set the `stacked` attribute to change the button's layout to stacked.
+- Default: `false`
+
+```html preview
+<vwc-button label='Stached' appearance='filled' icon='message-sent-line' stacked size='base-small'></vwc-button>
```
yinonov, line 84: Notice the typo steched

yinonov (changes requested):

libs/components/src/lib/button/README.md
rachelbt: I'm the worse with typos :(

libs/components/src/lib/button/README.md
```diff
@@ -76,12 +76,22 @@ Set the `appearance` attribute to change the button's appearance.
 <vwc-button label='outlined' appearance='outlined'></vwc-button>
 ```
 
+## Stacked
+Set the `stacked` attribute to change the button's layout to stacked.
+- Default: `false`
+
+```html preview
+<vwc-button label='stacked' appearance='filled' icon='message-sent-line' stacked size='base-small'></vwc-button>
+<vwc-button label='stacked' appearance='filled' icon='message-sent-line' icon-trailing stacked size='base'></vwc-button>
```
yinonov, line 85: this is another case where _base_ value isn't true.
_base_ stands for a row height of most common components (textfield, button, etc'.)

yinonov: needs to be discussed as I'm yet sure how this should be resolved

libs/components/src/lib/button/partials/variables.scss
rachelbt: added an underscore

libs/components/src/lib/button/README.md
rachelbt: Not sure I got it....

yinonov: the `base` keyword mean a certain row height of common controls. stack button is not the same height.
This is a problem in this new `base` convention

yinonov: > back to the planning table

rachelbt: oh! Now I get it.

yinonov: what do we think of 1 of the former suggestions?

**density**

possible values: `-1` | `0` | `1`
default: `0`

libs/components/src/lib/button/button.scss
```diff
@@ -48,54 +48,41 @@
...
+      @include mixins.get-size-properties(40px, button, 8px, 16px, 20px);
+    }
+  }
+
+  &.stacked {
+    @include mixins.get-size-properties(68px, body-2-bold, 10px, 16px, 18px);
+
+    &.icon-only {
+      .icon {
+        font-size: 90px;
```
rachelbt, line 80: that's a bit too big 🤣 :
image

rachelbt (commented):
in the popup readme - under dismissible, the button need to get `size="condensed"`.

libs/components/src/lib/button/button.scss
yinonov: of course 😅 it's on purpose to notice that this should be attended.

@jshenkman

- what should be the size of icon with no label within a stack button. I imagine it should be bigger?
- is shape pill applicable on a stacked button? don't feel it is designed well

yinonov (approved):

PR #206, opened 2022-02-17
refactor(popup): migrate to fast element

rinaok:


rinaok:
@yinonov @YonatanKra  I receive the following error when calling `computePosition`: 
`Uncaught (in promise) ReferenceError: process is not defined`.

Screen Shot 2022-03-02 at 18 25 44
Screen Shot 2022-03-02 at 18 26 06

Found this issue: https://github.com/floating-ui/floating-ui/issues/1506

YonatanKra:
@rinaok - I don't mean the version in `package.json`.
You should use `@floating-ui/dom` and not `core` IMO.

libs/components/src/lib/popup/popup.ts
```diff
@@ -138,11 +138,10 @@ export class Popup extends FoundationElement {
 		if (!this.open) {
 			return;
 		}
-		if (!this.anchorEl) {
-			console.error('Anchor is not defined');
+		if (!this.anchorEl || this.anchorEl === undefined) {
 			return;
 		}
-		if (!this.popupEl) {
+		if (!this.popupEl || this.popupEl === undefined) {
```
yinonov, line 144: same here

yinonov (changes requested):

yinonov (commented):
if popup strategy is set to _fixed_, it should probably close the popup upon scroll. similar to the browsers strategy. WDYT?

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Popup } from './popup';
+import '.';
+
+const COMPONENT_TAG = 'vwc-popup';
+
+global.ResizeObserver = jest.fn().mockImplementation(() => ({
+	observe: jest.fn(),
+	unobserve: jest.fn(),
+	disconnect: jest.fn(),
+}));
```
rinaok, line 13: @YonatanKra  @yinonov 
Jest fails with an error message stating that ResizeObserver inside floating-ui classes is undefined.
That fixed the problem.

rinaok:
> if popup strategy is set to _fixed_, it should probably close the popup upon scroll. similar to the browsers strategy. WDYT?

yes, I agree. It does look strange. I've added `hide()` Middleware and removed the `shift() `Middleware. I think it's better now. Tell me what you think.

libs/components/src/lib/popup/popup.scss
```diff
@@ -0,0 +1,43 @@
+@use 'sass:map';
+@use '../../../../shared/src/lib/sass/mixins/typography' as typography;
```
rachelbt, line 2: not sure what we need this for

YonatanKra (changes requested):
A really small change :)

libs/components/src/lib/popup/popup.scss
```diff
@@ -0,0 +1,43 @@
...
+
+.control {
+	--border-radius-size: 6px;
+
+	width: min-content;
+	background-color: var(#{--vvd-color-surface-4dp});
```
rachelbt, line 8: there's non need for `#{} `if you are not using a scss variable here.
But we should use them :)
you can set a variable in the start of the file (if we would have more than one its better to open.a new file for it) and set:
`$popup-bg: --vvd-color-surface-4dp;
`and use it in the CSS here and in the arrow bg like this:
`background-color: var(#{$popup-bg});`

change/@vonage-vivid-fff24903-c6b1-40dd-9364-7f55a550515e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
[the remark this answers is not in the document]
yinonov: Popup is alpha. And is probably underlying component for now. I vote 'none'

YonatanKra:
Note that the coverage is very small (reduces our whole repo's coverage by 5%).
image
Seems like some features are not tested:
1. `aria-hidden` is never set to false in the tests.
2. `part=vvid-theme-alternate` is a not-tested branch.
3. `renderArrow` is not called during the tests.
4. `renderDismissButton` is not called during the tests.

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,176 @@
...
+      </div>
+  </vwc-popup>
+</div>
+
+```
+## Strategy
+Use the `strategy` attribute to set the position of the popup.
+
+- Type: `'fixed'` | `'absolute'`
+- Default: `'fixed'`
```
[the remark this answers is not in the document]
rinaok: https://floating-ui.com/docs/computePosition#strategy

libs/components/src/lib/popup/popup.scss
rinaok: removed

libs/components/src/lib/popup/popup.scss
rinaok: done :)

yinonov:
> > if popup strategy is set to _fixed_, it should probably close the popup upon scroll. similar to the browsers strategy. WDYT?
> 
> @yinonov yes, I agree. It does look strange. I've added `hide()` Middleware and removed the `shift() `Middleware. I think it's better now. Tell me what you think.

Maybe we should add html snippet for fixed and express this exact dialog to users?

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,178 @@
...
+  <vwc-popup anchor="strategyAnchor" open strategy="absolute">
+      <div class="content">
+        <vwc-text font-face="body-1-bold" tight><p class="line">Popup title</p></vwc-text>
+        <vwc-text font-face="body-2" tight>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</vwc-text>
+      </div>
+  </vwc-popup>
+</div>
+
+```
+## Z-index
```
yinonov, line 149:
```suggestion
## Top-layer
```

yinonov (changes requested):

libs/components/src/lib/popup/README.md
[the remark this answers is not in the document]
rachelbt: according to floating UI our default is absolute...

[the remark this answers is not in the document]
rinaok: @rachelbt It is absolute by their default, yes. But our default is fixed.
@yinonov Why we changed the default to fixed?

rinaok:
> > > if popup strategy is set to _fixed_, it should probably close the popup upon scroll. similar to the browsers strategy. WDYT?
> > 
> > 
> > @yinonov yes, I agree. It does look strange. I've added `hide()` Middleware and removed the `shift() `Middleware. I think it's better now. Tell me what you think.
> 
> Maybe we should add html snippet for fixed and express this exact dialog to users?

@yinonov We already have a snippet for strategy, what do you want to show there?

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+			await elementUpdated(element);
+			expect(element.shadowRoot?.querySelector('.arrow')).not.toBeNull();
+		});
+
+		it('should render dismiss class', async () => {
+			element.dismissible = true;
+			await elementUpdated(element);
+			expect(element.shadowRoot?.querySelector('.dismissible')).not.toBeNull();
+		});
+	});
```
YonatanKra, line 82: These are two different API's.
```suggestion
	describe('render arrow', () => {
		it('should render an arrow icon if arrow is true', async () => {
			// TODO
		});
		
		it('should not render an arrow icon if arrow is false', async () => {
			// TODO
		});
		
		it('should remove the arrow class on the container if arrow is false', () => {
		        const controlElement = element.shadowRoot?.querySelector('.control');
			expect(controlElement.classList.contains('.arrow')).toEqual(false);
		});
		
		it('should set the arrow class on the container if arrow is true', () => {
		        element.arrow = true;
			await elementUpdated(element);
			const controlElement = element.shadowRoot?.querySelector('.control');
			expect(controlElement.classList.contains('.arrow')).toEqual(true);
		});
	});
```
This suggestion only covers the idea for `arrow`. Do the same for dismissible.

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+		it('should set to alternate', async () => {
+			expect(getControlElement(element).getAttribute('part')).toEqual('');
+			element.alternate = true;
+			await elementUpdated(element);
+			expect(getControlElement(element).getAttribute('part')).toEqual('vvd-theme-alternate');
+		});
+	});
+
+	describe('accessibility', () => {
+		it('should set aroa-hidden', async () => {
```
YonatanKra, line 94:
```suggestion
		it('should set aria-hidden', async () => {
```

YonatanKra (changes requested):

libs/components/src/lib/popup/popup.spec.ts
rinaok: @YonatanKra It won't work, the arrow and dismiss classes are not inside the control.
Also the "should remove the arrow class on the container if arrow is false" check is not necessary cause "initializes as a vwc-popup" test already checks it.

YonatanKra: That's exactly why we should add the correct test to document where we should expect the class.
BTW - is this class exposed as API? I think what's more important is that we have an icon/button, no?

YonatanKra (approved):

yinonov:
https://github.dev/Vonage/vivid-3/blob/374ff2f9087b8ff73e0a60b3d54067c1cf1f4225/libs/components/rollup.config.prod.ts#L41-L44

libs/components/rollup.config.prod.ts
```diff
@@ -36,7 +38,10 @@ module.exports = function setVividRollupConfig(config) {
 	delete output.entryFileNames;
 
 
-	const plugins = [...config.plugins];
+	const plugins = [...config.plugins,
+		replace({
+			'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+		})];
```
yinonov, line 44: This should be checked thoroughly

yinonov (changes requested):

libs/components/rollup.config.prod.ts
YonatanKra: That's the official solution

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,184 @@
...
+	 * HTML Attribute: anchor
+	 */
+	@attr anchor: string = '';
+
+	/**
+	 * popup's anchor element
+	 *
+	 * @private
+	 */
+	@attr private anchorEl: Element | null | undefined;
```
yinonov, line 85: why is this an interface? wasn't it a getter? have we decided to expose it as well?
we already have `anchor` member. it can, alone, accept _string_ or _HTMLElement_

yinonov (changes requested):

libs/components/src/lib/popup/popup.ts
rinaok: it's private, it's not an interface

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -77,10 +77,10 @@ describe('vwc-popup', () => {
...
 	describe('render dismiss', () => {
-		it('should remove the dismiss class on the container if arrow is false', async () => {
+		it('should remove the dismiss class on the container if dismissible is false', async () => {
 			expect(element.shadowRoot?.querySelector('.dismissible')).toBeNull();
 		});
-		it('should set the dismiss class on the container if arrow is true', async () => {
+		it('should set the dismiss class on the container if dismissible is true', async () => {
 			element.dismissible = true;
 			await elementUpdated(element);
 			expect(element.shadowRoot?.querySelector('.dismissible')).not.toBeNull();
```
yinonov, line 89: I know we talked about this today but I missed the part where `getControlElement(element)` is actually responsible for querying the `.control`. so we better use it

yinonov (changes requested):

change/@vonage-vivid-fff24903-c6b1-40dd-9364-7f55a550515e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
yinonov, line 2: exist in 2
```suggestion
  "type": "none",
```

libs/components/src/lib/enums.ts
```diff
@@ -35,14 +35,6 @@ export enum Size {
 	BaseLarge = 'base-large',
 }
 
-export enum Position {
-	Top = 'TOP',
-	Bottom = 'BOTTOM',
-	Start = 'START',
-	Center = 'CENTER',
-	End = 'END',
-}
```
yinonov, line 44: where will this move to? we might wanna have a single source of truth for these types. components using popup under the hood might need to proxy it

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+      </div>
+  </vwc-popup>
+  <vwc-popup id="popup" anchor="anchor" open corner="bottom">
+      <div class="content">
+        <vwc-text tight font-face="body-2">bottom</vwc-text>
+      </div>
+  </vwc-popup>
+</div>
+```
+## Top-layer
```
yinonov, line 204:
```suggestion
## Caveat
```

libs/components/src/lib/popup/popup.template.ts
```diff
@@ -0,0 +1,46 @@
...
+					<slot></slot>
+					${(x) => (x.dismissible ? renderDismissButton() : '')}
+				</div>
+				${(x) => (x.arrow ? renderArrow() : '')}
+			</div>
+		</vwc-elevation>
+	</div>`;
+
+const renderDismissButton = () => {
+	return html`<vwc-button @click="${x => x.handleDismissClick()}"
```
yinonov, line 40: small maybe?

```suggestion
	return html`<vwc-button size="base-small" @click="${x => x.handleDismissClick()}"
```

libs/components/src/lib/popup/ui.test.ts
```diff
@@ -0,0 +1,52 @@
+import * as path from 'path';
```
yinonov, line 1: snapshots are messed up. like, not anchored properly

libs/components/src/lib/popup/ui.test.ts
```diff
@@ -0,0 +1,52 @@
...
+</style>`;
+
+const componentName = 'popup';
+test('should have all connotations', async ({ page }: { page: Page }) => {
+	const template = extractHTMLBlocksFromReadme(path.join(__dirname, 'README.md'))
+		.reduce((htmlString: string, block: string) => `${htmlString} ${popupStyle} <div style="margin: 5px;">${block}</div>`, '');
+
+	await loadComponent({
+		page,
+		componentName,
```
yinonov, line 37: how did you import more components to the page? I'm seeing _icon_ which isn't imported by _popup_

yinonov (changes requested):

libs/components/src/lib/enums.ts
rinaok: I didn't move it. I'm using floating-ui's enums.

libs/components/src/lib/popup/ui.test.ts
yinonov: nice! 🥇

yinonov (approved):

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+```js
+<script type='module'>
+    import '@vonage/vivid/popup';
+</script>
+```
+## Anchor
+
+Use the `anchor` attribute to reference the ID to element in the popup’s owner document.
+
+Since the popup is positioned absolutely to its anchor, it is best to wrap it with a div element and set its position `relative`.
```
YonatanKra, line 15:
```suggestion
The popup is positioned absolutely to its anchor.  It is recommended to wrap it with a div element and set its position `relative`.
```
Would also appreciate a short description like: "how things can go wrong" or "what adding the wrapper solves" sentence.

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+  const popup = document.getElementById("popup");
+  const button = document.getElementById("anchor2");
+  button.addEventListener('click', toggleOpen);
+  function toggleOpen() {
+	  popup.open = !popup.open;
+  }
+</script>
+```
+## Arrow
+Use the `arrow` attribute to add small triangle to indicate the trigger element.
```
YonatanKra, line 103:
```suggestion
Adds a small triangle to indicate the trigger element.
```

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+  button.addEventListener('click', toggleOpen);
+  function toggleOpen() {
+	  popup.open = !popup.open;
+  }
+</script>
+```
+
+## Dismissible
+
+Use the `dismissible` attribute to add close button to the popup.
```
YonatanKra, line 66:
```suggestion
Adds a close button to the popup.
```

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+  <vwc-icon id="anchor3" type='info-line'></vwc-icon>
+  <vwc-popup anchor="anchor3" open arrow>
+      <div class="content">
+        <vwc-text tight font-face="body-2">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</vwc-text>
+      </div>
+  </vwc-popup>
+</div>
+```
+## Alternate
+Use the `alternate` attribute to set the color-scheme to dark.
```
YonatanKra, line 132:
```suggestion
Sets the color-scheme to dark or light (depending on current user's system settings).
```

libs/components/src/lib/popup/README.md
```diff
@@ -0,0 +1,211 @@
...
+  <vwc-popup anchor="anchor4" open alternate>
+      <div class="content">
+        <vwc-text tight font-face="body-2">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</vwc-text>
+      </div>
+  </vwc-popup>
+</div>
+```
+## Corner
+
+Use the `corner` attribute to set the placement of the popup.
```
YonatanKra, line 162:
```suggestion
Sets the placement of the popup around the anchor.
```
Is this correct? Around the anchor?

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,187 @@
...
+
+	constructor() {
+		super();
+		this.corner = 'left'; // default corner
+	}
+
+	override connectedCallback(): void {
+		super.connectedCallback();
+	}
```
YonatanKra, line 95: What's that for?

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,187 @@
...
+
+	override disconnectedCallback(): void {
+		super.disconnectedCallback();
+		if (this.cleanup) { this.cleanup(); }
```
YonatanKra, line 99: Is `cleanup` an api? It is not documented and not tested. Why would cleanup no exist?

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,187 @@
...
+		this.hide();
+	}
+
+	private assignPopupPosition(data: any): void {
+		const { x: popupX, y: popupY } = data;
+		const { referenceHidden } = data.middlewareData.hide;
+		Object.assign(this.popupEl.style, {
+			left: `${popupX}px`,
+			top: `${popupY}px`,
+			visibility: referenceHidden ? 'hidden' : 'visible',
```
YonatanKra, line 164: One of these cases is not tested. Don't know which without running locally.

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,187 @@
...
+		});
+	}
+
+	private assignArrowPosition(data: any): void {
+		const { x: arrowX, y: arrowY } = data.middlewareData.arrow;
+		const staticSide: any = { top: 'bottom', right: 'left', bottom: 'top', left: 'right' };
```
YonatanKra, line 170: Why not set as constant at the top of the file?
And about the content - it is a WTF moment :)
Can you explain the usecase plz? There's no test for it. The function `updatePosition` is an open API (public) but not tested, so hard to tell what this function is supposed to do.

libs/components/src/lib/components.ts
```diff
@@ -7,5 +7,5 @@ export * from './layout';
 export * from './elevation';
 export * from './breadcrumb-item';
 export * from './breadcrumb';
-export * from './side-drawer';
```
YonatanKra, line 10: What happened to side-drawer?

libs/components/src/lib/popup/ui.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { expect, Page, test } from '@playwright/test';
+import {
+	extractHTMLBlocksFromReadme,
+	loadComponents,
+	loadTemplate
+} from '../../visual-tests/visual-tests-utils';
+
+const components = ['popup', 'text', 'icon', 'button'];
+test('should show the component', async ({ page }: { page: Page }) => {
+	const template = extractHTMLBlocksFromReadme(path.join(__dirname, 'README.md'))
```
YonatanKra, line 11: Are these use cases include all of the `top`/`bottom`/`left`/`right` use cases possible?

YonatanKra (changes requested):

libs/components/src/lib/popup/README.md
rinaok: yes

libs/components/src/lib/popup/popup.ts
rinaok: You can read about cleanup here: https://floating-ui.com/docs/autoUpdate.

change/@vonage-vivid-fff24903-c6b1-40dd-9364-7f55a550515e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
YonatanKra, line 2: Why is the type `none`? Why not `minor`? Isn't it a new feature?

rinaok: @yinonov changed it to none

libs/shared/src/lib/test-utils/index.spec.ts
```diff
@@ -29,6 +29,11 @@ describe(`test-utils`, function () {
       expect(element.id).toEqual('test');
       expect(element.parentElement?.parentElement?.tagName).toEqual('BODY');
     });
+    it(`should return a dom element under div under body with add`, function () {
```
YonatanKra, line 32: image

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+		element = await fixture(`<${COMPONENT_TAG}></${COMPONENT_TAG}>`) as Popup;
+	});
+
+	describe('clean observable', () => {
+		it('should clean observable on disconnectedCallback', async function () {
+			await setPopupAndAnchor();
+			element.anchor = 'anchor';
+			await elementUpdated(element);
+			element.disconnectedCallback();
+			expect(true).toBeTruthy();
```
YonatanKra, line 35: This is not the test that should be here.
You should `spyOn` the `element.updatePosition` method and see that it stops running after the element disconnects if you change a position.

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+import { Popup } from './popup';
+import '.';
+
+
+const COMPONENT_TAG = 'vwc-popup';
+
+describe('vwc-popup', () => {
+	let element: Popup;
+
+	global.ResizeObserver = jest.fn()
```
YonatanKra, line 18: I think this should be defined in a `beforeEach` - otherwise `element.updatePosition` might differ in every test. Also, for testing the cleanup, you should set `observe` to run a spy.
`observeSpy = jest.fn(element.updatePosition)`
and then:
`observe: observeSpy,`

libs/components/src/lib/popup/ui.test.ts
YonatanKra, line 11: I think in this case, because we have so many use cases for positioning, it is worthwhile to hard code them here instead of taking out of docs an incomplete list of them.

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,189 @@
...
+	/**
+	 * Closes the popup
+	 *
+	 * @public
+	 */
+	hide(): void {
+		this.open = false;
+	}
+
+	handleDismissClick(): void {
```
YonatanKra, line 157: Is this tested? Do we actually test the `click` event?

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+		it('should clean observable on disconnectedCallback', async function () {
+			jest.spyOn(element, 'updatePosition');
+			await setAnchor();
+			element.anchor = 'anchor';
+			await elementUpdated(element);
+			expect(observeSpy).toHaveBeenCalled();
+
+			jest.clearAllMocks();
+			element.disconnectedCallback();
+			element.anchor = '';
```
YonatanKra, line 40: This is unclear. We need to change something that triggers the observer.

libs/components/src/lib/popup/popup.ts
```diff
@@ -0,0 +1,185 @@
...
+	override attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
+		super.attributeChangedCallback(name, oldValue, newValue);
+		switch (name) {
+			case 'anchor': {
+				this.anchorEl = this.getAnchorById();
+				break;
+			}
+		}
+		if (this.anchorEl && this.popupEl) {
+			this.cleanup = autoUpdate(this.anchorEl, this.popupEl, () => this.updatePosition());
```
YonatanKra, line 110: now I see something else that's troubling - shouldn't we run `clean` if `anchor` is undefined?
I mean - if we remove the `anchor` from the DOM, the observer is still there which is a possible memory leak.

libs/components/src/lib/popup/ui.test.ts
```diff
@@ -0,0 +1,105 @@
+import * as path from 'path';
+import { expect, Page, test } from '@playwright/test';
+import {
+	extractHTMLBlocksFromReadme,
+	loadComponents,
+	loadTemplate
+} from '../../visual-tests/visual-tests-utils';
+
+const components = ['popup', 'text', 'icon', 'button'];
+test('should show the component', async ({ page }: { page: Page }) => {
```
YonatanKra, line 10: Do we still need this if we have the new test? Does this test anything not tested below?

libs/components/src/lib/popup/ui.test.ts
```diff
@@ -0,0 +1,105 @@
...
+	const testWrapper = await page.$('#wrapper');
+	await page.waitForLoadState('networkidle');
+
+	expect(await testWrapper?.screenshot())
+		.toMatchSnapshot(
+			'./snapshots/popup.png',
+		);
+});
+
+test('should show all positionings', async ({ page }: { page: Page }) => {
```
YonatanKra, line 32: image

libs/components/src/lib/popup/popup.ts
rinaok: yes, you are right

libs/components/src/lib/popup/ui.test.ts
rinaok: this test only tests the corners.

YonatanKra (approved):

PR #208, opened 2022-02-18
ci(docs): deploy to deno

yinonov:


yinonov:
@YonatanKra can u please remove the - now redundant - production build of docs?

.github/workflows/pre-release.yml
```diff
@@ -30,4 +30,4 @@ jobs:
   # ! temporary. site deployment should move to latest release
   call-deploy-site:
```
YonatanKra, line 31: This should be in the `release` right after the `publish` step.

yinonov: thinking again, why? this workflow is called pre-release, meaning it handles the release flow.
We have a synchronous workflow where 'publish-package' is followed by 'deploy' which declaratively prerequisite it

YonatanKra: Oh - ok. So it is not `pre-release`, it is the `release`. 
The filename is misleading...

YonatanKra (approved):
Maybe change the `pre-release.yml` filename to `release-vivid.yml` or something?

YonatanKra (approved):

PR #211, opened 2022-02-20
refactor(elevation): migrate to fast element 

rinaok:
Screen Shot 2022-02-20 at 16 39 30


libs/components/src/lib/elevation/README.md
```diff
@@ -0,0 +1,90 @@
+# vwc-elevation
+
+The _elevation_ component enables a user to set perceived elevation to a certain level.  
+***Note: This component is responsible for the perceived elevation alone and not to DOM elements z-indexing.***
```
yinonov, line 4:
```suggestion
**Note: This component is responsible for the perceived visual elevation alone and not to the HTML elements stacking context.**
```

yinonov (changes requested):

libs/components/src/lib/elevation/elevation.scss
```diff
@@ -0,0 +1,31 @@
...
+@each $dp in $dps {
+	#{get-default-selector($dps, $dp)} {
+		background-color: var(--vvd-color-surface-#{$dp}dp);
+		filter: var(--vvd-shadow-surface-#{$dp}dp);
+	}
+}
+
+.control {
+	--elevation-border-radius: 6px;
+	width: fit-content;
```
yinonov, line 29: why even setting this? it default to a block flow. can be controlled with the `host` simply by setting `display: block` and a `width`
```suggestion
```

yinonov (changes requested):

libs/components/src/lib/elevation/elevation.scss
rachelbt: On one hand you're right. On the other - the elevation need to wrap its content in a way. Why does it bothers you that we contain it to fit its containing child?

rachelbt:
@rinaok the dp=0 is wrong - it suppose to give an outline "border".
like in the [StoryBook of vivid](https://vivid.vonage.com/?path=/story/alpha-components-elevation--all-elevations)

libs/components/src/lib/elevation/elevation.scss
rinaok: @yinonov I haven't added this. I copied the CSS file from Vivid. If you want to change it, perhaps you should also change it in Vivid.

rachelbt: because of the bug that @rinaok found in the example of the elevation hover effect - theres no need for it as it doesn't work with nested components :)
You can remove it.

libs/components/src/lib/elevation/README.md
```diff
@@ -0,0 +1,90 @@
+# vwc-elevation
+
+The _elevation_ component enables a user to set perceived elevation to a certain level.  
+**Note: This component is responsible for the perceived visual elevation alone and not to the HTML elements stacking context.**
```
yinonov, line 4:
```suggestion
**Note: This component is responsible for the perceived visual elevation alone and not for the HTML elements stacking context.**
```

yinonov (changes requested):

yinonov (changes requested):
let #229 in first and validate elevation tokens

change/@vivid-nx-styles-a3612f30-0b1f-4c32-94c7-60e77193d012.json
```diff
@@ -0,0 +1,7 @@
+{
```
yinonov, line 1: not relevant anymore?

yinonov (changes requested):

yinonov (approved):

PR #235, opened 2022-03-03
feat(button): add states style

rachelbt:
still missing design tokens of colors to match the design


libs/shared/src/lib/sass/mixins/_states.scss
```diff
@@ -0,0 +1,49 @@
...
+      background-color: var(--vvd-color-neutral-70);
+    }
+
+    //ghost & outlined
+    &:not(.appearance-filled) {
+      &:hover {
+        background-color: var(--vvd-color-neutral-10);
+      }
+      &:active {
+        background-color: var(--vvd-color-neutral-20);
```
yinonov, line 18: all variables should refer to scss consts

yinonov (changes requested):

libs/shared/src/lib/sass/mixins/_states.scss
```diff
@@ -0,0 +1,49 @@
...
+      }
+    }
+  }
+  $connotation: (
```
yinonov, line 22: we wont always have connotation, so maybe just have a default style if no connotation

libs/shared/src/lib/sass/mixins/_states.scss
```diff
@@ -0,0 +1,49 @@
...
+        background-color: var(--vvd-color-#{$key}-90);
+      }
+
+      &:not(.appearance-filled) {
+        &:hover {
+          background-color: var(--vvd-color-#{$key}-10);
+        }
+        &:active {
+          background-color: var(--vvd-color-#{$key}-20);
+        }
```
yinonov, line 45: we need to map this our in the mixin
like in
https://github.com/Vonage/vivid-3/blob/6e76ede9b7208b95eb20eb4b3d9677ba094cdc65/libs/shared/src/lib/sass/mixins/connotation/_variables.scss#L9-L46

yinonov (changes requested):

libs/shared/src/lib/sass/mixins/_states.scss
rachelbt: I added a default state

libs/shared/src/lib/sass/mixins/_states.scss
rachelbt: added colors to the map and used them in the state mixin

yinonov (changes requested):
Should we disable indication on a disabled/idle control?

rachelbt:
> 

Right!! I've restricted the style when not idle. I don't think we need to target when disabled (as attribute) because it added the class (*-idle) anyway, right?

libs/components/src/lib/button/button.scss
```diff
@@ -4,7 +4,7 @@
 
 @use '../../../../shared/src/lib/sass/mixins/connotation/config' with (
 	$subset: primary cta success alert,
-	$shades: diverse,
+	$shades: diverse contrast soft softer,
```
yinonov, line 7: Softer? Softest? Final.final.design?

rachelbt: is _light_ better?

yinonov: yeap

yinonov: muted, pale, dull, dim etc'...

rachelbt: ohhhh - pale I like!

yinonov:
@rachelbt run `npx beachball change` to create the change file. it's a `none` type

yinonov (dismissed):

yinonov:
> decided to control the states via appearance and not as a separated mixin

do we still need `libs/shared/src/lib/sass/mixins/_states.scss`?

PR #246, opened 2022-03-10
feat(breadcrumb-item): breadcrumb item component

YonatanKra:


libs/components/src/lib/breadcrumb-item/breadcrumb-item.scss
```diff
@@ -0,0 +1,20 @@
...
+
+.breadcrumb-item {
+	@include typography.typography-cat-shorthand('body-2');
+
+	display: flex;
+	align-items: center;
+	color: var(#{theme-consts.$vvd-color-neutral-70});
+}
+
+.control a {
```
rachelbt, line 12: there's no parent with class "control" so this make no sense... You can target the a within the breadcrumb-item class

libs/components/src/lib/breadcrumb-item/breadcrumb-item.scss
```diff
@@ -0,0 +1,20 @@
+@use '../../../../shared/src/lib/sass/mixins/typography' as typography;
+@use '../../../../../node_modules/@vonage/vivid-tokens/dist/sass/themes/consts' as theme-consts;
+
+.breadcrumb-item {
+	@include typography.typography-cat-shorthand('body-2');
+
+	display: flex;
+	align-items: center;
+	color: var(#{theme-consts.$vvd-color-neutral-70});
```
rachelbt, line 9: add here:
a { style }
to target only A inside the item

apps/docs/_data/components.json
```diff
@@ -41,5 +41,14 @@
       "/assets/modules/components/elevation/index.js",
       "/assets/modules/components/layout/index.js"
     ]
+  },
+  {
+    "title": "breadcrumb-item",
+    "markdown": "./libs/components/src/lib/breadcrumb-item/README.md",
+    "modules": [
+      "/assets/modules/components/breadcrumb-item/index.js",
+      "/assets/modules/components/anchor/index.js",
```
yinonov, line 50: why anchor?

yinonov (changes requested):

yinonov (changes requested):

libs/components/src/lib/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,38 @@
...
+	FoundationElementDefinition,
+} from '@microsoft/fast-foundation';
+import { classNames } from '@microsoft/fast-web-utilities';
+import {textAnchorTemplate} from '../text-anchor/text-anchor.template';
+import type { BreadcrumbItem } from './breadcrumb-item';
+import '../icon';
+
+const getClasses = (_: BreadcrumbItem) =>
+	classNames(
+		'control', 'breadcrumb-item'
```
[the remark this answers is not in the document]
yinonov: name of the component is out. _control_ is the new convention

libs/components/src/lib/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,38 @@
...
+export const BreadcrumbItemTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<BreadcrumbItem> =
+  (context: ElementDefinitionContext, definition: FoundationElementDefinition) => html`
+      <div roll="listitem" part="listitem" class="${getClasses}">
+        ${when(x => x.text && !x.href,
+		html<BreadcrumbItem>`${x => x.text}`)}
+
+  ${when(x => x.text && x.href && x.href.length > 0,
```
yinonov, line 33: you tweaked the conditions.
I think the original conditions were to decide if the breadcrum-item is a linkable text or just the commonly last item. which is the current page.

libs/components/src/lib/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,24 @@
+import { BreadcrumbItem as FastBreadcrumbItem } from '@microsoft/fast-foundation';
+import {attr} from '@microsoft/fast-element';
+
+/**
+ * Base class for breadcrumb-item
+ *
+ * @public
+ */
+export class BreadcrumbItem extends FastBreadcrumbItem {
```
yinonov, line 9: oh, I see the problem with extending our `text-anchor` here. it's still repeating same members. the aspiration is to not repeat our code.

notice how the `text-anchor`'s _text_ member will soon be modified with it's reflection strategy. but this class's _text_ member won't
https://github.com/Vonage/vivid-3/blob/91ac6640621faebd4c1670f3e30802a091954eb8/libs/components/src/lib/text-anchor/text-anchor.ts#L18-L19

yinonov (changes requested):

apps/docs/_data/components.json
```diff
@@ -41,5 +41,14 @@
...
+      "/assets/modules/components/icon/index.js"
```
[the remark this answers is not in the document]
YonatanKra: Actually - that's a good point!
Anchor is no longer there, right?
And another thing to note is this:
image
Seems like it is bundled.
Another confirmation we get for the bundling is the UI test that sets the `breadcrumb-item` in isolation - and still calls for `vwc-icon`.
So I guess this is working.
See the docs still work when I remove these two lines.

libs/components/src/lib/breadcrumb-item/breadcrumb-item.template.ts
YonatanKra: I have, because in the figma design we also have this case:
`link > ... > text`
Where the `...` is simple text indicating a lot of links?
Our need is a bit different than `fast`.
@yinonov 
Do we want the `breadcrumb-item` to decide if it is the last one? Or is it the concern of `breadcrumb` or even the user?

libs/components/src/lib/breadcrumb-item/breadcrumb-item.ts
YonatanKra: Good point. I extended `text-anchor` instead of `fastBreadcrumbItem`. Now we get all the goodies of `text-anchor`.
Thing is - now we need to test everything `text-anchor` does as well...

libs/components/src/lib/breadcrumb-item/breadcrumb-item.scss
YonatanKra: Also added `vwc-icon { style }`

libs/components/src/lib/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,15 @@
+import {attr} from '@microsoft/fast-element';
+import {TextAnchor} from '../text-anchor/text-anchor';
+
+/**
+ * Base class for breadcrumb-item
+ *
+ * @public
+ */
+export class BreadcrumbItem extends TextAnchor {
+	@attr({ mode: 'boolean' }) separator = false;
```
yinonov, line 10: TS don't yell on missing override annotation here?

YonatanKra: It is not an override. It is an extesion. Tect anchor doesn't have a separator

libs/components/src/lib/breadcrumb-item/breadcrumb-item.template.ts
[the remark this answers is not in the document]
rachelbt: I know. But here its a problem because the a also has this class name...

libs/components/src/lib/breadcrumb-item/breadcrumb-item.scss
```diff
@@ -0,0 +1,22 @@
...
+
+	a {
+		color: var(--vvd-color-cta-70);
```
rachelbt, line 12: this is not working?
color: var(#{theme-consts.$vvd-color-cta-70});

yinonov (approved):

PR #249, opened 2022-03-11
chore(text-anchor): feature underlying component

yinonov:


libs/components/src/lib/text-anchor/README.md
```diff
@@ -0,0 +1,14 @@
...
+underlying component to be used as inlined text anchor element
+
+TODO document for maintainers usage
+
+```js
+<script type="module">import '@vonage/vivid/text-anchor';</script>
+```
+
+```html preview
+<vwc-text-anchor>shouldn't show</vwc-text-anchor>
```
YonatanKra, line 12: That's not good IMO, as **some** might this it is a good example (just won't read the content inside the HTML.

libs/components/src/lib/text-anchor/text-anchor.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	describe('basic', () => {
+		it('should be initialized as a vwc-text-anchor', async () => {
+			expect(element).toBeInstanceOf(TextAnchor);
+			expect(element.text).toEqual('');
+		});
+	});
+
+	describe('text', () => {
+		it('set text property to node', async () => {
```
YonatanKra, line 22: All the a11y bindings are not tested.

libs/components/src/lib/text-anchor/README.md
yinonov: Of course. Readme should be all different

YonatanKra (approved):

PR #250, opened 2022-03-11
refactor(patterns): affix icon trailing condition

yinonov:


libs/components/src/shared/patterns/affix.ts
```diff
@@ -1,7 +1,7 @@
 import { attr, html } from '@microsoft/fast-element';
 import type { ViewTemplate } from '@microsoft/fast-element';
 import type { ElementDefinitionContext } from '@microsoft/fast-foundation';
-import { classNames } from '@microsoft/fast-web-utilities';
+// import { classNames } from '@microsoft/fast-web-utilities';
```
YonatanKra, line 4: If this is commented, it can be removed...
```suggestion

```

libs/components/src/shared/patterns/affix.ts
```diff
@@ -42,38 +42,38 @@ export class PrefixOrSuffix extends Prefix {
...
+// /**
+// 	* The template for the affixed (prefixed or suffixed) element.
+// 	* For use with {@link AffixIconWithTrailing}
+// 	*
+// 	* @param context
+// 	* @public
+// 	*/
+// export const prefixOrSuffixTemplate: (context: ElementDefinitionContext) => ViewTemplate<AffixIconWithTrailing> =
+// 	(context: ElementDefinitionContext) => {
+// 		const iconTag = context.tagFor(Icon);
```
YonatanKra, line 67: Can this be removed?
```suggestion

```

libs/components/src/shared/patterns/affix.ts
```diff
@@ -42,38 +42,38 @@ export class PrefixOrSuffix extends Prefix {
...
+// 			['icon-trailing', iconTrailing],
+// 		);
 
-		const iconTemplate = html`<span class="${classes}"><${iconTag} :type="${(x) => x.icon}"></${iconTag}></span>`;
+// 		const iconTemplate = html`<span class="${classes}"><${iconTag} :type="${(x) => x.icon}"></${iconTag}></span>`;
 
-		return html`${(x) => (x.icon ? iconTemplate : '')}`;
-	};
+// 		return html`${(x) => (x.icon ? iconTemplate : '')}`;
+// 	};
```
YonatanKra, line 79: Can these be removed as well?

libs/components/src/shared/patterns/affix.ts
```diff
@@ -26,7 +26,7 @@ export class Prefix {
 	*
 	* @public
 	*/
-export class PrefixOrSuffix extends Prefix {
+export class AffixIconWithTrailing extends AffixIcon {
```
YonatanKra, line 29: What does that do? What's the meaning of `trailing`? Can you plz add a test that documents this?

YonatanKra (changes requested):
The `Affix` classes are not tested.
Tests are there to spare us regressions as well as document.
If there were tests, I guess I'd have had less questions :)

libs/components/src/shared/patterns/affix.ts
yinonov: not sure how to test patterns, they are used in templates or classes and tested in the consuming component

YonatanKra (approved):

PR #256, opened 2022-03-16
feat(breadcrumb): breadcrumb component

### LINKED ISSUE #113: [breadcrumbs] create new component
refer https://shoelace.style/components/breadcrumb
to examine https://twitter.com/claviska/status/1478874761983840259?s=20

could suffix & prefix be defined by css alone? 

- [ ] Create breadcrumb item
- [ ] Create the breadcrumb "group"

YonatanKra:
Closes #113


libs/components/src/lib/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -1,14 +1,14 @@
...
  * Base class for breadcrumb-item
  *
  * @public
  */
-export class BreadcrumbItem extends TextAnchor {
-	@observable
-		separator: boolean = true;
+export class BreadcrumbItem extends FastBreadcrumbItem {
+	@attr
+		text: string = '';
```
yinonov, line 11: why the switch back?

libs/components/src/lib/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,11 @@
+import { Breadcrumb as FoundationElement } from '@microsoft/fast-foundation';
+import '../breadcrumb-item';
+
+/**
+ * Base class for breadcrumb
+ *
+ * @public
+ */
+export class Breadcrumb extends FoundationElement {
```
yinonov, line 9: not extending foundation's breadcrumb?

libs/components/src/lib/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,11 @@
+import { Breadcrumb as FoundationElement } from '@microsoft/fast-foundation';
+import '../breadcrumb-item';
```
yinonov, line 2: side effect imports should be imported in the `index.ts` only

libs/components/src/lib/breadcrumb/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { breadcrumbTemplate as template } from '@microsoft/fast-foundation';
+import { designSystem } from '../../shared/design-system';
+import styles from './breadcrumb.scss';
+import { Breadcrumb } from './breadcrumb';
+import '../breadcrumb-item';
```
yinonov, line 6: it is not included in the template, therefore not required to be imported in this context. it is the authors' responsibility to import a component they explicitly use

yinonov (changes requested):

libs/components/src/lib/breadcrumb-item/breadcrumb-item.ts
YonatanKra: Because internally fast's `breadcrumb` allows only `typeof Breadcrumb` to be handled.
Otherwise, we would have needed to reimplement the whole of breadcrumbs.

libs/components/src/lib/breadcrumb/breadcrumb.ts
YonatanKra: I am - look what I import as `FoundationElement`.

libs/components/src/lib/breadcrumb/breadcrumb.ts
YonatanKra: Correct. My bad. It was already in index, but for some reason added here...

libs/components/src/lib/breadcrumb/index.ts
YonatanKra: Really? Ok... removing then.

yinonov (approved):

PR #258, opened 2022-03-16
chore(docs): refine style

rachelbt:


apps/docs/assets/styles/main.css
```diff
@@ -73,6 +87,60 @@ li .navigation {
...
+
+/*Main container*/
+/*--------------------------*/
+vwc-side-drawer::before {
+  content: "";
+  display: block;
+  width: 100%;
+  height: 1px;
+  box-shadow: 0 1px 7px 0px var(--vvd-color-neutral-90);
+}
```
yinonov, line 113: not too fond of this one. it collides with the dynamic shadow's flow

yinonov (changes requested):

apps/docs/assets/styles/main.css
rachelbt: but theres no other way to add the shadow to the top app

yinonov: why add shadow?

rachelbt: I removed the shadow

yinonov (approved):

apps/docs/assets/styles/main.css
```diff
@@ -5,15 +5,20 @@ strong {
 /* #endregion reset */
 
 body {
-  color: var(--vvd-color-on-canvas);
```
yinonov, line 8: this breaks the style

yinonov: image

yinonov: typography and color are different concerns. text component should not control text color.
This must not merge as is

rachelbt: removed it after updating from main the change of the color-scheme

yinonov: the background and foreground will align to dark theme but wont be set as our tokens now

rachelbt: I added back the `bg color` and the `color` properties on `body`

yinonov (approved):

PR #259, opened 2022-03-16
fix(docs): remove breadcrumb item from docs

YonatanKra:
This includes

- [x] Simplification of tests
- [x] removal of breadcrumb item from docs
- [x] Solve RC in docs where items loaded after slotted items have been set

### DIFF
--- a/libs/components/src/lib/breadcrumb/README.md
+++ b/libs/components/src/lib/breadcrumb/README.md
@@ -2,6 +2,7 @@
 
 ```js
 <script type="module">import '@vonage/vivid/breadcrumb';</script>
+<script type="module">import '@vonage/vivid/breadcrumb-item';</script>
 ```
 
 ## Common Usage
--- a/libs/components/src/lib/breadcrumb/README.md
+++ b/libs/components/src/lib/breadcrumb/README.md
@@ -1,8 +1,8 @@
 # breadcrumb
 
 ```js
-<script type="module">import '@vonage/vivid/breadcrumb';</script>
 <script type="module">import '@vonage/vivid/breadcrumb-item';</script>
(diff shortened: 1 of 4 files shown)


libs/components/src/lib/breadcrumb/README.md
```diff
@@ -1,6 +1,7 @@
 # breadcrumb
 
 ```js
```
yinonov, line 3: think we should hint authors to pay attention to that order and describe the case of low performance result?

libs/components/src/lib/breadcrumb/breadcrumb.spec.ts
```diff
@@ -6,29 +6,13 @@ import '.';
...
-			text: 'bc4'
-		}
-	];
-	const breadcrumbItemsTemplate = breadcrumbItemsData.reduce((htmlStr: string, breadcrumbItemData: any) => {
-		const href = breadcrumbItemData.href ? `href="${breadcrumbItemData.href}"` : '';
-		const text = `text="${breadcrumbItemData.text}"`;
-		htmlStr += `<vwc-breadcrumb-item ${href} ${text}></vwc-breadcrumb-item>`;
-		return htmlStr;
-	}, '');
+fdescribe('vwc-breadcrumb', () => {
```
yinonov, line 9: typo
```suggestion
describe('vwc-breadcrumb', () => {
```

yinonov (changes requested):

libs/components/src/lib/breadcrumb/README.md
YonatanKra: Not really as 99% of them would use bundlers which solve this issue.

yinonov (approved):

PR #265, opened 2022-03-21
docs(button): summary

rinaok:


libs/components/src/lib/button/button.spec.ts
```diff
@@ -54,66 +54,62 @@ describe('vwc-button', () => {
...
 		it('sets correct internal connotation style', async () => {
 			const connotation = 'cta';
+
+			expect(getControlElement(element).classList.toString()).toEqual('control');
 			(element as any).connotation = connotation;
 			await elementUpdated(element);
-
-			const control = element.shadowRoot?.querySelector(`.control.connotation-${connotation}`);
-			expect(control).toBeInstanceOf(Element);
+			expect(getControlElement(element).classList.toString()).toEqual(`control connotation-${connotation}`);
```
yinonov, line 68: can we align with -
https://github.com/Vonage/vivid-3/blob/ef68d14820fc14e1954600862fac14b7611ca61a/libs/components/src/lib/text/text.spec.ts#L29-L31

also I dont understand why we need to verify the `control` class exist

yinonov (changes requested):

yinonov (changes requested):

PR #271, opened 2022-03-26
refactor(docs): css update

yinonov:


apps/docs/transformers/code-block-demo.js
```diff
@@ -117,11 +117,9 @@ const style = `
 	overflow: hidden;
 }
 .${CBD_DEMO} {
-	padding: 20px;
```
rachelbt, line 120: the padding removal is needed UI-wisely :)
I get that it breaks the design so we can add:
.${CBD_DEMO} iframe {
margin: 20px;
}

yinonov: it doesn't look good in some demos. like if there's shadow it just cuts it. the inner body should probably have padding

rachelbt: I guess that the shadow is also doesn't look complete when the elements are aligned to the top & right ;)

apps/docs/transformers/code-block-demo.js
```diff
@@ -12,9 +12,9 @@ const CBD_DETAILS = 'cbd-details';
 const CBD_BUTTON_SHOW = 'cbd-button-show';
 const CBD_CODE_BLOCK = 'cbd-code-block';
 
-const MAIN_STYLE = '<link rel="stylesheet" href="/assets/styles/main.css">';
+const MAIN_STYLE = '<link rel="stylesheet" href="/assets/styles/iframe.css">';
```
rachelbt, line 15: 👍

rachelbt (dismissed):
now the preview looks great!! :)

rachelbt (approved):

PR #276, opened 2022-03-30
fix(docs): color

rinaok:


yinonov (commented):
why not use `vwc-layout`?

rinaok:
> why not use `vwc-layout`?

@yinonov I use vwc-layout

rachelbt (approved):
👍

PR #277, opened 2022-03-30
chore(icon): add icon tests

YonatanKra:
This should put us in 100% coverage.
@yinonov - notice that it seems like there's a bug in the logic:
In the case where we ask for an icon (fetch), it takes more than 500ms so we have a placeholder and then the icon arrives anywhere between 500 and 2500ms - the placeholder will not be cleared.
It is easy to fix - but I just wanted to test and make sure what works now is working.
If you want, I can add the fix here (already have it as a patch).


libs/components/src/lib/icon/__snapshots__/icon.spec.ts.snap
```diff
@@ -0,0 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
yinonov, line 1: what is that file? 1st of its kind in the repo

libs/components/src/lib/icon/icon.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+	let element: Icon;
+
+	beforeEach(async () => {
+		element = (await fixture(
+			`<${COMPONENT_TAG}></${COMPONENT_TAG}>`
+		)) as Icon;
+	});
+
+	it('should set placeholder on loading state', async function () {
+		element.placeholder = 'hhh';
```
yinonov, line 17: this is not a public interface. should be referred to as state. maybe shouldn't be decorated by _**observable**_

libs/components/src/lib/icon/icon.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+		element.state = 'loading';
```
yinonov, line 18: same as its previous sibling

libs/shared/src/lib/test-utils/index.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  describe(`setAttribute`, function () {
+    it(`should set attribute on the element`, function () {
+      const ele = document.createElement('div');
+      setAttribute(ele, 'style', 'background: black;');
+      expect(ele.getAttribute('style')).toEqual('background: black;');
+    });
+
+    it(`should return the element after rAF`, function (done) {
```
yinonov, line 57: what?

libs/shared/src/lib/test-utils/index.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+  });
+
+  describe(`getControlElement`, function () {
+    it(`should return undefined if no shadow dom exists`, function () {
+      const lightElement = document.createElement('div');
+      expect(getControlElement(lightElement)).toEqual(undefined);
+    });
+
+    it(`should return the control element inside shadowDOM`, function () {
```
yinonov, line 40: I don't think using _control_ as ref will last long as convention
see https://github.com/microsoft/fast/issues/5757

when setting a relevant ref in a template we can plainly query it as property
`el.control // or any other ref name`

yinonov (changes requested):

libs/shared/src/lib/test-utils/index.spec.ts
YonatanKra: I didn't invent it - took it off from existing code as convention.
If convention is obsolete - we can change it. Not relevant to this PR methinks.

libs/shared/src/lib/test-utils/index.spec.ts
YonatanKra: Actually - this is wrong :)

libs/components/src/lib/icon/__snapshots__/icon.spec.ts.snap
yinonov: why is it back? and only with icon?

yinonov (approved):

yinonov (approved):

YonatanKra:
On purpose. I dont want the fetch to complete before the timers run out.

PR #295, opened 2022-04-05
refactor(tooltip): migrate-to-fast-element

rinaok:


libs/components/src/lib/tooltip/tooltip.scss
```diff
@@ -0,0 +1,18 @@
...
+@use '../../../../shared/src/lib/sass/mixins/typography' as typography;
+
+.control {
+	pointer-events: none;
+}
+
+.tooltip {
+	width: var(#{variables.$tooltip-inline-size}, 240px);
+
+	&-title {
```
yinonov, line 12: does tooltip have title? or only text content?

yinonov (changes requested):

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -0,0 +1,30 @@
...
+export const TooltipTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Tooltip> = () => html`
+<vwc-popup class="${getClasses}" arrow alternate="true" 
+corner=${(x) => x.corner} open=${(x) => x.open} anchor=${(x) => x.anchor}
+  exportparts="vvd-theme-alternate">
+  <div class="tooltip">
+    <header part="vvd-theme-alternate" class="tooltip-header">
+      <div class="tooltip-title">${(x) => x.text}</div>
```
YonatanKra, line 27: If I delete this row, tests still pass... need to add a test case for this.

libs/components/src/lib/tooltip/tooltip.scss
```diff
@@ -0,0 +1,18 @@
...
+		@include typography.typography-cat-shorthand('body-2-bold');
+        
+		padding: 0.5rem 0.75rem;
+		color: var(#{theme-consts.$vvd-color-primary});
```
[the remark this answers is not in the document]
rachelbt: tooltip doesn't need background,  it "uses" the bg of the popup.
I guess `$vvd-color-on-canvas `is more accurate here (even-though it gets the same result :) )

libs/components/src/lib/tooltip/tooltip.template.ts
rinaok: @YonatanKra Which row are you exactly talking about?

yinonov (changes requested):

libs/components/src/lib/tooltip/tooltip.template.ts
YonatanKra: Row 27. Delete it - tests still pass.
It's not such a bad thing cause it is tested in the `ui-tests`, but just be aware of that.

YonatanKra (approved):
LGTM

yinonov (approved):

PR #300, opened 2022-04-07
fix(side-drawer): update margin and height

rinaok:


yinonov (commented):
this pattern we adopt to tweak code for individual cases will not scale well.
we should enable options in our docs environment.
create a configuration or presets of what variant of demo canvas we need
e.g.
- padded - `true` | `false`
- canvas-theme - `'transparent'` | `'gradient'` | `'checkered'`

YonatanKra:
> this pattern we adopt to tweak code for individual cases will not scale well. we should enable options in our docs environment. create a configuration or presets of what variant of demo canvas we need e.g.
> 
> * padded - `true` | `false`
> * canvas-theme - `'transparent'` | `'gradient'` | `'checkered'`

@yinonov - can you explain farther about the pattern? We are using the word `pattern` to describe a lot of things it seems.
What I see is missing API - did I get it right?

yinonov:
> > this pattern we adopt to tweak code for individual cases will not scale well. we should enable options in our docs environment. create a configuration or presets of what variant of demo canvas we need e.g.
> > 
> > * padded - `true` | `false`
> > * canvas-theme - `'transparent'` | `'gradient'` | `'checkered'`
> 
> @yinonov - can you explain farther about the pattern? We are using the word `pattern` to describe a lot of things it seems. What I see is missing API - did I get it right?

yes. sometimes I just use pattern as part of the vocabulary. it happens that we need to alter the iframe body quite often. then it's the API for docs that is missing

yinonov:
@rinaok can either close this or change extend the docs API

PR #307, opened 2022-04-09
refactor(components): banner align with affix pattern

yinonov:


libs/components/src/lib/banner/banner.ts
```diff
@@ -20,7 +35,10 @@ export class Banner extends FoundationElement {
 	@attr() role: string | undefined;
 	@attr() text: string | undefined;
 	@attr() connotation: BannerConnotation | undefined;
-	@attr() icon: string | undefined;
```
YonatanKra, line 23: If this is removed, how do I get `this.icon`?

yinonov: applyMixin of affixIcon

YonatanKra: Oh... magic... ok, as long as it is covered by tests, I don't mind the implementation detail :)

YonatanKra (approved):

PR #311, opened 2022-04-10
refactor(accordion-item): migrate-to-fast-element

rinaok:


libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,182 @@
...
+ <vwc-accordion-item heading="accordion item with Icon" icon="chat-solid">
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+
+## Icon-Trailing
+Add the `icon-trailing` attribute to add an icon to the right of the heading text.
```
rachelbt, line 115:
```suggestion
Mind that `icon-trailing` will overide the Indicator```

File: libs/components/src/lib/accordion-item/README.md [hunk 2]
```diff
@@ -0,0 +1,182 @@
...
+  <vwc-accordion-item heading="accordion item with heading">
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+
+## Heading-Level
+Use the `heading-level` attribute to change the accordion item's size.
```
rachelbt, line 30:
```suggestion
Use the `heading-level` attribute to change the accordion heading to fit the page hierarchy.```

**Review by rachelbt — COMMENTED** [review]
👍  😄

File: libs/components/src/lib/accordion-item/README.md [hunk 3]
```diff
@@ -0,0 +1,182 @@
...
+
+- Type: `2` | `3` | `4` | `5` | `6`
+- Default: `3`
+
+```js
```
yinonov, line 35:
```suggestion
```html preview
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,182 @@
...
+<div class="wrapper">
+ <vwc-accordion-item heading="accordion item without indicator" no-indicator>
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+
+## Icon
```
yinonov, line 93: icon is a common member. sort to the end (icon-trailing too)

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,182 @@
...
+</style>
+<div class="wrapper">
+ <vwc-accordion-item heading="accordion item with Metadata" meta="meta-data">
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+## Dense
```
yinonov, line 155: let's make sure this is still required @jshenkman

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,100 @@
...
+	 * HTML Attribute: heading
+	 */
+	@attr({ mode: 'fromView' }) heading = '';
+
+	/**
+	 *
+	 * @public
+	 * HTML Attribute: heading-level
+	 */
+	@attr({ mode: 'fromView', attribute: 'heading-level' }) headingLevel: 2 | 3 | 4 | 5 | 6 = 3;
```
yinonov, line 24: no need to assign value. just default to 3 if no assigned value
```suggestion
	@attr({ attribute: 'heading-level' }) headingLevel: 2 | 3 | 4 | 5 | 6;
```

libs/components/src/lib/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,100 @@
...
+	@attr({ mode: 'fromView' }) icon = '';
+
+	/**
+	 * Indicates the icon affix alignment.
+	 *
+	 * @public
+	 * 
+	 * HTML Attribute: icon-trailing
+	 */
+	@attr({ mode: 'boolean', attribute: 'icon-trailing' }) iconTrailing = false;
```
yinonov, line 49: use the affix pattern mixin both to applyMixin on class and render in template

libs/components/src/lib/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	/**
+	 * Toggles the accordion-item.
+	 *
+	 * @public
+	 */
+	toggleOpen(): void {
+		this.open = !this.open;
+	}
```
yinonov, line 99: feels like another pattern right? along with the `open` member. we often run into this set. better if it comes from a single source of truth

libs/components/src/lib/accordion-item/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { designSystem } from '../../shared/design-system';
+import styles from './accordion-item.scss';
+
+import { accordionItem } from './accordion-item';
+import { accordionItemTemplate as template } from './accordion-item.template';
+import '../icon';
+
+export const vividaccordionItem =
```
yinonov, line 9:
```suggestion
export const vividAccordionItem =
```

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,88 @@
...
+	['noIndicator', noIndicator],
+);
+
+export const accordionItemTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<accordionItem> = () => html<accordionItem>`
+	<div class="${getClasses}">
+		${x => renderPanelHeader(x.headingLevel)}
+		<div class="accordion-item-body">
```
yinonov, line 27:
```suggestion
		<div class="accordion-item-body" id="content" role="region" aria-labelledby="header">
```

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,88 @@
...
+	return html<accordionItem>`
+	<${header} class="accordion-item-header">
+		${renderHeaderButton()}
+	</${header}>
+	`;
+};
+
+const renderHeaderButton = () => {
+	return html<accordionItem>`
+	<button class="accordion-item-button" @click=${x=> x.toggleOpen()}
```
yinonov, line 44:
```suggestion
	<button class="accordion-item-button" id="header" @click=${x=> x.toggleOpen()}
```

yinonov (changes requested):

libs/components/src/lib/accordion-item/README.md
rinaok: It's not a code example, so it does not have a preview

yinonov: doesn't have to annotate with _preview_. but it's an _html_ and not _js_

yinonov: I'm not sure it's very informative for readers. they dont have interior knowledge of the template.
maybe drop in a quote from accordion aria practice that explains the heading hierarchy.
add a snippet preview with saying something like "the following will output..."

yinonov (changes requested):
we discussed the conditional rendering of indicator in correlation with icon position.
it's now known to us but maybe confusing to consumers.

I think we should address it in the docs

edit:
found it
https://github.com/Vonage/vivid-3/pull/311/files#r851794034

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,179 @@
...
+ <vwc-accordion-item heading="accordion item with icon" icon="chat-solid">
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+
+## Icon-Trailing
+Add the `icon-trailing` attribute to add an icon to the right of the heading text.  Mind that `icon-trailing` will overide the Indicator.
```
yinonov, line 154:
```suggestion
Add the `icon-trailing` attribute to add an icon to the right of the heading text.  Mind that `icon-trailing` will override the Indicator.
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,179 @@
...
+  </vwc-accordion-item>
+</div>
+```
+
+  ## Methods
+
+| Method       | Type       | Description                                      |
+| ------------ | ---------- | ------------------------------------------------ |
+| `hide`       | `(): void` | Closes the accordion item from the open state.  |
+| `show`       | `(): void` | Opens the accordion item from the closed state. |
```
yinonov, line 179: I started aligning methods to our members convention. there shouldn't be any difference I think. consider? or think it should be like this?

see event context
https://vivid-dy25ra84f90g.deno.dev/components/calendar

rinaok: It is more clear this way. Perhaps we can discuss it further. It isn't relevant for this PR since all the components need to change.

docs/components.md
```diff
@@ -27,7 +27,7 @@
 | dropdown             | deprecated | obsolete   |
 | elevation            | GA         | GA         |
 | empty state          | GA         | \-         |
-| expension panel      | GA         | WIP        |
+| accordion item      | GA         | WIP        |
```
yinonov, line 30:
```suggestion
| accordion item       | GA         | WIP        |
```

docs/components.md
```diff
@@ -39,7 +39,7 @@
 | layout               | alpha      | GA         |
 | linear progress      | GA         | WIP        |
 | list                 | GA         | WIP        |
-| list expansion panel | GA         | obsolete   |
+| list accordion item | GA         | obsolete   |
```
yinonov, line 42:
```suggestion
| list accordion item  | GA         | obsolete   |
```

libs/components/src/lib/accordion-item/README.md
yinonov: right but this is a good place to start

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+			const control = getControlElement(element);
+			const hasClassOpenBeforeShow = control.classList.contains('open');
+
+			element.show();
+			await elementUpdated(element);
+			const hasClassOpenAfterShow = control.classList.contains('open');
+
+			expect(element.open).toEqual(true);
+			expect(hasClassOpenBeforeShow).toEqual(false);
+			expect(hasClassOpenAfterShow).toEqual(true);
```
yinonov, line 41:
```suggestion
			expect(hasClassOpenBeforeShow).toBeFalsy();
			expect(hasClassOpenAfterShow).toBeTruthy();
```

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+			await elementUpdated(element);
+			const control = getControlElement(element);
+			const hasClassOpenBeforeHide = control.classList.contains('open');
+
+			element.hide();
+			await elementUpdated(element);
+			const hasClassOpenAfterHide = control.classList.contains('open');
+
+			expect(hasClassOpenBeforeHide).toEqual(true);
+			expect(hasClassOpenAfterHide).toEqual(false);
```
yinonov, line 57: same as above

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+			const hasClassOpenAfterShow = control.classList.contains('open');
+
+			expect(element.open).toEqual(true);
+			expect(hasClassOpenBeforeShow).toEqual(false);
+			expect(hasClassOpenAfterShow).toEqual(true);
+		});
+	});
+
+	describe('hide', () => {
+		it('should set "open" to false and remove "open" class', async () => {
```
yinonov, line 46: I think we need to keep definition as simple as possible
```suggestion
		it('should unset "open"', async () => {
```

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+			await elementUpdated(element);
+			const hasClassOpenAfterHide = control.classList.contains('open');
+
+			expect(hasClassOpenBeforeHide).toEqual(true);
+			expect(hasClassOpenAfterHide).toEqual(false);
+		});
+	});
+
+	describe('toggle', () => {
+		it('should toggle when open attribute changes', async () => {
```
yinonov, line 62:
```suggestion
		it('should toggle "open" state', async () => {
```

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,61 @@
...
+	['no-indicator', noIndicator],
+);
+
+export const AccordionItemTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<AccordionItem> = (context) => html<AccordionItem>`
+	<div class="${getClasses}">
+		${x => renderPanelHeader(context, x.headingLevel)}
+		<div class="accordion-item-body" id="content" role="region" aria-labelledby="header">
```
yinonov, line 28: css convention of scoped portions was needed in large scale style files to namespace parts and avoid conflicts
this is redundant for web components and might even deviate from the web components templates convention. templates has their own encapsulated tree

prefixes are no longer required

```suggestion
		<div class="body" id="content" role="region" aria-labelledby="header">
```

yinonov: let's remove it from other classes?

yinonov (changes requested):

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,188 @@
...
+</h4>
+<h5 class="header">
+	<button class="button"></button>
+</h5>
+<h6 class="header">
+	<button class="button"></button>
+</h6>
+```
+## Open
+Use the `open` attribute to indicate whether the accordion item is open.
```
yinonov, line 53:
```suggestion
Use the `open` attribute to set the accordion-item's open state.
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,188 @@
...
+- Type: `boolean`
+- Default: `false`
+
+```html preview
+<style>
+  .wrapper{
+    width: 500px;
+  }
+</style>
+<div class="wrapper">
```
yinonov, line 64: this adds more reading material. remove from all?

yinonov (changes requested):

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,188 @@
...
+ <vwc-accordion-item heading="accordion item with metadata" meta="meta-data">
+    <vwc-text font-face="body-1">
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+    </vwc-text>
+  </vwc-accordion-item>
+</div>
+```
+## Dense
```
yinonov, line 112: still need to discuss (preferably omit). maybe comment out first?

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+## Methods
+
+### Show
+
+- Type: function
+- Returns: `void`
+
+Opens the accordion item.
```
yinonov, line 181: should be ordered above the type & default

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Hide
+
+- Type: function
+- Returns: `void`
+
+Closes the accordion item.
```
yinonov, line 188: why do we have these methods anyway? I think it's a good opportunity to try and question these. are they redundant? do they add some capability which lacks in a property member? maybe it should accept callback and that's its purpose?

yinonov (changes requested):

libs/components/src/lib/accordion-item/README.md
rinaok: @yinonov but then the indicator is far on the right of the example.

libs/components/src/lib/accordion-item/README.md
rinaok: @yinonov My opinion is that we should remove all the hide and show methods.. from the side-drawer and from other components as well.

yinonov: go ahead

libs/components/src/lib/accordion-item/README.md
yinonov:
```suggestion
```html preview
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+```html preview
+<vwc-accordion-item heading="my heading" heading-level="2"></vwc-accordion-item>
+```
+
+which will output the shadow tree heading element to be rendered as a `<h2>` tag
+
+e.g.
+
+```js
```
yinonov, line 39:
```suggestion
```html
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,113 @@
...
+## Heading-Level
+
+Use the `heading-level` attribute (or `headingLevel` property) to set the accordion heading level so it fits correctly within the outline of the page. Read more on [heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+
+- Type: `2` | `3` | `4` | `5` | `6`
+- Default: `3`
+
+The following sets the heading of the accordion-item to level 2
+
+```html preview
```
yinonov, line 31:
```suggestion
```html
```

libs/components/src/lib/accordion-item/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Heading-Level
+
+Use the `heading-level` attribute (or `headingLevel` property) to set the accordion heading level so it fits correctly within the outline of the page. Read more on [heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+
+- Type: `2` | `3` | `4` | `5` | `6`
+- Default: `3`
+
+The following sets the heading of the accordion-item to level 2
```
yinonov, line 29:
```suggestion
The following sets the heading of accordion-item to level 2
```

yinonov (approved):

yinonov (approved):

PR #312, opened 2022-04-11
feat(banner): change banners api

### LINKED ISSUE #310: Change banner's API
- [x] Banner will be shown by default (no more `open` attribute)
- [x] `dismissible` will be replaced by `removable` and will remove the element from the DOM
- [x] The native `remove` method should be monkey patched (in the context of element only) in order to add the remove animation
- [ ] Make the `remove` monkeypatch a mixin `OnRemovalAnimationDelay` as a consumable pattern
- [x] Add a single action button/anchor
  - [x] `action-text` and `action-href` attributes
  - [x] Add anchor tag when button has `href`
  - [x] Add button that fires `vwc-banner:action` when clicked

YonatanKra:
Closes #310


libs/components/src/lib/banner/banner.scss
```diff
@@ -19,7 +19,7 @@ $banner-transition-delay: --transition-delay;
 		@include appearance.appearance;
```
rachelbt, line 20: referring to the code above:

- we should use "`.control`" as the main class as in all our other components.
- you can write the class on the .control itself no need for it to be nested inside `:host`

libs/components/src/lib/banner/banner.scss
```diff
@@ -66,6 +66,6 @@ $banner-transition-delay: --transition-delay;
 	}
 }
 
-.banner.open {
```
rachelbt, line 69: we are using scss you can nest it inside the the class with & :)

libs/components/src/lib/banner/README.md
```diff
@@ -66,25 +58,42 @@ vwc-banner {
...
+
+```html preview
+<vwc-banner text="A banner with an action button">
+    <vwc-button slot="actionItems" appearance="filled" connotation="cta" label="Learn More"></vwc-button>
+</vwc-banner>
+```
+
+## Methods
+
+### remove()
```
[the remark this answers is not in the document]
YonatanKra: Because we override it.

libs/components/src/lib/banner/banner.spec.ts
```diff
@@ -387,38 +261,95 @@ describe('vwc-banner', () => {
...
 
-		it('should close the button only if "dismissible" is true', async function () {
-			await openBanner(element);
+		it('should remove the banner only if "removable" is true', async function () {
+			element.removable = false;
 			element.focus();
+			const spy = jest.spyOn(element, 'remove');
 			element.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' }));
-			expect(element.open).toEqual(true);
+			expect((spy as any).mock.calls.length).toEqual(0);
```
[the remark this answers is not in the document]
YonatanKra: I create a spy and then ask how many times it was called. Huddle me if you need farther explanation.

libs/components/src/lib/banner/banner.scss
yinonov: actually not control but something more general like 'base', see https://github.com/microsoft/fast/issues/5757

yinonov: no need to attend now. can be later refactored with all others

PR #321, opened 2022-04-17
chore(workspace): upgrade playwright to 1.21.0

YonatanKra:


yinonov (approved):

yinonov (approved):

yinonov (commented):
in layout the text font family style is defined by the user agent.
there's no definition of how it should be styled as the text only informs the reader of documentation what this layout is set to

so that text is a wild card

PR #327, opened 2022-04-19
chore(components): migrate calendar event

yinonov:


libs/components/src/lib/calendar-event/calendar-event.template.ts
```diff
@@ -34,12 +37,12 @@ export const CalendarEventTemplate: (
...
 	definition: FoundationElementDefinition
 ) => ViewTemplate<CalendarEvent> = () => html`
-<section
+<div
   style="${getStyles}"
   class="${getClasses}"
   role="button"
   tabindex="0"
 >
   <h2><strong>${x => x.heading}</strong></h2>
```
rachelbt, line 46: if you'll add when - if there's no heading the H2 tag will not be rendered :)
`${when(x => x.heading, <h2><strong>${x => x.heading}</strong></h2>)}` 

and I'n not sure you need the <strong> tag - because you are setting the  bold in the typography in the css

libs/components/src/lib/calendar-event/calendar-event.template.ts
```diff
@@ -34,12 +37,12 @@ export const CalendarEventTemplate: (
...
   <p>${x => x.description}</p>
```
rachelbt, line 47: same with `when`

rachelbt (commented):
LGTM.
I think that if we should add height to the description and add clamp. It will look better:
image

libs/components/src/lib/calendar/calendar.spec.ts
```diff
@@ -292,6 +293,20 @@ describe('vwc-calendar', () => {
 				grid.querySelector('[role="gridcell"i]:nth-child(3)')
 			);
 		});
+
+		it('should move focus from calendar event to its containing gridcell on \'arrowDown\'', async () => {
```
YonatanKra, line 297:
```suggestion
		it(`should move focus from calendar event to its containing gridcell on 'ArrowDown'`, async () => {
```

yinonov:
> LGTM. I think that if we should add height to the description and add clamp. It will look better: image

can this be done based on size constrain?

rachelbt:
> > LGTM. I think that if we should add height to the description and add clamp. It will look better: image
> 
> can this be done based on size constrain?

sadly no...
But we can restrict it maybe only to `duration="1"` and one line clamp and for the rest 2 lines?

yinonov:
> > > LGTM. I think that if we should add height to the description and add clamp. It will look better: image
> > 
> > 
> > can this be done based on size constrain?
> 
> sadly no... But we can restrict it maybe only to `duration="1"` and one line clamp and for the rest 2 lines?

it's a good idea, but let's avoid specifying design on the fly. it doesn't scale well as in the top sticky header.
mind opening a ticket for it?
image

rachelbt:
> > > > LGTM. I think that if we should add height to the description and add clamp. It will look better: image
> > > 
> > > 
> > > can this be done based on size constrain?
> > 
> > 
> > sadly no... But we can restrict it maybe only to `duration="1"` and one line clamp and for the rest 2 lines?
> 
> it's a good idea, but let's avoid specifying design on the fly. it doesn't scale well as in the top sticky header. mind opening a ticket for it? image

opened:
https://github.com/Vonage/vivid-3/issues/462

rachelbt (approved):

PR #331, opened 2022-04-24
refactor(components): remove from-view from string attributes

### LINKED ISSUE #330: Ban `fromView` as a rule
Using `fromView` has its specific usage.  In most cases it causes inconsistency between property state to DOM attribute state (can change the property, the value itself changes but the attribute in the DOM remains untouched).
This task's goal is to remove all `fromView` attributes from the components.
This should be done carefully as **there are cases in which `fromView` is the correct usage**.

yinonov:
Closes #330


libs/components/src/lib/accordion-item/accordion-item.ts
```diff
@@ -10,13 +10,13 @@ import { AffixIconWithTrailing } from '../../shared/patterns/affix';
...
-	 * 
+	 *
 	 *
 	 * @public
-	 * 
+	 *
 	 * HTML Attribute: heading
 	 */
-	@attr({ mode: 'fromView' }) heading = '';
+	@attr heading?: string;
```
yinonov, line 19: Should update all readme

YonatanKra (approved):
LGTM

PR #339, opened 2022-04-27
refactor(side-drawer): remove hide and show, update docs

rinaok:


libs/components/src/lib/side-drawer/side-drawer.ts
```diff
@@ -63,33 +63,24 @@ export class SideDrawer extends FoundationElement {
...
-	 * @public
-	 */
-	hide(): void {
-		this.open = false;
-	}
-
 	handleScrimClick = (): void => {
 		if (this.modal && this.open) {
-			this.hide();
+			this.close();
```
yinonov, line 68: why do we still keep a method?

yinonov (changes requested):

libs/components/src/lib/side-drawer/side-drawer.ts
yinonov: So rename to #close

yinonov (approved):

PR #344, opened 2022-04-28
fix(appearance): rename color tones

rachelbt:


rachelbt:
this issue was solved in https://github.com/Vonage/vivid-3/pull/384

PR #346, opened 2022-04-28
fix(appearance): prevent active to overide disable

rachelbt:


libs/components/src/lib/button/button.scss
```diff
@@ -11,7 +11,7 @@
 
 @use '../../../../shared/src/lib/sass/mixins/appearance/config' as appearance-config with (
 	$appearances: filled outlined ghost,
-	$states: idle hover disabled active,
+	$states: idle hover active disabled,
```
yinonov, line 14: is that all there is? order of declaration? it's implicit and could be hard to obtain.
we probably need more hardened measures

yinonov (changes requested):
should we set the button's cursor to default on disabled?

yinonov (approved):

PR #347, opened 2022-04-28
refactor(fab): migrate to fast-element

rinaok:


rinaok:
The disabled fab, without connotation, is white and not grey like in vivid-2.
@rachelbt  @yinonov Can you check why?

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+			expect(element.icon).toBeUndefined();
+			expect(element.connotation).toBeUndefined();
+			expect(element.appearance).toBeUndefined();
+			expect(element.iconTrailing).toBeFalsy();
+			expect(element.disabled).toBeFalsy();
+		});
+	});
+
+	describe('icon', () => {
+		it('adds an icon to the fab', async () => {
```
YonatanKra, line 29:
```suggestion
		it('should add an icon to the fab', async () => {
```

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+			element.icon = 'home';
+			await elementUpdated(element);
+
+			const icon = element.shadowRoot?.querySelector(ICON_SELECTOR) as Icon;
+			expect(icon).toBeInstanceOf(HTMLElement);
+			expect(icon.type).toEqual('home');
+		});
+
+		it('setting `iconTrailing` set the order of element', async () => {
```
YonatanKra, line 38:
```suggestion
		it('should  set icon to end when `iconTrailing` is set', async () => {
```

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+			const trailingIcon = element.shadowRoot?.querySelector(
+				`.icon-trailing ${ICON_SELECTOR}`,
+			);
+			expect(trailingIcon).toBeInstanceOf(HTMLElement);
+		});
+	});
+
+	describe('label', () => {
+		it('set label property to node', async () => {
```
YonatanKra, line 51:
```suggestion
		it('should set label property to node', async () => {
```

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+			const label = 'lorem';
+
+			element.label = label;
+			await elementUpdated(element);
+			expect(getControlElement(element).textContent?.trim()).toEqual(label);
+		});
+	});
+
+	describe('connotation', () => {
+		it('sets correct internal connotation style', async () => {
```
YonatanKra, line 61:
```suggestion
		it('should set correct connotation class', async () => {
```

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+			const connotation = 'cta';
+  
+			expect(getControlElement(element).classList.toString()).toEqual('control');
+			(element as any).connotation = connotation;
+			await elementUpdated(element);
+			expect(getControlElement(element).classList.toString()).toEqual(`control connotation-${connotation}`);
```
YonatanKra, line 67: There's no need to see if there's `.control` class.
You can do something like this:
```suggestion
      const expectedConnotation = `connotation-${connotation}`;
      const hasConnotationClassBefore = getControlElement(element).classList.contains(expectedConnotation);
      
			(element as any).connotation = connotation;
			await elementUpdated(element);
			
			expect(hasConnotationClassBefore).toEqual(false);
			expect(getControlElement(element).classList.contains(expectedConnotation)).toEqual(true);
```

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+			expect(getControlElement(element).textContent?.trim()).toEqual(label);
+		});
+	});
+
+	describe('connotation', () => {
+		it('sets correct internal connotation style', async () => {
+			const connotation = 'cta';
+  
+			expect(getControlElement(element).classList.toString()).toEqual('control');
+			(element as any).connotation = connotation;
```
YonatanKra, line 65: If you set the element to be of type `Fab` in the first place, it would obsolete the `as any`, right?

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+	describe('icon-only', () => {
+		it('sets correct internal icon-only style', async () => {
+			const icon = 'home-line';
+
+			expect(getControlElement(element).classList.toString()).toEqual('control');
+			(element as any).icon = icon;
+			await elementUpdated(element);
+			expect(getControlElement(element).classList.toString()).toEqual('control icon-only');
+		});
```
YonatanKra, line 90: See comments regarding description and test flow in the `connotation` test

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+	});
+
+	describe('disabled', function () {
+		it('should set disabled class when disabled is true', async () => {
+			const appearance = 'filled';
```
YonatanKra, line 95: Do we need appearance `filled` in order for `disabled` class to be added?

libs/components/src/lib/fab/fab.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export class Fab extends FoundationElement {
+	/**
+	 * Indicates the fab's label.
+	 *
+	 * @public
+	 * @remarks
+	 * HTML Attribute: label
+	 */
+	@attr({ mode: 'fromView' }) label = '';
```
YonatanKra, line 41: Why `fromView`?

YonatanKra (changes requested):
See my comments

libs/components/src/lib/fab/fab.ts
rinaok: @YonatanKra The linter doesn't give me another choice.

libs/components/src/lib/fab/fab.spec.ts
rinaok: @YonatanKra yes

rinaok: @yinonov Can remove the appearance from the fab? Cause it has only filled.

rachelbt:
@rinaok , @yinonov 

why aren't we using the vwc-button component?
It seems like we have duplicated code - both in the template and in the css.
we can wrap the button with div and elevation, something like this:

```
<vwc-elevation dp='4'>
<div class="control" style for control that is needed="display: inline-block; border-radius: 24px;">
<vwc-button appearance='filled' label='add to cart' size='base-large' icon="cart-line" connotation="cta" shape="pill"></vwc-button>
</div>
</vwc-elevation>
```

The thing we need to sort out is the "white" variant that is actually an alternate for button - maybe we need to add part to the button for this use case - same as popup and tooltip.

rinaok:
> @rinaok , @yinonov
> 
> why aren't we using the vwc-button component? It seems like we have duplicated code - both in the template and in the css. we can wrap the button with div and elevation, something like this:
> 
> ```
> <vwc-elevation dp='4'>
> <div class="control" style for control that is needed="display: inline-block; border-radius: 24px;">
> <vwc-button appearance='filled' label='add to cart' size='base-large' icon="cart-line" connotation="cta" shape="pill"></vwc-button>
> </div>
> </vwc-elevation>
> ```
> 
> The thing we need to sort out is the "white" variant that is actually an alternate for button - maybe we need to add part to the button for this use case - same as popup and tooltip.

I agree!

libs/components/src/lib/fab/fab.ts
YonatanKra: `@attr label?: string;` - this should work

rachelbt:
> The disabled fab, without connotation, is white and not grey like in vivid-2. @rachelbt @yinonov Can you check why?

because you're using appearance but the "white" one is not mapped as an appearance. Check my comment regarding using the `vwc-button.`
https://github.com/Vonage/vivid-3/pull/347#issuecomment-1114788708

rinaok:
> 

Updated it with vwc-button. I made it alternate, but it has a problem with being alternate and disabled at the same time.

libs/components/src/lib/fab/fab.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+			element.connotation = connotation as FabConnotation;
+			await elementUpdated(element);
+
+			expect(hasConnotationClassBefore).toEqual(false);
+			expect(getControlElement(element).classList.contains(expectedConnotation)).toEqual(true);
+		});
+	});
+
+	describe('icon-only', () => {
+		it('sets correct internal icon-only style', async () => {
```
YonatanKra, line 76: use the `should` syntax plz like above for all `it`s

libs/components/src/lib/fab/fab.spec.ts
YonatanKra: This indeed looks like a non-intuitive API. Need to decide on this one...

yinonov:
A word on difference between buttons.
FAB has unique style and members which justifies its independence. Extending the fast foundation button is exactly where this fits

rachelbt:
> A word on difference between buttons. FAB has unique style and members which justifies its independence. Extending the fast foundation button is exactly where this fits

but the only difference is the "white" variation

rinaok:
> > A word on difference between buttons. FAB has unique style and members which justifies its independence. Extending the fast foundation button is exactly where this fits
> 
> but the only difference is the "white" variation

@rachelbt @yinonov 

I did it with alternate because the vwc-button has a default connotation which is primary. And there is no color white in connotation to override it. This is the only thing that differs the vwc-fab from the vwc-button.

libs/components/src/lib/fab/partials/variables.scss
```diff
@@ -0,0 +1,3 @@
+$icon-gap: --fab-icon-gap;
```
rachelbt, line 1: I don't think this file is needed anymore...

YonatanKra (dismissed):
LGTM

rachelbt (commented):
I don't think that all of the FAB buttons need the alternate - only the "primary" one need it. The cta, alert success looks off with wrong colors.

rinaok:
> I don't think that all of the FAB buttons need the alternate - only the "primary" one need it. The cta, alert success looks off with wrong colors.

@rachelbt fixed it :)

rachelbt (changes requested):
you need also to update the ui-test png :)

libs/components/src/lib/fab/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Connotation
+
+Set the `connotation` attribute to change the FAB's connotation.
+It accepts a subset of predefined values.
+
+- Type: `'cta'`
+- Default: `''`
+
+```html preview
+<vwc-fab icon='plus-line' connotation='cta'></vwc-fab>
```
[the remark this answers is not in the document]
rinaok: @rachelbt  Can you please check why the disable mode is not with a grey background?

libs/components/src/lib/fab/fab.scss
```diff
@@ -0,0 +1,101 @@
...
+	outline: 0 none;
+	vertical-align: middle;
+}
+
+.control:not([class*="connotation"]) {
+	background: var(#{theme-consts.$vvd-color-surface-4dp});
+}
+.control[class*="connotation"] {
+	background: var(#{connotation.$vvd-color-connotation});
+}
```
yinonov, line 42: missing `color`

yinonov (changes requested):

libs/components/src/lib/fab/fab.scss
```diff
@@ -0,0 +1,101 @@
...
+	#{variables.$block-size}: 48px;
+
+	position: relative;
+	display: inline-flex;
+	box-sizing: border-box;
+	align-items: center;
+	border: 0 none;
+	block-size: var(#{variables.$block-size});
+	border-radius: 24px;
+	filter: var(#{theme-consts.$vvd-shadow-surface-4dp});
```
yinonov, line 32: should have elevation changed on hover, active states

yinonov (changes requested):

libs/components/src/lib/fab/fab.scss
rinaok: @yinonov In vivid-2 the elevation changes only on active state: https://vivid.vonage.com/?path=/story/alpha-components-fab--basic&args=icon:store

libs/components/src/lib/fab/fab.scss
```diff
@@ -0,0 +1,105 @@
...
+	&:hover {
+		background: var(#{theme-consts.$vvd-color-neutral-10});
+	}
+	&:active{
+		@include elevation.get-elevation(8);
+
+		background: var(#{theme-consts.$vvd-color-neutral-30});
+	}
+}
+
```
rachelbt, line 74:
```suggestion
.control:not(:disabled) {
	@include elevation.get-elevation(4);
	cursor: pointer;

	&:not([class*="connotation"]) {
		&:hover {
			background: var(#{theme-consts.$vvd-color-neutral-10});
		}
		&:active{
			background: var(#{theme-consts.$vvd-color-neutral-30});
		}
	}
	&:active{
		@include elevation.get-elevation(8);
	}
}
```
this way the hover state from appearance will work and 79-84 is not needed :)

libs/components/src/lib/fab/fab.scss
```diff
@@ -0,0 +1,105 @@
...
+.control[class*="connotation"] {
+	background: var(#{connotation.$vvd-color-connotation});
+	cursor: pointer;
```
rachelbt, line 77:
```suggestion
	
```

libs/components/src/lib/fab/README.md
```diff
@@ -0,0 +1,67 @@
...
+</script>
+```
+
+## Icon
+
+Use the `icon` attribute to change the FAB's icon.
+You can choose an icon from the [icons gallery](https://icons.vivid.vonage.com).
+
+- Type: `String`
+- Default: `''`
```
rachelbt, line 17:
```suggestion
- Default: `undefined`
```

libs/components/src/lib/fab/README.md
```diff
@@ -0,0 +1,67 @@
...
+```html preview
+<vwc-fab icon='plus-line'></vwc-fab>
+```
+
+## Label
+
+Use the `label` attribute to add text to the FAB.
+
+- Type: `String`
+- Default: `''`
```
rachelbt, line 28:
```suggestion
- Default: `undefined`
```

yinonov (changes requested):

libs/components/src/lib/fab/fab.scss
```diff
@@ -0,0 +1,92 @@
...
+/* Icon */
+.icon {
+	font-size: 24px;
+
+	.icon-trailing & {
+		order: 1;
+		margin-inline-start: var(#{variables.$icon-gap});
+	}
+
+	:not(.icon-trailing) & {
```
rachelbt, line 83: this is not needed and also `margin-inline-start: var(#{variables.$icon-gap});` is not needed. 
They both can be replaced with adding to `.control` `column-gap: var(#{variables.$icon-gap});`|
and of-course defining its value there like the block size.
If there we will not support fab sizes than you can simply write the value and not the variable :)

libs/components/src/lib/fab/partials/variables.scss
```diff
@@ -0,0 +1,2 @@
...
+$block-size: --fab-block-size;
```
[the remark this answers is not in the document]
rinaok: @rachelbt The `--fab-icon-gap` should also be with `_`?

rinaok:
@yinonov Should we disable the focus?

yinonov (commented):
surface is hardly seen. is this correct? @jshenkman 
image

rachelbt:
> surface is hardly seen. is this correct? @jshenkman image

maybe we need to change the "canvas" mapping:
image
and use the surface instead of:
`basis: theme-consts.$vvd-color-canvas`
this:
`basis: theme-consts.$color-surface-4dp`

yinonov (approved):

PR #348, opened 2022-05-01
style(components): enforce extensions to align with esm

yinonov:
see reason why barrel file does not satisfy extension rule
https://github.com/import-js/eslint-plugin-import/issues/1920#issuecomment-719953032

------- edit
removing eslint import/extensions rule in favor of typescript moduleResolution "nodeNext"

see
https://github.com/microsoft/TypeScript/issues/42813#issuecomment-942633095


libs/components/.eslintrc.json
```diff
@@ -44,6 +44,7 @@
     // functional
     "compat/compat": "error",
     "import/prefer-default-export": "off",
+		"import/extensions": ["error", "always"],
     "class-methods-use-this": "off",
```
yinonov, line 48:
```suggestion
    "import/prefer-default-export": "off",
    "import/extensions": ["error", "always"],
    "class-methods-use-this": "off",
```

yinonov:
waiting on https://github.com/rollup/plugins/issues/1058#issuecomment-1114011903

Issue #343, opened 2022-04-28
[appearance]: naming color tones


rachelbt:
under the `connotation/_variables.scss` there are color mapping - we need to decide regarding the names:

starting option:

```
primary: (
  contrast: theme-consts.$vvd-color-neutral-90,
  distance-from-base-2: theme-consts.$vvd-color-neutral-80,
  distance-from-base-1: theme-consts.$vvd-color-neutral-90,
  base: theme-consts.$vvd-color-primary, //60 - connotation
  soft: theme-consts.$vvd-color-neutral-20,
  pale: theme-consts.$vvd-color-neutral-10,
),

```

PR #349, opened 2022-05-01
chore(components): fix outline tokens reference

yinonov:


libs/shared/src/lib/sass/mixins/appearance/_mixins.scss
```diff
@@ -10,18 +10,6 @@
...
-  text: 'text',
-  ghost: 'ghost',
-);
-
-//$appearance-set: filled ghost outlined;
-//$states: idle disabled hover;
-
-
 @function get-variable($type) {
   $prefix: --_appearance-color;
```
rachelbt, line 14: :)

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
```diff
@@ -82,7 +95,12 @@ $states-mapping: (
       outline: transaprent,
     ),
     outlined: (
-      text: var(#{theme-consts.$vvd-color-neutral-30}),
+      text: var(#{theme-consts.$vvd-color-neutral-30}-strong),
```
rachelbt, line 98: its disabled - not sure we need strong- not in outlined or duotone

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
```diff
@@ -74,26 +86,30 @@ $states-mapping: (
...
-      text: var(#{theme-consts.$vvd-color-neutral-30}),
+      text: var(#{theme-consts.$vvd-color-neutral-50}),
       fill: transaprent,
-      outline: var(#{theme-consts.$vvd-color-neutral-30}),
+      outline: var(#{theme-consts.$vvd-color-neutral-50}),
+    ),
+    duotone: (
+      text: var(#{theme-consts.$vvd-color-neutral-50}),
+      fill: transaprent,
+      outline: var(#{theme-consts.$vvd-color-neutral-50}-basis),
```
rachelbt, line 103: I think the basis here is not needed for the variable :)

libs/shared/src/lib/sass/mixins/connotation/_variables.scss
```diff
@@ -9,48 +9,55 @@ $compound-placeholder: '[compound]'; // will be replaced by actual compound
 $connotation-mapping: (
   // should map to colors e.g. purple-60
   primary: (
+    basis: theme-consts.$vvd-color-neutral,
     contrast: theme-consts.$vvd-color-neutral-90,
     fierce: theme-consts.$vvd-color-neutral-80,
-    strong: theme-consts.$vvd-color-neutral-90,
+    strong: theme-consts.$vvd-color-primary,
```
rachelbt, line 15: this causes the button-primary-filled to have no hover color change

rachelbt (commented):
I like the basis option :)
We should add it also to the colors I think.

yinonov:
> I like the basis option :)
> 
> We should add it also to the colors I think.
> 
> 

U mean tokens? We discussed adding the actual 60 suffix

rachelbt:
> > I like the basis option :)
> > We should add it also to the colors I think.
> 
> U mean tokens? We discussed adding the actual 60 suffix

I meant here:
https://vivid.deno.dev/designs/color
60 will be also great :)

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
yinonov:
```suggestion
      outline: var(#{theme-consts.$vvd-color-neutral-50}),
```

libs/shared/src/lib/sass/mixins/connotation/_variables.scss
yinonov: introduced a new key instead

yinonov:
@rachelbt I think we're already doing the renaming here. consider closing #343?

yinonov:
> > > I like the basis option :)
> > > We should add it also to the colors I think.
> > 
> > 
> > U mean tokens? We discussed adding the actual 60 suffix
> 
> I meant here: https://vivid.deno.dev/designs/color 60 will be also great :)

oh in that case, it's not a new token coming from figma. it's a self mapping we use "under the hood", not an actual token

rachelbt (approved):
I think that the mapping is still not fully "there" yet, but we have a different PR for that. Apart from that - all look good to me.

PR #353, opened 2022-05-02
feat(workspace): upgrade nx to v14

YonatanKra:


libs/components/.eslintrc.json
```diff
@@ -15,7 +15,7 @@
   ],
   "rules": {
     // lint
-    "@nrwl/nx/workspace/no-attribute-default-value": "error",
+    "@nrwl/nx/workspace/no-attribute-default-value": "off",
```
yinonov, line 18: really? 🤨

yinonov (changes requested):

libs/components/.eslintrc.json
YonatanKra: Mind taking a look at it? It is broken for some reason...
Try to activate it and see.

yinonov (approved):

Issue #381, opened 2022-05-11
Textfield type number


YonatanKra:
The type number is not supported by fast.
Diving deep into the topic shows input of type number is problematic.
[Here's an example for that](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).
We should discuss and define how numbers are treated.
## Suggestion 1: 

1. When type is number, use `type="text"` with the `inputmode` and a pattern match for numbers to prevent non-numeric values (or leave it as concern for the user?).
2. Increment/decrement can be done in 1's only (e.g. `step="1"`)
3. The buttons will do the increment and decrement.

Questions:
1. Do we go with this way of dealing with numer?
2. If so, do we support decimals? 

## Suggestion 2:
Fast solved it by creating [a special component for number-field](https://explore.fast.design/components/fast-number-field).  This can also be a venue to pursue.

There will probably be more questions as we dive deep into this one.

PR #382, opened 2022-05-12
feat(textfield): add the text field component

YonatanKra:
Note that `number-field` is still in debate #381 
Also need to setup the ui-tests and documentation


libs/components/src/lib/textfield/README.md
```diff
@@ -0,0 +1,107 @@
...
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+Add a `label` attribute to add label to the text field.
+
+```html preview
+    <vwc-textfield label="My Label" name="test-field"></vwc-textfield>
+```
+
+## Placeholder
```
yinonov, line 17: text field has a vast parity with members of a native input. maybe just describe that part and link to w3c/mdn

yinonov (changes requested):

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,116 @@
...
+import {Appearance, BlockSize, Shape} from '../enums';
+import type {Textfield} from './textfield';
+
+const getStateClasses = (textField: Textfield) => classNames(
+	['error', !!textField.errorValidationMessage],
+	['disabled', textField.disabled],
+	['active', !!textField.value],
+	['readonly', textField.readOnly],
+	['placeholder', !!textField.placeholder],
+	`density-${textField.density ?? BlockSize.Normal}`,
```
yinonov, line 17: no need to default to a class.
style defaults are handled by css & scss mixins
see
[[`shape-${shape}`, Boolean(shape)],](https://github.com/Vonage/vivid-3/blob/dac03d5be96241b8c0a3b722e4c5128c4d714449/libs/components/src/lib/badge/badge.template.ts#L14-L15)

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,116 @@
...
+	`appearance-${textField.appearance ?? Appearance.Outlined}`,
```
yinonov, line 18: same here

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,116 @@
...
+	`shape-${textField.shape ?? Shape.Rounded}`,
```
yinonov, line 19: same here

yinonov (changes requested):

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @param context
+ * @public
+ */
+export const TextfieldTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Textfield> = () => html<Textfield>`
+	<div class="root ${getStateClasses}">
+	  	${when(x => x.charCount && x.maxlength, renderCharCount())}
+			<div class="form-field">
```
yinonov, line 73: is this extra div required? I think we can manage styling without it. let me try?

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,116 @@
...
+ * The template for the {@link @microsoft/fast-foundation#Textfield} component.
+ *
+ * @param context
+ * @public
+ */
+export const TextfieldTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Textfield> = () => html<Textfield>`
+	<div class="root ${getStateClasses}">
```
yinonov, line 71: `root` is kind of wrong here. the actual root is the host. stick to `base`?

libs/components/src/lib/textfield/README.md
```diff
@@ -0,0 +1,107 @@
...
+    <vwc-textfield label="condensed" density="condensed"></vwc-textfield>
+    <vwc-textfield label="normal" density="normal"></vwc-textfield>
+</vwc-layout>
+```
+
+## Shape
+
+Use the `shape` attribute to change the button's edges.
+
+- Type: `'rounded'` | `'pill'`
```
yinonov, line 84: pill is still in doubt an need to be concluded

yinonov (changes requested):

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,120 @@
...
+	ElementDefinitionContext,
+	FoundationElementDefinition,
+} from '@microsoft/fast-foundation';
+import {classNames} from '@microsoft/fast-web-utilities';
+import { focusTemplateFactory } from '../../shared/patterns/focus';
+import {Appearance, BlockSize, Shape} from '../enums';
+import type {Textfield} from './textfield';
+
+const getStateClasses = (textField: Textfield) => classNames(
+	['error', !!textField.errorValidationMessage],
```
yinonov, line 13: not sure about performance but I use Boolean() in favor of legibility

yinonov (changes requested):
affix pattern should be added to enable decorative icons support

yinonov (commented):
why is value reflected? by fast's design?
image

libs/components/src/lib/textfield/textfield.template.ts
```diff
@@ -0,0 +1,120 @@
...
+	['disabled', textField.disabled],
+	['active', !!textField.value],
```
yinonov, line 15: why not `has-value`?

yinonov (changes requested):

libs/components/src/lib/textfield/textfield.template.ts
YonatanKra: No significat perf issue: https://jsben.ch/Iujqh
And a boolean would mean to "pollute" the class with another "get" as this one's a complex condition (dirty and invalid).

libs/components/src/lib/textfield/README.md
```diff
@@ -0,0 +1,130 @@
...
+You can validate the textfield like any other native `input`.  Here's an example using `pattern`:
+
+```html preview
+<vwc-textfield id="invalid-textfield" label="invalid" pattern="123" value="5"></vwc-textfield>
+<script>
+    document.getElementById("#invalid-textfield").dirtyValue = true;
+    document.getElementById("#invalid-textfield").validate();
+</script>
+```
+### Input types
```
yinonov, line 130: should add also disabled

yinonov (changes requested):

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,129 @@
...
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+```html preview
+    <vwc-text-field label="Helper text below"
+                   helper-text="Help me!!!">
+    </vwc-text-field>
+```
+
+### char-count
```
yinonov, line 59:
```suggestion
### Character Count
```

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Set the `value` attribute to set the default value for the text field. Setting the property on the element will not change the default value, but will change the value shown in the view as well as the submitted value in a form (imitating the native behavior).
+
+```html preview
+    <vwc-text-field label="With default value"
+                   value="5"
+    </vwc-text-field>
+```
+
+### Helper-text
```
yinonov, line 46:
```suggestion
### Helper Text
```

yinonov (changes requested):

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,129 @@
...
+Add a `label` attribute to add label to the text field.
+
+```html preview
+<form method="post" action="">
+
+    <vwc-text-field label="My Label" type="email" name="email" autocomplete="email"></vwc-text-field>
+</form>
+```
+
+### Placeholder
```
yinonov, line 24: this is a native input member. we should mention it as part of all proxied members and link to a mdn/w3c document

yinonov (changes requested):

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,129 @@
...
+You can validate the textfield like any other native `input`.  Here's an example using `pattern`. The value is 5, which is not of the pattern `123`.  In addition, we programatically `dirtied` the field and then called `validate` because the change was not done by a user.
+
+```html preview
+<vwc-text-field id="invalid-textfield" label="invalid" pattern="123" value="5"></vwc-text-field>
+<script>
+    document.getElementById("#invalid-textfield").dirtyValue = true;
+    document.getElementById("#invalid-textfield").validate();
+</script>
+```
+### Input types
```
yinonov: we should mention it as part of the proxied members and link to a mdn/w3c document

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,130 @@
...
+```html preview
+<vwc-text-field label="Helper text below" helper-text="Help me!!!"></vwc-text-field>
+```
+
+### Character Count
+
+- Type: `boolean`
+- Default: `false`
+
+Use the `char-count` attribute along with the `maxlength` attribute to show a character count.
```
yinonov, line 59: on 2nd thought, if `maxlength` is a prerequisite for `char-count`, maybe default should be true.
automatically shown unless author opts-out

yinonov (changes requested):

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,130 @@
...
+<vwc-text-field placeholder='ghost' appearance='ghost'></vwc-text-field>
+<vwc-text-field label='fieldset' appearance='fieldset'></vwc-text-field>
+```
+
+
+## Use cases
+
+### Validation
+
+You can validate the text field like any other native `input`.  Here's an example using `pattern`. The value is 5, which is not of the pattern `123`.  In addition, we programatically `dirtied` the field and then called `validate` because the change was not done by a user.
```
yinonov, line 121: u meant to put dirtied in quotes? if not, is it part of the API as implied by that format?

libs/components/src/lib/text-field/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+```html preview
+<vwc-text-field placeholder='ghost' appearance='ghost'></vwc-text-field>
+<vwc-text-field label='fieldset' appearance='fieldset'></vwc-text-field>
+```
+
+
+## Use cases
+
+### Validation
```
yinonov, line 119: validation is eager and show error whole user focused and still inputting. I believe our spec is to have a polite message when user blurs (@jshenkman)

yinonov (changes requested):

libs/components/src/lib/text-field/README.md
yinonov: I think u should refer to this guide
https://developer.mozilla.org/en-US/docs/web/css/:user-valid

explain that this the behavior of our text field is somewhat similar to the proposed `:user-valid`

yinonov (changes requested):

libs/components/src/lib/text-field/text-field.ts
```diff
@@ -0,0 +1,41 @@
...
+		mode: 'boolean'
+	}) charCount = false;
+	@attr density?: TextFieldDensity;
+	@attr appearance?: TextFieldAppearance;
+	@attr shape?: TextFieldShape;
+	@observable userValid = true;
+
+	@volatile
+	get errorValidationMessage() {
+		return this.userValid ? '' : this.validationMessage;
```
yinonov, line 29: could there be any way to trigger validity indication programmatically? this question was raised a couple of times.

not sure if it's a correct handling but I can think of a use-case where this is needed. WDYT?

rachelbt:
The input needs a min width - otherwise when placing the text-field inside flex it look like this:
image

rachelbt:
In Firefox the input doesn't stretch to fill the wrapper and the text is not aligned:
image

rachelbt:
@jshenkman 
in disabled state - I think that the label should also be grayed. 
image

jshenkman:
> @jshenkman in disabled state - I think that the label should also be grayed. image

Definitely! By design it is :) 
Screen Shot 2022-05-24 at 12 40 39
https://www.figma.com/file/r8OYLPmRmDqfLKt6KHdJDB/Vivid-UI-Kit---2.0?node-id=9637%3A49884

rachelbt:
I think we can add some accessibility to the input as in this article:
https://www.ovl.design/text/inclusive-inputs/

yinonov:
> @YonatanKra, @yinonov I think we can add some accessibility to the input as in this article: https://www.ovl.design/text/inclusive-inputs/

can you point to a specific section?

rachelbt:
> > @YonatanKra, @yinonov I think we can add some accessibility to the input as in this article: https://www.ovl.design/text/inclusive-inputs/
> 
> can you point to a specific section?

- **Identifying Input Purpose** - adding nave - we do :)
- **Marking required** (adding an extra span for required text & hiding it for for screen readers only).
- **The required attribute** 
- **Validation** - various aria labels

yinonov:
> > > @YonatanKra, @yinonov I think we can add some accessibility to the input as in this article: https://www.ovl.design/text/inclusive-inputs/
> > 
> > 
> > can you point to a specific section?
> 
> * **Identifying Input Purpose** - adding nave - we do :)
> * **Marking required** (adding an extra span for required text & hiding it for for screen readers only).
> * **The required attribute**
> * **Validation** - various aria labels

make sense but each has a discussion of its own. I suggest merging this MVP and start iterations

rachelbt:
> https://www.ovl.design/text/inclusive-inputs/

Opened an issue for that:
https://github.com/Vonage/vivid-3/issues/419

libs/components/src/lib/textfield/README.md
rinaok: @YonatanKra @YonatanKra  I also think that we should add a link

PR #383, opened 2022-05-12
refactor(components): rename size to density

yinonov:
rename size (for block size members) to density due to its being a reserved name in input element


YonatanKra (approved):
LGTM

PR #403, opened 2022-05-24
docs(components): fix language annotation in code block

yinonov:


yinonov:
@YonatanKra can we limit tests to only html language with the _preview_ attribute?
https://github.com/Vonage/vivid-3/blob/96c6e80b9ed34cd7acd2c51fcf6b9fd25a8df415/libs/components/src/visual-tests/visual-tests-utils.ts#L22-L23

rinaok (approved):

PR #404, opened 2022-05-25
refactor(header): migrate to fast element

rinaok:
### Changes:
1. Added a `fixed` attribute.
2. Added `heading` attribute instead of `slot`.

### Changes to consider:
1. Change the name to _header_ / _pageHeader_ / _globalHeader_


yinonov:
can we have an example in **side-drawer** with top-app-bar?

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,200 @@
...
+        background-color: var(--vvd-color-neutral-10);
+        margin:0;
+    }
+</style>
+<vwc-top-app-bar heading="Alternate Top App Bar" alternate></vwc-top-app-bar>
+```
+
+## Slots
+
+### Meta
```
yinonov, line 77: I'm not sure that's its right name

libs/components/src/lib/top-app-bar/top-app-bar.scss
```diff
@@ -0,0 +1,57 @@
...
+	width: 100%;
+	height: var(#{variables.$top-app-bar-height});
+	box-sizing: border-box;
+	justify-content: space-between;
+	padding: 8px 12px;
+	background-color: var(#{theme-consts.$vvd-color-canvas});
+	color: var(#{theme-consts.$vvd-color-on-canvas});
+	column-gap: 12px;
+	
+	&.shadow {
```
yinonov, line 26: elevated?

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,188 @@
...
+## Properties
+
+### Heading
+Add the `heading` attribute to set the heading text.
+
+- Type: `string`
+- Default: `undefined`
+
+```html preview full
+<style>
```
yinonov, line 18: don't see how this helps the demo. unecessary decoration

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,188 @@
...
+### Fixed
+Add the `fixed` attribute to set the top-app-bar's position to be fixed.
+
+- Type: `boolean`
+- Default: `false`
+
+```html preview full
+<style>
+    body {
+        background-color: var(--vvd-color-neutral-10);
```
yinonov, line 35: again, unecessary

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,188 @@
...
+<vwc-top-app-bar heading="Alternate Top App Bar" alternate></vwc-top-app-bar>
+```
+
+## Slots
+
+### Meta
+The `meta` slot is for action content next to the heading.
+
+```js
+<vwc-button slot="meta" icon="menu-line"></vwc-button>
```
yinonov, line 63: what is that for? it's no js

libs/components/src/lib/top-app-bar/README.md
rinaok: Screen Shot 2022-05-29 at 16 53 09

without it, it looks like this

libs/components/src/lib/top-app-bar/README.md
rinaok: Screen Shot 2022-05-29 at 16 53 43

with the background you can understand where is the top bar

libs/components/src/lib/top-app-bar/README.md
rinaok: it's an example for a slot. Should it be html? I copied it from the side-drawer: https://vivid.deno.dev/components/side-drawer#app-content
It's `js` there too

libs/components/src/lib/top-app-bar/README.md
rinaok: what is the right name?

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,188 @@
...
+    <vwc-side-drawer id="sideDrawer" open slot="app-content">
+        <div>
+            <vwc-sidenav-item href="#" text="1st level item" icon="home-line"></vwc-sidenav-item>
+            <p><vwc-text font-face="body-2-bold">SECTION TITLE</vwc-text></p>
+            <vwc-sidenav-item href="#" text="1st level item" icon="chat-line"></vwc-sidenav-item>
+            <vwc-sidenav-item href="#" text="1st level item" icon="chat-line"></vwc-sidenav-item>
+        </div>
+        <div slot="app-content">
+            <vwc-button id="button" shape="pill" icon='menu-solid'></vwc-button>
+            <vwc-text font-face="body-1">
```
yinonov, line 180: wrap with layout? it's a usage example so decoration is relevant

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,188 @@
...
+        <vwc-button icon="heart-solid"></vwc-button>
+    </div>
+</vwc-top-app-bar>
+```
+
+### App-content
+
+To add content aside the side-drawer, add a slot called `app-content`.
+
+```js
```
yinonov, line 109: why the duplication of source? and this ain't no JS

rinaok:
@yinonov There is only one thing left to do: choose a name instead of top-app-bar.

yinonov (changes requested):
that effect is quite unbelievable 😁
image

libs/components/src/lib/top-app-bar/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Z-index
+
+Controls the z-index of the top-app-bar.
+
+- `--top-app-bar-z-index`
+- Type: `Number`
+- Default: `7`
+
+### Height
```
yinonov, line 203:
```suggestion
### Block Size
```

yinonov (changes requested):

libs/components/src/lib/top-app-bar/README.md
yinonov: is it by spec to expose block size value as a wild card?

rinaok:
> a scrolled app-header (non fixed) should scroll back instantly when page scrolls back down. start by applying the following
> 
> ```css
> position: sticky;
> top: calc(-1 * var(--top-app-bar-height));
> ```

We've already tried that. @rachelbt , can you elaborate?

libs/components/src/lib/top-app-bar/README.md
rinaok: @jshenkman Should we expose the Top Bar's height to the user?

yinonov: I think it should be a `min-block-size` only. if content exceeds, the top-bar extends. naturally

libs/components/src/lib/top-app-bar/README.md
yinonov: can you provide a few from other libraries usages?

rinaok:
> that effect is quite unbelievable 😁 image

@yinonov Does it look good now?

libs/components/src/lib/top-app-bar/README.md
rinaok: @yinonov @rachelbt  The `top-app-bar-height` effects also on the `padding-top` of the `app-content`.
So how can I do it when the content exceeds?

libs/components/src/lib/top-app-bar/README.md
rinaok: @yinonov material calls this slot: **navigationIcon**

libs/components/src/lib/top-app-bar/README.md
rachelbt: if we are using padding - we can't use the min-height. We need a predefined size either by us or by the user

rachelbt:
> > a scrolled app-header (non fixed) should scroll back instantly when page scrolls back down. start by applying the following
> > ```css
> > position: sticky;
> > top: calc(-1 * var(--top-app-bar-height));
> > ```
> 
> We've already tried that. @rachelbt , can you elaborate?

I think its the mixture of shadow & light dom that makes sticky not properly work.
It can work id we "fix" the top-app-bar host to the screen with:
```
position: fixed;
inset: 0;
overflow-y: auto;
```
But I'm not sure this is a good idea...

libs/components/src/lib/top-app-bar/README.md
yinonov: right but it's often not what it contains

libs/components/src/lib/top-app-bar/README.md
rinaok: @yinonov It's not related to this PR, it should be fixed in a new PR cause it happens in all the components.


Screen Shot 2022-06-09 at 12 28 43

libs/components/src/lib/header/header.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+			window['pageYOffset'] = scrollDist;
+			window.dispatchEvent(new CustomEvent('scroll'));
+			await elementUpdated(element);
+		}
+
+		beforeEach(function () {
+			element.fixed = true;
+		});
+
+		it('should default to no elevation with no scrolling', async function () {
```
YonatanKra, line 76: I can't see any relation to scrolling in the test itself.
Maybe this is more accurate:
```suggestion
		it('should default to no elevation', async function () {
```

libs/components/src/lib/header/header.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+			await elementUpdated(element);
+			const control = getControlElement(element);
+			expect(element.elevated).toEqual(false);
+			expect(control.classList.contains('elevated')).toEqual(false);
+		});
+
+		it('should set "elevated" to true and add "elevated" class', async () => {
```
YonatanKra, line 83:
```suggestion
		it('should set "elevated" to true and add "elevated" class when scrolled', async () => {
```

libs/components/src/lib/header/header.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+		it('should remove elevated state when not fixed', async function () {
+			element.fixed = false;
+			const control = getControlElement(element);
+			await emulateScroll();
+			const hasClassElevated = control.classList.contains('elevated');
+
+			expect(hasClassElevated).toEqual(false);
+		});
+
+		it('should remove elevated state when offset is 0', async function () {
```
YonatanKra, line 103:
```suggestion
		it('should remove elevated state when scroll offset is 0', async function () {
```

libs/components/src/lib/header/README.md
```diff
@@ -0,0 +1,236 @@
...
+        </p>
+      </vwc-text>
+    </vwc-layout>
+  </div>
+</vwc-header>
+```
+
+## CSS Custom Properties
+
+### Z-index
```
yinonov, line 230:
```suggestion
### Stacking Context
```

libs/components/src/lib/header/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+Controls the z-index of the header.
```
yinonov, line 232: _"use the `--header-z-index` ..."_

libs/components/src/lib/header/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+- `--header-z-index`
```
yinonov, line 234:
```suggestion
```

libs/components/src/lib/header/header.template.ts
```diff
@@ -0,0 +1,41 @@
...
+	FoundationElementDefinition,
+} from '@microsoft/fast-foundation';
+import { classNames } from '@microsoft/fast-web-utilities';
+import type { header } from './header';
+
+const getClasses = ({ fixed, alternate, elevated }: header) => classNames(
+	'control',
+	['fixed', fixed],
+	['alternate', alternate],
+	['elevated', elevated],
```
yinonov, line 14:
```suggestion
	['fixed', Boolean(fixed)],
	['alternate', Boolean(alternate)],
	['elevated', Boolean(elevated)],
```

libs/components/src/lib/header/header.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export class header extends FoundationElement {
+	/**
+	 *
+	 *headerEl
+	 * @public
+	 *
+	 * HTML Attribute: heading
+	 */
+	@attr heading?: string;
```
yinonov, line 17: is this defined / required by design?

libs/components/src/lib/header/header.ts
```diff
@@ -0,0 +1,57 @@
...
+	 * sets the header to be fixed
+	 *
+	 * @public
+	 */
+	@attr({
+		mode: 'boolean',
+	}) fixed = false;
+
+	/**
+	 * applies scheme alternate region
```
yinonov, line 29:
```suggestion
	 * applies scheme alternate to header region
```

libs/components/src/lib/header/header.ts
```diff
@@ -0,0 +1,57 @@
...
+		window.addEventListener('scroll', this.#updateElevatedState);
+	}
+
+	override disconnectedCallback() {
+		super.disconnectedCallback();
+		window.removeEventListener('scroll', this.#updateElevatedState);
+	}
+
+	#updateElevatedState = () => {
+		this._elevated = !this._elevated;
```
yinonov, line 50: if not used in template then do `#elevated`

libs/components/src/lib/header/header.ts
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	@attr({
+		mode: 'boolean',
+	}) alternate = false;
+
+	@observable _elevated = false;
+
+	override connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('scroll', this.#updateElevatedState);
```
yinonov, line 41: this type of code will not comply to an SSR practice in future plans.
not saying no as there might not be another option. but asking, is there another option that will not throw in node run time?

libs/components/src/lib/header/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { designSystem } from '../../shared/design-system';
+import styles from './header.scss';
+
+import { header } from './header';
+import { headerTemplate as template } from './header.template';
+
+export const vividheader = header.compose<FoundationElementDefinition>({
```
yinonov, line 8:
```suggestion
export const vividHeader = Header.compose<FoundationElementDefinition>({
```

yinonov (changes requested):

libs/components/src/lib/header/README.md
rachelbt: I think this should come before the usage examples

rinaok: @yinonov Why z-index is not a good title?
Side-Drawer has this property: 

Screen Shot 2022-06-22 at 12 10 42

libs/components/src/lib/header/header.ts
rinaok: @yinonov I can't make an `@observable ` to be a private member.

libs/components/src/lib/header/header.ts
rinaok: @yinonov Card, Note, Calendar and Accordion have this attribute so I think that it should be consistent.

libs/components/src/lib/header/header.ts
rinaok: @yinonov How can I debug it?

libs/components/src/lib/header/README.md
yinonov: it's not bad, thought the suggestion is better but if doesn't ring right, resolve the comment

libs/components/src/lib/header/header.ts
yinonov: I think we should reconsider the way we define top text hierarchy. given that _heading_ should probably introduce an actual heading element (Hx) in template.

certain cases, like note, where Hx isn't wrapping the text, title might be a better member name

image

@rachelbt FYI. should we open this up?

libs/components/src/lib/side-drawer/side-drawer.scss
```diff
@@ -48,12 +48,14 @@ $inline-size: var(#{variables.$side-drawer-inline-size}, 280px);
...
-			::slotted(vwc-top-app-bar),
-			::slotted(vwc-top-app-bar-fixed) {
-				--mdc-top-app-bar-width: calc(100% - #{$inline-size});
-			}
+
+			// TODO: check
+			// @rinaok - this is a hack to get the drawer to be the same width as the app
+			// ::slotted(vwc-header) {
+			// 	--mdc-header-width: calc(100% - #{$inline-size});
+			// }
```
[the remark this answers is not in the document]
yinonov: no

libs/components/src/lib/header/header.ts
rinaok: @yinonov @rachelbt  Maybe you should create a new PR for that and change everything there.

libs/components/src/lib/header/ui.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+	extractHTMLBlocksFromReadme,
+	loadComponents,
+	loadTemplate,
+} from '../../visual-tests/visual-tests-utils.ts';
+
+const components = ['header', 'button', 'text', 'side-drawer', 'sidenav-item', 'layout' ];
+
+test('should show the component', async ({ page }: { page: Page }) => {
+	const template = extractHTMLBlocksFromReadme(
```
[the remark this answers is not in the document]
rinaok: @YonatanKra But what can I check there? window's pageYOffset? I check it in the unit tests

libs/components/src/lib/header/header.ts
yinonov: https://github.com/Vonage/vivid-3/issues/480

YonatanKra (approved):

PR #406, opened 2022-05-25
ci(components): visual regression redundant build

yinonov:
@YonatanKra isn't this run redundant? on update it does build components. but does it on check as well?


.github/workflows/_visual-regression.yml
```diff
@@ -14,7 +14,7 @@ jobs:
           node-version: '16'
           cache: 'npm'
       - run: npm ci
-      - run: npm run nx build components
+#       - run: npm run nx build components
```
yinonov, line 17:
```suggestion
```

YonatanKra (approved):

PR #411, opened 2022-05-27
fix(components): export missing components from package

yinonov:


libs/components/src/lib/components.ts
```diff
@@ -1,17 +1,23 @@
-export * from './text-anchor';
+export * from './accordion';
```
yinonov, line 1: not sure this is even required

libs/components/package.json
```diff
@@ -2,9 +2,29 @@
...
+    "./icon": "./icon",
+    "./layout": "./layout",
+    "./note": "./note",
+    "./popup": "./popup",
+    "./progress": "./progress",
+    "./progress-ring": "./progress-ring",
+    "./side-drawer": "./side-drawer",
+    "./sidenav-item": "./sidenav-item",
+    "./text": "./text",
+    "./tooltip": "./tooltip"
```
YonatanKra, line 28: Do we really need that? So maybe we should add this to the generator.
What does that help with?

yinonov: I linked the package to a dependent project and that made it work smoothly

YonatanKra (dismissed):

YonatanKra (approved):

Issue #415, opened 2022-05-29
fix(button): accessibility


rachelbt:
change `outline:none `to `outline:transparent`

Started reading with the focus.
Interesting to know that it is better to set elements with outline: transparent rather than outline: none, even if we are styling the focus state.
Why?
So that people that use high-contrast mode it set custom color to the focus state can use the native focus as they use to.

taken from [here](https://www.nicchan.me/blog/tips-for-focus-styles/#it's-probably-best-to-avoid-using-outline:-none-even-if-you-override-focus-styles)

yinonov:
I think we already apply the rule of high contrast

See @jshenkman's word on the topic

> Hey all,
Following our discussion about focus state I wanted to share some information and resources so we can all be on the same page and understand the importance of applying focus state to our system and products.

Focus state indicator
1. Focus states let users know which element they’re currently on and is ready to be interacted with.
2. Keyboard users typically navigate their way through Web sites by pressing the tab key. This allows them to jump from one interactive element on the page to another.
3. A keyboard user’s cursor equivalent is the focus indicator. By designing and implementing accessible focus indicators, we can make our products accessible to keyboard users, as well as users of assistive technology that works through a keyboard or emulates keyboard functionality

Accessibility 
1. A visible focus indicator is a requirement for a site to be considered accessible under the Web Content Accessibility Guidelines (WCAG).
2. According to https://www.w3.org/TR/WCAG22/#focus-visible, which states that any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.
3. Focus indicators on UI controls fall under the “non-text” components category, as they are used to identify a component’s state. To pass this criterion, our focus indicators must have a color contrast ratio of at least 3:1 against adjacent colors. Accordion to Success Criterion https://www.w3.org/TR/WCAG21/#non-text-contrast.
4. In WCAG 2.2, two success criteria were added to define how accessible a focus indicator is depending on both its color and area: https://www.w3.org/TR/WCAG22/#focus-appearance-minimum

Resources, references and examples
1. https://www.sarasoueidan.com/blog/focus-indicators/
2. https://m3.material.io/ - Implementation of focus state try moving around with the tab key
3. https://carbondesignsystem.com/components/button/style - Button styles, notice the focus state for each button type
4. https://spectrum.adobe.com/page/button/#Behaviors - Adds a blue focus ring
[...]

PR #417, opened 2022-05-29
fix(button): set focus outline to transparent

rachelbt:


yinonov (commented):
I recall it was mentioned in the article u shared. can u please describe how this helps?
consider we do have a dedicated element to reflect outline and we don't want this style to be visible on certain preferences

rachelbt:
> I recall it was mentioned in the article u shared. can u please describe how this helps? consider we do have a dedicated element to reflect outline and we don't want this style to be visible on certain preferences

all is in the issue:
https://github.com/Vonage/vivid-3/issues/415

YonatanKra:
What about other `outline: none` and `outline: 0 none` in our code?
Namely sidenav item, text-field etc.

rachelbt:
> What about other `outline: none` and `outline: 0 none` in our code? Namely sidenav item, text-field etc.

didn't want to do them both in the same issue, cause they are not the same.
Buttons actually has focus component inside them - so it may be that we can set focus none on buttons.

YonatanKra (approved):

rachelbt:
due to the face that we are setting a color of the outline  in the focus component that is inside button - I'm closing this isusse.

PR #426, opened 2022-05-31
chore(scripts): add scripts

rinaok:


yinonov (commented):
@YonatanKra objected this. lets hear is word on this

YonatanKra:
I don't mind actually. If it helps most then we can go for it.

package.json
```diff
@@ -3,9 +3,12 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
-    "start": "nx serve",
-    "build": "nx build",
-    "test": "nx test",
+    "start": "npx nx run docs:serve",
+    "build": "npx nx build docs && npx nx build components",
```
YonatanKra, line 7: I think components should be built first

YonatanKra (changes requested):

package.json
rinaok: @YonatanKra updated

yinonov: but why the redundancy? doesn't doc build components? it slows things down

YonatanKra: I think so. That's why these commands are redundant :)
I mean - what does this save you? Not much.
`npx nx build docs` => `npm run build`
Less declarative and and less `monorepo-ish` cause we have several builds. So what are you building? The components? the docs? the react demo app? the angular demo app?
The only thing I can agree upon is the `change` command which can be really annoying :)

rinaok: I don't mind about the build but `npm start` vs `npx nx run docs:serve` saves me a lot.

YonatanKra: @rinaok - ok, so let's remove the `build` command as it is redundant and leave `npm start` as an alias to the `docs:serve` command.
And another thing - if you are running commands from an `npm script` you don't need `npx` - just run `nx run docs:serve` and it'll work.
`npx` is when you are running from command line and unsure if you have the target library installed.

rinaok: @YonatanKra removed the build

YonatanKra (approved):

PR #429, opened 2022-06-02
refactor(components): focus ring visibility

yinonov:
show focus ring on keyboard navigation only


yinonov:
@rachelbt can we fix this in action-group somehow?
image

YonatanKra:
@yinonov 
It is a big PR. Is this still valid?
If so, need a few things here but mostly context:
What is the purpose of this PR? What's a focus ring? Where do you make sure it shows only on keyboard navigation?

yinonov:
> @yinonov It is a big PR. Is this still valid? If so, need a few things here but mostly context: What is the purpose of this PR? What's a focus ring? Where do you make sure it shows only on keyboard navigation?

@YonatanKra it has a dedicated pattern in css to show only on keyboard activation.

guess this should have its own test. in another PR?

jshenkman (approved):

PR #433, opened 2022-06-06
chore(workspace): fix deploy even with bump failure

YonatanKra:


.github/workflows/pull-request.yml
```diff
@@ -19,13 +19,24 @@ jobs:
...
           fetch-depth: 0
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
           cache: 'npm'
+
+      - name: Bump and Changelog
+        id: bump
+        run: |
+          npx beachball bump -m "Publishing a new version" --no-push -y > tmp.txt
```
yinonov, line 34: does this actually commit?

yinonov (changes requested):
it wont hurt but what's the reason behind "always deploy"

.github/workflows/pull-request.yml
YonatanKra: Don't look at it - I'm testing now. The PR file changes will be reverted.

yinonov (approved):

PR #434, opened 2022-06-07
fix(layout): docs

rinaok:


libs/components/src/lib/layout/README.md
```diff
@@ -41,6 +41,14 @@ Use the `gutters` attribute to set the initial preferred `margin`.
 - Default: `undefined`
 
 ```html preview full
+<style>
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 400px; 
+  }
```
yinonov, line 50: is this required?

libs/components/src/lib/layout/README.md
```diff
@@ -65,6 +73,14 @@ Use the `column-spacing` attribute to set the initial preferred `spacing` of a c
 - Default: `'medium'`
 
 ```html preview full
+<style>
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 300px; 
+  }
```
yinonov, line 82: is this required?

libs/components/src/lib/layout/README.md
```diff
@@ -135,6 +159,17 @@ With `fill`, everything is the same as `fit`, except empty tracks are not collap
 - Default: `'fit'`
 
 ```html preview full
+<style>
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 200px; 
+  }
```
yinonov, line 168: is this required?

libs/components/src/lib/layout/README.md
```diff
@@ -93,6 +109,14 @@ Use `block` to stack elements one above the other.
 - Default: `'medium'`
 
 ```html preview full
+<style>
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 1000px; 
+  }
```
yinonov, line 118: is this required?

libs/components/src/lib/layout/README.md
```diff
@@ -158,6 +193,17 @@ Controls the grid-template-columns of the layout.
 - Default: auto
 
 ```html preview full
+<style>
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 200px; 
+  }
```
yinonov, line 202: is this required?

libs/components/src/lib/layout/README.md
```diff
@@ -41,6 +41,14 @@ Use the `gutters` attribute to set the initial preferred `margin`.
...
+  vwc-card {
+    width: 100%;
+  }
+  html { 
+    block-size: 400px; 
+  }
+</style>
 <div class="demo-gradient">
   <vwc-layout>
     <vwc-card elevation="2" text="default"></vwc-card>
```
yinonov, line 54: adding redundant members, such as elevation here, only affect the looks and beautifies the docs but functionality-wise it reduces the readability and dev experience

yinonov (changes requested):
as mentioned before, I believe the gradient div & class doesn't belong inside the exposed doc.

in the future we should enable that from another preview attribute

libs/components/src/lib/layout/README.md
rinaok: @yinonov yes, it doesn't look good without it

libs/components/src/lib/layout/README.md
rachelbt: it's not elevation its card inside layout. 
There's no need for layout if there are no components inside it :)

rinaok: @yinonov  It was a div before and you requested that it be changed to a card component.

libs/components/src/lib/layout/README.md
rinaok: @yinonov yes, it doesn't look good without it

libs/components/src/lib/layout/README.md
rinaok: @yinonov yes, it doesn't look good without it

libs/components/src/lib/layout/README.md
rinaok: @yinonov yes, it doesn't look good without it

yinonov: I don't get it. why we need to apply size to html here?

rinaok: @yinonov What would be a better solution?

libs/components/src/lib/layout/README.md
rinaok: Screen Shot 2022-06-14 at 12 08 49

libs/components/src/lib/layout/README.md
rinaok: Screen Shot 2022-06-14 at 12 09 13

libs/components/src/lib/layout/README.md
rinaok: Screen Shot 2022-06-14 at 12 09 30

libs/components/src/lib/layout/README.md
rinaok: Screen Shot 2022-06-14 at 12 09 50

PR #444, opened 2022-06-09
fix(action-group): refine tests

rachelbt:


libs/components/src/lib/action-group/action-group.spec.ts
```diff
@@ -22,27 +22,26 @@ describe('vwc-action-group', () => {
 	});
 
 	describe('appearance', function () {
-		it('should set the fieldset class on the base', async function () {
+		it('should set the appearance-fieldset or ghost class on the base', async function () {
```
YonatanKra, line 25: What is `ghost` doing here? Is it a different use case?

YonatanKra (changes requested):

libs/components/src/lib/action-group/action-group.spec.ts
rachelbt: can't we just check weather theres an appearance class?
fieldset & ghost are classes related to appearance.
Same for shape - it can be either none defined - than it is rounded and can be set to pill

YonatanKra: The unit test shouldn't care about appearance values. That's our concern. The logic is to just reflect the value in the class (see example [here](https://github.com/Vonage/vivid-3/blob/main/libs/components/src/lib/banner/banner.spec.ts#L214)).

rachelbt: so the test it OK but the description should be: 
`should set the appearance class on the base`
?

YonatanKra: Yeap!

YonatanKra (approved):

PR #452, opened 2022-06-09
fix(text-field): firefox stretch and text alignment

rachelbt:


YonatanKra (commented):
@rachelbt - how come this is related to Safari but the changes in visuals is in Firefox?

rachelbt:
> @rachelbt - how come this is related to Safari but the changes in visuals is in Firefox?

It was from the start a FF problem.... I'll fix the PR name :)

PR #454, opened 2022-06-10
fix(text-field): make sure validation is not eager

YonatanKra:


libs/components/src/lib/calendar-event/ui.test.ts
```diff
@@ -7,7 +7,7 @@ import {
 
 const components = ['calendar', 'calendar-event'];
 
-test.only('should show the component', async ({ page }: { page: Page }) => {
+test('should show the component', async ({ page }: { page: Page }) => {
```
yinonov, line 10: sorry 😬

libs/components/src/lib/text-field/text-field.ts
```diff
@@ -24,18 +24,31 @@ export class TextField extends FoundationTextfield {
...
+			this.#blurred = false;
+		});
+	}
+
+	override validate = () => {
 		super.validate();
+		this.userValid = !this.userValid;
 		if (this.proxy instanceof HTMLElement) {
-			this.userValid = this.dirtyValue ? !this.validationMessage : true;
+			this.userValid = (this.#blurred && this.dirtyValue) ? !this.validationMessage : true;
```
yinonov, line 49: this assigns it wrong according to https://developer.mozilla.org/en-US/docs/web/css/:user-valid.
I think the composite condition should be assigned to a `reflectInvalid` flag instead

yinonov (changes requested):

libs/components/src/lib/text-field/text-field.ts
YonatanKra: @yinonov - this actually follows the `user-valid` spec as we discussed.

yinonov: yes we discussed it after this comment. resolving

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -355,22 +365,27 @@ describe('vwc-text-field', () => {
 			element.validate();
 		}
 
-		it('should add class error to root', async function () {
+		it('should add class error to base if invalid', async function () {
```
yinonov, line 368: maybe the intent is wrong. maybe it should be adding `user-not-valid` class to be completely correct

yinonov: notice it is `not-valid` rather than `invalid`

YonatanKra: @yinonov 
Looked a bit deeper.
Do we really need the `user-not-valid` class?
We already have the semantic `error` error class that speaks our css design system language.

PR #459, opened 2022-06-13
fix(text-field): default 20ch inline-size

rachelbt:


rachelbt:
was fixed here: 
https://github.com/Vonage/vivid-3/pull/472

PR #463, opened 2022-06-14
chore(components): fab normal size

yinonov:


rachelbt (approved):

rinaok (changes requested):
Why did you change the size? Fab's size should be 48px. The mini Fab is 40px.

rinaok (approved):

PR #468, opened 2022-06-17
ci(components): viewport size to improve ui tests performance

yinonov:


rachelbt (commented):
NICE!!
there are 2 ong's that are missing:
calendar-event: Safari
popup: FF

yinonov:
> NICE!! there are 2 ong's that are missing: calendar-event: Safari popup: FF

not sure why they are missing. perhaps I just hit their exact width. up until now the viewports were different for each browser

rachelbt (approved):

PR #475, opened 2022-06-22
chore(docs): note on icon-only a11y pitfall

yinonov:
- [ ] apply for FAB
- [ ] apply for Button

question to answer before merge -
will this benefit authors given i18n taken into account? icons could have a suprising language (english) for some users, as icons are named in english only.
Also, icon name might not add any meaninful context to users. authors using automated a11y checks wont discovers missing proper labels for icon only buttons, which will result a feature to become a pitfall


yinonov:
alternatively, maybe just add note in docs as pit fall on icon only. that aria-label/labelledby must be provided

PR #497, opened 2022-06-27
docs(side-drawer): add an example with the header component

rinaok:


libs/components/src/lib/side-drawer/README.md
```diff
@@ -294,6 +294,54 @@ To add content aside the side-drawer, add a slot called `app-content`.
 </script>
 ```
 
+## Example Usage
+
+### Side-drawer wrapping a [Header](../header)
```
rachelbt, line 299: I know this is the correct hierarchy but the text looks really big :(

rachelbt (commented):
- the links in both readme doesn't work.
- There's something strange in the background of the code preview when scrolling - half is transparent and half has white bg.

rinaok:
> * the links in both readme doesn't work.
> * There's something strange in the background of the code preview when scrolling - half is transparent and half has white bg:
>   image

@rachelbt The links work in localhost, so it's strange.

libs/components/src/lib/side-drawer/README.md
yinonov: what do u suggest? keeping the text short? or resizing the font size?

yinonov:
@AyalaBu u might wanna join in the review

apps/docs/assets/styles/main.css
```diff
@@ -36,6 +36,7 @@ body a:visited {
 vwc-side-drawer {
   display: block;
   width: 100%;
+  height: 100%;
```
rinaok, line 39: @yinonov It fixed the mobile view issue. Send me a screenshot if you see any other issues.

yinonov (changes requested):
part of the aside body in now hidden. notice the design system heading is not shown
image

rinaok:
> part of the aside body in now hidden. notice the design system heading is not shown image

@yinonov changed it to be: calc(100% - 64px)

apps/docs/assets/styles/main.css
rinaok: @yinonov changed it to: `height: calc(100% - 64px)`

yinonov (changes requested):

libs/components/src/lib/header/README.md
```diff
@@ -133,9 +133,9 @@ To add content below the header, add a slot called `app-content`.
 
 ## Example Usage
 
-### Fixed Header With Side-Drawer
+### Header wrapping a [Side-drawer](../../components/side-drawer)
```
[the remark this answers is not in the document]
rinaok: @yinonov  This is the title? `Header as Primary`

yinonov:
think this should be closed considering #502?

PR #498, opened 2022-06-27
refactor(components): change ts type and css property

yinonov:


yinonov:
@rachelbt think we can contain body scroll as well in modal mode?
(can reproduce setting the docs website's drawer to modal)

rachelbt:
> @rachelbt think we can contain body scroll as well in modal mode? (can reproduce setting the docs website's drawer to modal)
```

background-color: var(--vvd-color-on-canvas);
position: fixed;
inset: 0px;
color: wheat;
pointer-events: all;
overscroll-behavior: contain;
opacity: 0.5;
overflow: overlay;
```
don’t forget to remove the inert:
`<div class="side-drawer-app-content" inert="">`

yinonov:
this solution only works if overlay is scrollable. If all content fit on screen within an element setting `overscroll-behavior: contain`, but there is no scroll inside — it will not stop the body scroll

yinonov:
handled better in #527

PR #502, opened 2022-06-29
refactor(components): side-drawer & header example affected API changes

yinonov:


yinonov:
wrong interface causes - 
scrolled document to set elevation on header
image

while aside body scrolled does not
image

rinaok:
Screen Shot 2022-06-30 at 15 24 22


alternate is not working

rinaok:
Screen Shot 2022-06-30 at 15 28 35

Position="end" is broken

libs/components/src/lib/header/header.spec.ts
```diff
@@ -16,105 +18,36 @@ describe('vwc-header', () => {
...
-			element.fixed = true;
-			await elementUpdated(element);
-			expect(hasClassFixedOnInit).toEqual(false);
+	describe('elevation shadow', () => {
+		it('should default elevation to no-shadow', async () => {
+			const elevation = element.shadowRoot?.querySelector(ELEVATION_SELECTOR) as Elevation;
 
-			const hasClassFixed = control.classList.contains('fixed');
-			expect(hasClassFixed).toEqual(true);
+			expect(elevation.noShadow).toBeTruthy();
```
YonatanKra, line 30: Truthy can be almost everything. You expect this to equal `true`.
```suggestion
			expect(elevation.noShadow).toEqual(true);
```

libs/components/src/lib/header/header.spec.ts
```diff
@@ -16,105 +18,36 @@ describe('vwc-header', () => {
...
 
-			expect(hasClassElevated).toEqual(false);
-		});
-
-		it('should remove scroll listener after diconnection', async function () {
-			const originalElevationState = element._elevated;
-			element.disconnectedCallback();
-			await emulateScroll();
-			expect(element._elevated).toEqual(originalElevationState);
+			expect(container.getAttribute('part')).toEqual('vvd-theme-alternate');
```
YonatanKra, line 50: Are we doing `part`s now?

libs/components/src/lib/header/header.spec.ts
yinonov: dont mind but it should only be other than `undefined`

libs/components/src/lib/header/README.md
```diff
@@ -1,235 +1,245 @@
...
+<vwc-header>
+  Customized header element style
+</vwc-header>
+```
+
+## Usage Examples
+
+### Fixed Header With Side-Drawer
+
+A *fixed* Header with the *default*, `actionItems` and `app-content` slots.
```
rachelbt, line 146: there are no `actionItems  in the code....

libs/components/src/lib/header/README.md
```diff
@@ -1,235 +1,245 @@
...
+  </main>
 </vwc-header>
 ```
 
-## Example Usage
+## CSS Custom Properties
 
-### Fixed Header With Side-Drawer
+### Block Size
+
```
rachelbt, line 87: if its "read-only" why are we pointing it out?
How does one do change the header height?

libs/components/src/lib/side-drawer/side-drawer.scss
```diff
@@ -5,105 +5,59 @@
...
-	block-size: inherit;
-	color: var(#{variables.$side-drawer-color}, var(#{theme-consts.$vvd-color-on-canvas}));
-	inline-size: #{$inline-size};
+	background-color: var(#{theme-consts.$vvd-color-canvas});
+
+	// block-size: 100vh;
+	color: var(#{theme-consts.$vvd-color-on-canvas});
+	inline-size: $inline-size;
+	inset-block: 0;
+	overflow-y: auto;
```
rachelbt, line 20:
```suggestion
&.modal {
  z-index: 1;
}
```

in modal mode we need to add z-index to prevent this:
image

libs/components/src/lib/side-drawer/README.md
```diff
@@ -18,13 +18,9 @@ You can also close the side drawer by pressing the `ESC` key.
 
 ```html preview full
```
rachelbt, line 19: in the samples - the nav items in the nav are lacking spaces. I think that if we are wrapping the content with layout we can do it also to the nav for 2 reasons:
1. the nav in the side-drawer looks not good and it looks a bit broken
2. if one copy paste it for using with navigation than using layout for spaces are a good idea :)

libs/components/src/lib/side-drawer/README.md
```diff
@@ -250,96 +200,62 @@ To add a header to the side drawer, add a slot called `header`.
...
 
-### App-content
+## CSS Custom Properties
+
+### Application Content Offset
 
-To add content aside the side-drawer, add a slot called `app-content`.
+`--side-drawer-app-content-offset` controls the offset of the side drawer's app-content from the window's edge.
+some designs may choose side-drawer to overlap the app-content, so the app-content should be offset by the side-drawer's width.
+Additionally, as aside element (which represents the actual side-drawer), is styled with `position: fixed`, customizing its inline size directly will not affect the application content offset interchangeably. Hence, using this CSS custom property is mandatory to account for side-drawer inline size altercations.
```
rachelbt, line 209:
```suggestion
`--side-drawer-app-content-offset` controls the offset of the side drawer's app-content from the window's edge.  
some designs may choose side-drawer to overlap the app-content, so the app-content should be offset by the side-drawer's width.  
Additionally, as aside element (which represents the actual side-drawer), is styled with `position: fixed`,  
customizing its inline size directly will not affect the application content offset interchangeably. Hence, using this CSS custom property is mandatory to account for side-drawer inline size altercations.  
```
added spaces for line breaking

rachelbt (commented):
👍

libs/components/src/lib/header/README.md
yinonov: it doesn't change. it just helps consumers use a token of block size, rather than hardcode a magic string of `64px`. if header size is to be considered in application element, this token can be used (e.g. fixed header requires applying a margin to application content wrapper)

libs/components/src/lib/side-drawer/README.md
```diff
@@ -6,340 +6,203 @@
...
-- Type: `Number`
-- Default: `280px`
+<vwc-side-drawer id="sidedrawer">
 
-Controls the inline size of the side drawer.
+  <vwc-layout slot="app-content" gutters="medium">
+    <vwc-text>
+      Toggle the side drawer by clicking the FAB.
+      <br>
+      Notice that the side drawer overlaps the application content.
```
rachelbt, line 177: and it not that good...
image

also - users don't expect to find the open close drawer button at the bottom of the navigation...

libs/components/src/lib/header/README.md
```diff
@@ -1,235 +1,299 @@
...
-      </p>
-    </vwc-text>
-  </vwc-layout>
-</vwc-header>
-```
-
-### Fixed
-
-Add the `fixed` attribute to set the header's position to be fixed.
+Header is an elevated element and applies elevation surface background color (more noticeable in **dark** themes) but leaves out elevation shadow (more noticeable in **light** themes) for author's to opt-in.
```
rachelbt, line 15: the `vwc-text `for body bold is set to be block so there's an unnecessary line break...
image

libs/components/src/lib/header/header.template.ts
```diff
@@ -23,19 +21,26 @@ const getClasses = ({ fixed, alternate, elevated }: Header) => classNames(
...
-        <section class="header-content end" id="actions" role="toolbar">
-          <slot name="actionItems"></slot>
-        </section>
-    </header>
+    <${elevationTag} dp="4" ?no-shadow=${x => !x.elevationShadow}>
+      <header class="base" part="base">
+        <!-- a container is needed to distinguish the surface background color of the
+        element from its shadow when applying elevation with alternate -->
+        <div class="container" part="${getPartAlternate}">
+          <section class="header-content start">
```
rachelbt, line 33:
```suggestion
          <section class="header-content">
```
you removed the classes in css - so no use for it here :)

libs/components/src/lib/header/header.template.ts
```diff
@@ -23,19 +21,26 @@ const getClasses = ({ fixed, alternate, elevated }: Header) => classNames(
...
+            <slot></slot>
+          </section>
+          <section class="header-content end" id="actions" role="toolbar">
```
rachelbt, line 36:
```suggestion
          <section class="header-content" id="actions" role="toolbar">
```

libs/components/src/lib/header/partials/variables.scss
```diff
@@ -1 +1 @@
-$header-z-index: --header-z-index;
\ No newline at end of file
+$header-block-size: --_header-block-size; // cannot be set as API but can be used in nodes assigned to app-content slot
```
rachelbt, line 1: in side-drawer you changed the name of the file to _variables. I guess this one need to be changed as well

libs/components/src/lib/side-drawer/side-drawer.scss
```diff
@@ -2,108 +2,65 @@
...
-	overflow: auto;
-	flex-direction: column;
-	flex-shrink: 0;
-	background-color: var(#{variables.$side-drawer-background-color}, var(#{theme-consts.$vvd-color-canvas}));
-	block-size: inherit;
-	color: var(#{variables.$side-drawer-color}, var(#{theme-consts.$vvd-color-on-canvas}));
-	inline-size: #{$inline-size};
+	background-color: var(#{theme-consts.$vvd-color-canvas});
+
+	// block-size: 100vh;
```
rachelbt, line 14:
```suggestion
```

apps/docs/_layouts/base.njk
```diff
@@ -50,37 +51,85 @@
...
+      </div>
+    </vwc-header>
+
+    <vwc-layout gutters="small" column-basis="block">
       {% include "navigation.njk" %}
-			<main slot="app-content">
+    </vwc-layout>
+
+    <vwc-header id="header-main" slot="app-content">
+      <vwc-button slot="actionItems" label="View on GitHub" icon="github-mono" icon-trailing href="https://github.com/vonage/vivid-3"></vwc-button>
```
rachelbt, line 111: this is a link - it leads the user to a new URL, it's not semantic to use in with a button and if we are keeping the button it will not work with href but with on click...

rachelbt (commented):
Looking really good and promising in easier usage of the components.
About the default slot name - I'm wondering about the default - we ditched using this in other components. Not sure main will suit here though....

apps/docs/_layouts/base.njk
yinonov: ure right, I plan on exposing the anchor PR to follow right after this is merged. will also remove the star button

libs/components/src/lib/header/README.md
yinonov: true. one of the reasons to progress on #507 
it's not in the scope of this PR to resolve. it's a side effect of text and markdown conversion

libs/components/src/lib/side-drawer/README.md
yinonov: that's the most recent design request. I agree it would need another iteration

yinonov:
> Looking really good and promising in easier usage of the components. About the default slot name - I'm wondering about the default - we ditched using this in other components. Not sure main will suit here though....

if you recall, main was a necessity due to technological issues. we favor default where possible

rachelbt (approved):

PR #503, opened 2022-06-30
refactor(sidenav-disclosure): add new component

rinaok:


yinonov (commented):
looking great!

that's how the a11y tree look
image

honestly this doesn't violate anything as far as I can tell, but there should be some research on whether this is a good UX. can we extract some data? ask around? see other implementations? go over it with SR?

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.template.ts
```diff
@@ -0,0 +1,50 @@
...
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Disclosure> = (
+	context: ElementDefinitionContext,
+) => {
+	const affixIconTemplate = affixIconTemplateFactory(context);
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html`<details class="disclosure" ${ref('details')}>
+        <summary class="${getClasses}"
```
yinonov, line 33: noticed the exact style as _sidenav-item_ is now set on this summary tag. was it a small chunk of style? since we already have that same style on sidenav-item, do we want to extract a pattern here?

rinaok: @yinonov It's the same style cause the first item (the button) should look and feel the same as it's children (the sidenav-items). I copied it from sidenav-item's control class. It is the same style in two components, do you think it's something worth to extract?

yinonov: yes, I think this could be used in options, list items etc'.

rinaok:
> looking great!
> 
> that's how the a11y tree look image
> 
> honestly this doesn't violate anything as far as I can tell, but there should be some research on whether this is a good UX. can we extract some data? ask around? see other implementations? go over it with SR?

@yinonov 

All the sidenavs I've found use lists as their implementation of this. Even [Fast](https://github.com/microsoft/fast/blob/5b313a3262b977acbee3c2efac87d0d14a82f792/sites/fast-website/src/app/components/side-navigation/side-navigation.template.ts). 
As far as I can tell, I haven't found a single example with summary and details.

In other design systems, summary details are simply a component but they don't use it inside other components.

libs/components/src/lib/sidenav-disclosure/README.md
```diff
@@ -0,0 +1,163 @@
...
+<vwc-sidenav-item href="#" text="1st level item"></vwc-sidenav-item>
+<vwc-sidenav-disclosure label="1st level item">
+    <vwc-sidenav-item href="#" text="2nd level item"></vwc-sidenav-item>
+    <vwc-sidenav-item href="#" text="2nd level item"></vwc-sidenav-item>
+    <vwc-sidenav-disclosure label="2nd level item">
+        <vwc-sidenav-item href="#" text="3rd level item"></vwc-sidenav-item>
+        <vwc-sidenav-item href="#" text="3rd level item"></vwc-sidenav-item>
+        <vwc-sidenav-item href="#" text="3rd level item"></vwc-sidenav-item>
+    </vwc-sidenav-disclosure>
+</vwc-sidenav-disclosure>
```
yinonov, line 56: there's so much noise here. plainly focus on showing the label

yinonov: same goes for the rest of the sections

libs/components/src/lib/sidenav-disclosure/index.ts
```diff
@@ -0,0 +1,18 @@
+import '../icon';
+import '../focus';
+
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { designSystem } from '../../shared/design-system';
+import styles from './sidenav-disclosure.scss';
+
+import { Disclosure } from './sidenav-disclosure';
```
yinonov, line 8:
```suggestion
import { SidenavDisclosure } from './sidenav-disclosure';
```

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.template.ts
```diff
@@ -0,0 +1,50 @@
...
+            class="invoker"
+            role="button"
+            aria-controls="disclosure-content"
+            aria-expanded="${x => x.expanded}"
+        >
+            <slot name="start">${x => affixIconTemplate(x.icon)}</slot>
```
yinonov, line 39: we don't want to allow this

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.template.ts
```diff
@@ -0,0 +1,50 @@
...
+            <slot name="summary">${x => x.label}</slot>
+            <slot name="end">
```
yinonov, line 41: or this

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.ts
```diff
@@ -0,0 +1,22 @@
...
+import { applyMixins } from '@microsoft/fast-foundation';
+import { AffixIcon } from '../../shared/patterns/affix';
+
+/**
+ * A Sidenav Item Custom HTML Element.
+ * Based largely on the {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a | <a> element }.
+ *
+ * @public
+ */
+export class Disclosure extends FoundationDisclosure {
```
yinonov, line 12:
```suggestion
export class SidenavDisclosure extends FoundationDisclosure {
```

yinonov: I dont think we need to extend foundation disclosure. it adds nothing as it rely itself on html disclosure

yinonov (changes requested):

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
```diff
@@ -0,0 +1,85 @@
...
+	&.icon-only {
+		display: flex;
+		block-size: $min-block-size;
+		inline-size: $min-block-size;
+		place-content: center;
+	}
+
+	.toggleIcon {
+		position: absolute;
+		right: 16px;
```
rachelbt, line 60: why absolute? Its parent is flex- we can set it to be on the right like this:
```suggestion
		margin-inline-start: auto;
```

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.control {
+	$min-block-size: 40px;
+	$gap: 8px;
+
+	@include typography.typography-cat-shorthand("body-2");
+	@include appearance.appearance;
+	@include connotation.connotation;
+
+	position: relative;
```
rachelbt, line 26:
```suggestion

```
if we remove the position absolute for the chevron we don't need this

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
```diff
@@ -0,0 +1,85 @@
...
+	display: inline-flex;
```
rachelbt, line 27:
```suggestion
	display: flex;
```

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
```diff
@@ -0,0 +1,85 @@
...
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	border-radius: 6px;
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	color: var(#{appearance.get-appearance-token(text)});
+	cursor: pointer;
+	gap: $gap;
+	hyphens: auto;
+	inline-size: 100%;
+	min-block-size: $min-block-size;
+	padding-block: 10px;
```
rachelbt, line 39:
```suggestion

```
I'm not sure this is needed because we are setting height to the row and aligning its children to be centered

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
```diff
@@ -0,0 +1,85 @@
...
+			outline: none;
+		}
+	}
+
+	&.icon-only {
+		display: flex;
+		block-size: $min-block-size;
+		inline-size: $min-block-size;
+		place-content: center;
+	}
```
rachelbt, line 56:
```suggestion

```
this one is linked to nothing so has no effect on nothing. I don't think it is needed at all.

rachelbt (commented):
looks good :)
Maybe we can consider adding animation for the open close details content.
https://css-tricks.com/how-to-animate-the-details-element/
There are sone only-css options in the comments.
Maybe its for a separate PR...

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
rachelbt: I don't see a chevron there...

rinaok: @rachelbt Right, not the chevron but all your other comments

rinaok:
> looks good :) Maybe we can consider adding animation for the open close details content. https://css-tricks.com/how-to-animate-the-details-element/ There are sone only-css options in the comments. Maybe its for a separate PR...

@yinonov Said that we can add it later

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.scss
rinaok: @rachelbt When I removed it, the focus became on the details tag rather than the summary tag.

rinaok: Screen Shot 2022-07-14 at 14 11 49

change/@vonage-vivid-7799b301-8f1d-4b08-8bef-001a8cbd785c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
YonatanKra, line 2: It is a `prerelease` as it adds a new component/feature.

libs/components/src/lib/sidenav-disclosure/README.md
```diff
@@ -0,0 +1,78 @@
...
+The arrow points down when the controlled content is hidden, indicating that pressing the button will reveal additional content.
+The arrow points up when the content is visible.
+
+```js
+<script type="module">
+    import '@vonage/vivid/sidenav-disclosure';
+</script>
+```
+
+## Members
```
YonatanKra, line 14: I'm missing real-life examples using the side-nav with content. Can you add one?
In addition, this cool little component works with indefinite levels - so one can create a nested nav system:
```
<vwc-sidenav>
    <vwc-sidenav-disclosure label="1st level item">
<vwc-sidenav-disclosure label="Thirst level item">
        <vwc-sidenav-item href="#" text="2nd level item"></vwc-sidenav-item>
 </vwc-sidenav-disclosure>
    </vwc-sidenav-disclosure>
<vwc-sidenav>
</vwc-sidenav></vwc-sidenav>
```
 is that a use case? @jshenkman ?

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+			expect(element).toBeInstanceOf(SidenavDisclosure);
+			expect(element.label).toEqual(undefined);
+			expect(element.icon).toBeUndefined();
+			expect(element.open).toBeFalsy();
+		});
+	});
+
+	describe('open', () => {
+		it('should open the sidenav disclosure', async () => {
+			expect(getBaseElement(element).attributes.getNamedItem('open')).toBeFalsy();
```
YonatanKra, line 29: Is this the same as writing:
`getBaseElement(element).hasAttribute('open')`?
If so, we should apply the simpler code in favor of the more complex one.
Also apply to rest of the code.

YonatanKra (changes requested):

libs/components/src/lib/sidenav-disclosure/README.md
rinaok: @YonatanKra There were real-life examples of nested sidenavs. However, Yinon prefers examples with few contents. The sidenav readme includes an example of this.

YonatanKra (approved):
LGTM

PR #505, opened 2022-06-30
feat(dialog): migrate the dialog component

YonatanKra:


libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,9 @@
+# dialog
+
+```js
+<script type="module">import '@vonage/vivid/dialog';</script>
+```
+
+```html preview
```
yinonov, line 7: use this practice to control the iframe block size
https://github.com/Vonage/vivid-3/blob/6661b5639a9aa553758c1359fce94b531442056b/libs/components/src/lib/header/README.md#L19-L23

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function icon() {
+	return html<Dialog>`
+		<vwc-icon class="icon" type="${x => x.icon}"></vwc-icon>
+	`;
+}
+
+function heading() {
+	return html<Dialog>`
+		<vwc-text font-face="subtitle-2" class="heading">
```
yinonov, line 22: within components it's better to address typography by css only
text component have issues anyway and might be discouraged
see #507

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -0,0 +1,35 @@
...
+export class Dialog extends FoundationElement {
+	/**
+	 * Indicates dialog's state
+	 *
+	 * @public
+	 * @remarks
+	 * HTML Attribute: open
+	 */
+	@attr({mode: 'boolean'}) open = false;
+	@attr({mode: 'boolean'}) stacked = false;
```
yinonov, line 18: what's that about?

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -0,0 +1,35 @@
...
+	@attr({attribute: 'return-value'}) returnValue?: string;
```
yinonov, line 19: is that by native spec?

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -0,0 +1,35 @@
...
+	@attr icon?: string;
+	@attr content?: string;
+	@attr heading?: string;
```
yinonov, line 22: remember to keep it close to card practice with props and wrapping main slot.
I'd dare an say it's even a pattern now

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	close() {
+		if (!this.open) {
+			return;
+		}
+		this.open = false;
+		this.dispatchEvent(new CustomEvent('close', {bubbles: true, composed: true, detail: this.returnValue}));
```
yinonov, line 29: we wanna prefix that event name? maybe we can have a shared file that holds all event consts...

yinonov (changes requested):

libs/components/src/lib/dialog/dialog.ts
YonatanKra: See Figma - footer has stacked and non stacked forms

libs/components/src/lib/dialog/dialog.ts
YonatanKra: Yes

libs/components/src/lib/dialog/dialog.ts
YonatanKra: so `vwc:dialog:close`? Or just `vwc:close`?

change/@vonage-vivid-86f6e2a8-1642-47c7-a3d5-ec0ba3e525a9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Add the dialog component",
```
yinonov, line 3:
```suggestion
  "comment": "Migrate dialog component to fast",
```

yinonov (changes requested):

libs/components/src/lib/header/header.ts
```diff
@@ -10,6 +10,7 @@ export class Header extends FoundationElement {
 	/**
```
yinonov, line 10: better to revert any change to header as a big PR is on the way. don't need this to conflict

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -326,6 +326,9 @@ describe('vwc-text-field', () => {
 			element.dispatchEvent(new Event('blur'));
 		}
 
+		/**
+		 *
+		 */
```
yinonov, line 331: revert?
```suggestion
```

libs/components/src/types/style.d.ts
```diff
@@ -2,3 +2,8 @@ declare module '*.scss' {
 	const classes: import('@microsoft/fast-element').ComposableStyles;
 	export default classes;
 }
+
```
yinonov, line 5: where did u encounter such import problem? are u using import css assert?

libs/components/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "target": "ES2017",
-    "module": "ES2015",
+    "module": "ESNext",
```
yinonov, line 5: relates to dialog?

yinonov (changes requested):

libs/components/src/types/style.d.ts
YonatanKra: Yes - importing the polyfill `css`.

libs/components/tsconfig.json
YonatanKra: No. Just bumping it up and everything works :)

package.json
```diff
@@ -20,6 +20,7 @@
     "babel-polyfill": "^6.26.0",
     "blocking-elements": "^0.1.1",
     "core-js": "^3.22.8",
+    "dialog-polyfill": "^0.5.6",
```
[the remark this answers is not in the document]
YonatanKra: Support for Safari 14

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,260 @@
...
+            headline="Open Dialog">
+            <img slot="graphics" src="https://doodleipsum.com/40x40/hand-drawn?bg=7463D9&amp;i=af462b28146d2ac91599602e083ddee5">
+</vwc-dialog>
+```
+
+### footer
+Use the footer `slot` in order to add action buttons to the bottom of the dialog.
+```html preview
+<style>
+.demo-footer {
```
yinonov, line 153: adding the following will affect the iframe height
```css
html {
  block-size: 200px;
}
```

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,260 @@
...
+- Type: `boolean`
+- Default: `false`
+
+```html preview
+<vwc-dialog open
+            headline="Open Dialog">
+</vwc-dialog>
+```
+
+### returnValue
```
yinonov, line 67:
```suggestion
### Return Value
```

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,260 @@
...
+- Default: `undefined`
+
+```html preview
+<vwc-dialog open
+            content="This is the content">
+</vwc-dialog>
+```
+
+
+### open
```
yinonov, line 56:
```suggestion
### Open
```

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,260 @@
...
+    (function() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.addEventListener('close', (e) => console.log(e.detail));
+    })();
+</script>
+```
+
+
+## Methods
+### show
```
yinonov, line 93:
```suggestion

### Show
```

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,260 @@
...
+            onclick="openDialog()"></vwc-button>
+<script>
+    function openDialog() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.show();
+    }
+</script>
+```
+
+### showModal
```
yinonov, line 108:
```suggestion
### Show Modal
```

yinonov: our convention is to title nicely and include the code name in description. if we wanna deviate we need to align everywhere

yinonov (changes requested):

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,271 @@
+# Dialog
+
+```js
+<script type="module">import '@vonage/vivid/dialog';</script>
+```
+
+```html preview
+<style>
+  html { 
+    block-size: 200px;
```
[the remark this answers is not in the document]
YonatanKra: On my screen it looks good with 200

yinonov:
few questions before merge -
- should there be a way to prevent light dismiss?
- should the dialog align to top or be positioned in the center? remember past request to allow both

jshenkman:
> few questions before merge -
> 
> * should there be a way to prevent light dismiss?
> * should the dialog align to top or be positioned in the center? remember past request to allow both

I think we should keep both positions of the dialog (top and center) as we had in vivid2, as Yinon mentioned it was a request that came from users and was an essential addition for them. Can we add the `topPosition` property as we had in vivid-2?

https://vivid.vonage.com/?path=/story/components-dialog--top-position&args=topPosition:xlarge
Maybe we don't need all values (4 sizes)

yinonov:
> > few questions before merge -
> > 
> > * should there be a way to prevent light dismiss?
> > * should the dialog align to top or be positioned in the center? remember past request to allow both
> 
> I think we should keep both positions of the dialog (top and center) as we had in vivid2, as Yinon mentioned it was a request that came from users and was an essential addition for them. Can we add the `topPosition` property as we had in vivid-2?
> 
> https://vivid.vonage.com/?path=/story/components-dialog--top-position&args=topPosition:xlarge Maybe we don't need all values (4 sizes)

notice we are discussing a position terminology at [Consolidate position terminology](https://github.com/Vonage/vivid-3/issues/510)

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+    html { 
+        block-size: 200px; 
+    }
+    .demo-footer {
+        display: flex;
+        justify-content: flex-end;
+        margin-top: 16px;
+    }
+    .demo-footer  vwc-button {
+        margin-left: 8px;
```
yinonov, line 195: why not gap?

yinonov (changes requested):
strange color change on activate button

https://user-images.githubusercontent.com/15140652/178141413-967b8c36-7d9c-4fe5-92b2-cd6ebc344c67.mov

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+- Default: `false`
+
+```html preview
+<vwc-dialog open
+            headline="Open Dialog">
+</vwc-dialog>
+```
+
+### Top Position
+Use the `top-oposition` attribute or the `topPosition` property to override the dialog centering to the screen and allow setting a unique top.
```
yinonov, line 71: this terminology doesn't align with other components. review #510

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+### Top Position
+Use the `top-oposition` attribute or the `topPosition` property to override the dialog centering to the screen and allow setting a unique top. 
+
+- Type: `string`
+- Default: `small` | `medium` | `large` | `xlarge`
+  Sets or returns whether a dialog should be open or not
+
+- Type: `boolean`
+- Default: `false`
```
yinonov, line 78: in style, top position is default

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+  }
+</style>
+<vwc-dialog icon="info" headline="headline" content="Content" open>This is the content!</vwc-dialog>
+```
+
+## Members
+
+### Headline
+
+Use the `headline` attribute to set the dialog's title.
```
yinonov, line 20: in some scope _headline_ may equal _title_ but I'm not sure we should use it as equivalent term

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+```html preview
+<vwc-dialog open
+            headline="This is the title">app
```
yinonov, line 27: here too

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+    function closeDialog() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.returnValue = 'Value';
+        dialog.close();
+    }
+    (function() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.addEventListener('close', (e) => console.log(e.detail));
+    })();
+</script>
```
yinonov, line 110: couldn't tell it logs something to console. maybe print something to DOM? and the closing button might be redundant to reader here. simply close by the exist button of the dialog as above snippets

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+    function openDialog() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.show();
+    }
+</script>
+```
+
+### Show Modal
+
+Shows the dialog and makes it the top-most modal dialog.
```
yinonov, line 134: what other value does it feature? no light dismiss, different role? maybe add a couple of paragraphs and a reference to read more on the topic?

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+```html preview
+<vwc-dialog open
+            icon="info">
+</vwc-dialog>
+```
+
+### Content
```
yinonov, line 44: in Card the term is _text_

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Slots
+
+### Graphics
+
+Use the `graphics` slot in order to replace the icon.
+
+```html preview
+<vwc-dialog open
+            headline="Open Dialog">
```
yinonov, line 175: maybe remove headline to focus on the _graphic_ only?

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+    function closeDialog() {
+        const dialog = document.querySelector('vwc-dialog');
+        dialog.close();
+    }
+</script>
+```
+
+## Slots
+
+### Graphics
```
yinonov, line 169: align to _graphic_ (no 's'). as in card
https://github.com/Vonage/vivid-3/blob/d411e3363c233187f29a21cb5bc0b7e3e7323f8f/libs/components/src/lib/card/README.md#L70

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+vwc-text {
+    text-align: center;
+}
+</style>
+<vwc-dialog open>
+    <div slot="main" class="demo-main">
+       <vwc-text font-face="body-1-bold">Use main slot for your own layout and content</vwc-text>
+    </div>
+</vwc-dialog>
+```
```
yinonov, line 228: use our layout instead of arbitrary code
https://github.com/Vonage/vivid-3/blob/d411e3363c233187f29a21cb5bc0b7e3e7323f8f/libs/components/src/lib/card/README.md#L117-L126

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+## CSS Custom Properties
+
+### Inline-size
```
yinonov, line 228: not sure that's enough anymore in terms of fluidness. max/min size would probably be needed. given that also some requests regarding position were raised, the necessity of stacking context manoeuvres, the inability to apply styles on host (which contains both backdrop and the actual nested dialog), and the fact that dialog is partially categorized as layout, maybe _part_ access would be ideal?

yinonov (changes requested):

libs/components/src/lib/dialog/README.md
YonatanKra: `placement` then?

libs/components/src/lib/dialog/README.md
YonatanKra: I do not understand the comment. What do you mean? The default is `no top position`.
For `modal` it is the center of the screen, for `non-modal` it is top.

yinonov: why the difference?

YonatanKra: See HTML spec

libs/components/src/lib/dialog/README.md
YonatanKra: #510 resolution should include the dialog.

libs/components/src/lib/dialog/README.md
YonatanKra: Have no idea. It works like that... if you want me to change to something, I will. I am not attached :)

libs/components/src/lib/dialog/README.md
YonatanKra: It is intentional not to use layout to show how one can add a custom padding.

libs/components/src/lib/dialog/README.md
yinonov: so refactor to align with header action items and the likes...

libs/components/src/lib/dialog/README.md
YonatanKra: Maybe wait and see if feedback requires this before comitting to an API?
Or remove the `inline`?

yinonov: Remove inline for now to avoid introducing breaking changes. I expect we will revisit this very soon

libs/components/src/lib/dialog/README.md
rachelbt: so we should write:
```suggestion
Use the `headline` attribute to set the dialog's headline.
```

libs/components/src/lib/dialog/README.md
rachelbt: @yinonov if you are referring to the margin left of the button - than yes, I've forgot gap :(

Regarding the footer - the slot is always there - so if we will use gap - than we will always have it - even if not needed.
Wondering if we need to add margin-top to slotted items. Not sure its the right way.

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,297 @@
...
+    html { 
+        block-size: 200px; 
+    }
+    .demo-footer {
+        display: flex;
+        justify-content: flex-end;
+        margin-top: 16px;
+    }
+    .demo-footer  vwc-button {
+        margin-left: 8px;
```
rachelbt, line 180:
```suggestion
        column-gap: 8px;
        margin-top: 16px;

```

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,282 @@
...
+<vwc-dialog id="dialog" text="Text content"></vwc-dialog>
+
+<vwc-button label="Toggle Dialog Open" onclick="dialog.open = !dialog.open"></vwc-button>
+```
+
+### Return Value
+
+Sets or returns the dialog's return value when closing
+
+- Type: `string`
```
yinonov, line 69: I'm not sure the type is always a string. value of dialog might return a string representation of returnValue

libs/components/src/lib/dialog/README.md
```diff
@@ -0,0 +1,282 @@
...
+- Default: `undefined`
+
+```html preview
```
yinonov, line 72: note to follow mdn's example
https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/returnValue#examples

rachelbt:
> strange color change on activate button
> 
>  Screen.Recording.2022-07-10.at.13.39.08.mov

It is as all buttons are. 
Try adding `rgb(42, 42, 42)` that is the value for ` --vvd-color-surface-12dp `background color to buttons wrapper div

yinonov (approved):
there are still issues which can be addressed in another scope -

- elborate docs
- dialog box flexibility, including size & proportions

rachelbt (approved):

rachelbt (approved):

PR #507, opened 2022-06-30
feat(styles): typography css support

yinonov:


libs/styles/src/lib/typography/desktop.scss
```diff
@@ -0,0 +1,99 @@
...
+	strong, .body-1-bold {
+			@include typography.typography-cat-shorthand(body-1-bold);
+	}
+
+	code, .body-1-code {
+			@include typography.typography-cat-shorthand(body-1-code);
+	}
+
+	a:any-link, .body-1-link {
+			@include typography.typography-cat-shorthand(body-1-link);
```
rachelbt, line 53: I'm not sure link need a font-size...
Or if it has font-size - we need to think about cases like this:
image

rachelbt: I also not sure we need to do it on any a tag in general. I can effect things we don't want to be effected.

libs/styles/src/lib/typography/desktop.scss
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	.button-dense {
+			@include typography.typography-cat-shorthand(button-dense);
+	}
+
+	.button-enlarge {
+			@include typography.typography-cat-shorthand(button-enlarge);
+	}
+
```
rachelbt, line 83: currently we don't have different font size to dens + enraged. 
We don't refer them by this names any,more but `condensed` & `extended`

yinonov: maybe we don't need to provide these button categories at all in typography style? they are merely imported directly to button component

libs/styles/src/lib/typography/desktop.scss
yinonov: this is a valid point but selecting only classes will require authors to apply these classes to tags all over their apps.
maybe this should be split to font-face & typography

rachelbt: not sure I understand you. "font-face & typography" - font-face (font family) is part of typography

libs/styles/src/lib/typography/desktop.scss
```diff
@@ -0,0 +1,98 @@
...
+			@include typography.typography-cat-shorthand(body-1-bold);
+	}
+
+	code {
+			@include typography.typography-cat-shorthand(body-1-code);
+	}
+
+	a:any-link {
+			text-decoration: underline;
+	}
```
yinonov, line 53:
```suggestion
	// a:any-link {
	//		text-decoration: underline;
	// }
```

apps/docs/_layouts/base.njk
```diff
@@ -10,6 +9,8 @@
 
     <!-- Load fonts -->
     <link rel="preload" href="https://fonts.resources.vonage.com/fonts/v1/Spezia_Web_Complete_Upright.woff2" as="font" type="font/woff2" crossorigin>
+    <link rel="preload" href="https://fonts.resources.vonage.com/fonts/v1/Spezia_Web_Monospace_Complete.woff2" as="font" type="font/woff2" crossorigin>
```
YonatanKra, line 12: So we now need to load 2 fonts?

apps/docs/index.njk
```diff
@@ -4,4 +4,4 @@ title: yo title
 permalink: /
 ---
 
-{% renderFile './README.md' %}
+{% renderFile './libs/components/README.md' %}
```
YonatanKra, line 7: Wasn't this from a different PR?

change/@vonage-vivid-945371a4-98ce-41b7-84d7-5672b0fc1635.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
```
YonatanKra, line 2: This doesn't seem to affect the package itself.

libs/components/README.md
```diff
@@ -115,7 +115,7 @@ Global content delivery networks can help quickly integrate content within html
 Such practice is often used when working on POCs or reproduction environments.
 Tools like [UNPKG](https://unpkg.com), [jsDeliver](https://www.jsdelivr.com), [Skypack](https://www.skypack.dev) etc' are bound to deliver any content registered in the npm registry.
 
-The following example will output a fully rendered Vivid button component
+The following snippet fully renderes a Vivid button component
```
YonatanKra, line 118:
```suggestion
The following snippet fully renders a Vivid button component
```

libs/styles/src/lib/typography/desktop.scss
```diff
@@ -0,0 +1,98 @@
...
+	// 		@include typography.typography-cat-shorthand(caption-bold);
+	// }
+
+	// .caption-code {
+	// 		@include typography.typography-cat-shorthand(caption-code);
+	// }
+
+	// .caption-link {
+	// 		@include typography.typography-cat-shorthand(caption-link);
+	// }
```
YonatanKra, line 97: Why is that still here? Delete or uncomment if needed.

apps/docs/_layouts/base.njk
yinonov: yes, these are preloads to enforce loading even if aren't used. due to our high usage of monospace within our docs I've added that preload

apps/docs/index.njk
yinonov: was removed before cause it failed the a11y test. now it passes with text component removed

change/@vonage-vivid-945371a4-98ce-41b7-84d7-5672b0fc1635.json
yinonov:
```suggestion
  "type": "none",
```

libs/styles/src/lib/typography/desktop.scss
yinonov: there's much to do in this css and we need to be reminded of what the text component supported. there's more exploration to do but will have to be made progressively. this must be revisited

YonatanKra (approved):

YonatanKra:
@yinonov - maybe we should add to the docs how this works. 
1. What's with the double import of fonts (mono etc.)
2. That it auto applies styles to certain elements
Doesn't have to be a part of this PR but a task for laters.

PR #511, opened 2022-07-04
Header side drawer modifications sticky

rachelbt:
a quick & dirty POC for sticky header


rachelbt:
PR for 
POC for sticky.

PR #514, opened 2022-07-05
fix(sidenav-item and accordion-item): add appearance

rinaok:


libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -29,6 +44,8 @@
 
 .button {
 	@include typography.typography-cat-shorthand('subtitle-2');
+	@include appearance.appearance;
+	@include connotation.connotation;
```
rachelbt, line 48: not sure these are needed...

rinaok: Screen Shot 2022-07-07 at 12 27 58

@rachelbt

PR #515, opened 2022-07-05
fix(breadcrumb): change template to nav

YonatanKra:


libs/components/src/lib/breadcrumb-item/breadcrumb-item.spec.ts
```diff
@@ -189,4 +191,20 @@ describe('vwc-breadcrumb-item', () => {
 			expect(anchorElement?.getAttribute(attribute)).toEqual(text);
 		});
 	});
+
+	describe('a11y', () => {
+		it('should pass accessibility test', async () => {
+			const { shadowRoot } = element;
+			if (!shadowRoot) { return; }
+
+			const results = await axe(shadowRoot.innerHTML, {
```
yinonov, line 200: Can we check breadcrumb with its items?
Need to check how's it testing the items templates as well

libs/components/src/lib/breadcrumb/breadcrumb.template.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * The template for the {@link @microsoft/fast-foundation#Breadcrumb} component.
+ *
+ * @param context
+ * @param definition
+ * @public
+ */
+export const breadcrumbTemplate: FoundationElementTemplate<ViewTemplate<Breadcrumb>> = () => html`
+    <nav role="navigation" class="base">
+        <div role="list" class="list" part="list">
```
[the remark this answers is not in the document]
YonatanKra: Yes - this is how `fast` implemented it.
I think UL might be more fitting, but don't we go with `fast` on such decisions?

libs/components/src/lib/breadcrumb-item/breadcrumb-item.spec.ts
YonatanKra: It is tested with its children:
```
<vwc-breadcrumb>
 <vwc-breadcrumb-item href="#" text="breadcrumb"></vwc-breadcrumb-item>
    <vwc-breadcrumb-item text="..."></vwc-breadcrumb-item>
    <vwc-breadcrumb-item href="#" text="breadcrumb"></vwc-breadcrumb-item>
    <vwc-breadcrumb-item text="breadcrumb"></vwc-breadcrumb-item>
<vwc-breadcrumb>
```

libs/components/src/lib/breadcrumb-item/breadcrumb-item.spec.ts
```diff
@@ -189,4 +190,23 @@ describe('vwc-breadcrumb-item', () => {
 			expect(anchorElement?.getAttribute(attribute)).toEqual(text);
 		});
 	});
+
+	describe('a11y', () => {
+		it('should pass accessibility test', async () => {
```
yinonov, line 195: this should actually fail cause list items with aren't nested in list are wrong

yinonov (changes requested):

libs/components/src/lib/breadcrumb-item/breadcrumb-item.spec.ts
YonatanKra: It doesn't fail though... an issue with `axe`?

yinonov (approved):

PR #527, opened 2022-07-08
fix(components): side drawer blocking elements

yinonov:
- [ ] add missing tests to ensure blocking elements work

### DIFF
--- a/apps/docs/assets/styles/main.css
+++ b/apps/docs/assets/styles/main.css
@@ -47,7 +47,7 @@ vwc-side-drawer:not([open]) {
 }
 
 vwc-side-drawer::part(base) {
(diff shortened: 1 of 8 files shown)


yinonov:
blocking elements and inert polyfills will no longer be needed this upcoming fall

apps/docs/assets/styles/main.css
```diff
@@ -47,7 +47,7 @@ vwc-side-drawer:not([open]) {
 }
 
 vwc-side-drawer::part(base) {
-  z-index: 2;
+  /* z-index: 2; */
```
yinonov, line 50:
```suggestion
```

rinaok (approved):

PR #533, opened 2022-07-10
fix(tooltip): accessibility

rinaok:


rinaok:
@yinonov We should add an accessibility section to popup's readme as well, but I'm not sure about the ARIA role it should have.

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -21,7 +21,7 @@ export const TooltipTemplate: (
 ) => ViewTemplate<Tooltip> = () => html`
 <vwc-popup class="${getClasses}" arrow alternate="true"
```
yinonov, line 22: `alternate="true"` can just be `alternate`

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -21,7 +21,7 @@ export const TooltipTemplate: (
 ) => ViewTemplate<Tooltip> = () => html`
 <vwc-popup class="${getClasses}" arrow alternate="true" 
 corner=${(x) => x.corner} open=${(x) => x.open} anchor=${(x) => x.anchor}
-  exportparts="vvd-theme-alternate">
+  exportparts="vvd-theme-alternate" role="tooltip">
```
yinonov, line 24: sorry to bring this up, unrelated to changes - is `exportparts` required?

yinonov (changes requested):

libs/components/src/lib/tooltip/tooltip.template.ts
rinaok: @yinonov yes, without it tooltip can't be alternate

libs/components/src/lib/popup/popup.ts
```diff
@@ -97,6 +98,8 @@ export class Popup extends FoundationElement {
 		switch (name) {
 			case 'anchor': {
 				this.#anchorEl = this.#getAnchorById();
+				// close the popup if pressed escape
+				this.#anchorEl?.addEventListener('keydown', (e) => this.#handleKeydown(e as KeyboardEvent));
```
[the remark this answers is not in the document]
rinaok: @yinonov we have the cleanup method. Do you think we should remove it anyways?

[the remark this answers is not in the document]
rinaok: removed it at disconnectedCallback

rachelbt (approved):
LGTM

libs/components/src/lib/popup/README.md
```diff
@@ -185,20 +187,20 @@ Use the `corner` attribute to set the placement of the popup around the anchor.
...
 
-This helps ensure elements don't render over top a popup undesirebly.
+Updates popup's position.
 
-## Methods
+## Caveat
 
-### updatePosition()
+Document elements display precedence is formed by the imaginary z-axis [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), commonly by order of which elements are rendered and special properties (e.g. _z-index_).
+Popup component is a low level element, unaware of its document context, but in most cases required to overlay on top of all elements.
```
yinonov, line 202: adjusting the syntax. this is copied from another page right? can u align with this?
```suggestion
Popup component is a low level element, unaware of its document context, but is, in most cases, required to overlay on top of all elements.
```

libs/components/src/lib/popup/README.md
```diff
@@ -185,20 +187,20 @@ Use the `corner` attribute to set the placement of the popup around the anchor.
...
 
-A common practice used in apps / frameworks to promote a popup component to top other elements z-axis, is to utilise a service that dynamically appends a popup component to the end of the body element, when called for.
+- Type: `function`
+- Returns: `void`
 
-This helps ensure elements don't render over top a popup undesirebly.
+Updates popup's position.
 
-## Methods
+## Caveat
```
yinonov, line 199: maybe we should have it in tooltip too?

rinaok: I'm not sure, maybe @rachelbt @YonatanKra have an opinion about it?

yinonov: with pleasure
https://github.com/material-components/material-web/issues/832

in fast doc -
https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/dialog/dialog.spec.md#caveat

there are dozen more if need

rinaok: @yinonov OK, added the Caveat to tooltip as well

yinonov (approved):

PR #534, opened 2022-07-11
chore(ci): optimize ci

YonatanKra:


.github/workflows/_lint.yml
```diff
@@ -11,5 +11,14 @@ jobs:
         with:
           node-version: '16'
           cache: 'npm'
-      - run: npm ci --only=dev
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: node_modules/
+          key: ${{ runner.os }}-${{ hashFiles('package-lock.json') }}
```
yinonov, line 18: cant this be env variable we pass on from main action to reusable actions?

yinonov (changes requested):

.github/workflows/_lint.yml
YonatanKra: What should be env var? The key?

YonatanKra: It can - do you have a use case for this now? Because it is going to complicate the PR significantly.

yinonov (approved):

.github/workflows/_lint.yml
YonatanKra: See https://github.com/Vonage/vivid-3/issues/537

PR #535, opened 2022-07-12
refactor(tokens): migrate design tokens renderers to next

yinonov:


libs/components/project.json
```diff
@@ -119,5 +119,5 @@
     }
   },
   "tags": [],
-  "implicitDependencies": ["styles"]
+  "implicitDependencies": ["tokens"]
```
YonatanKra, line 122: About time!!!
image

libs/tokens/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+	"presets": [["@nrwl/web/babel", { "useBuiltIns": "usage" }]]
```
YonatanKra, line 2: What does this do?

libs/tokens/README.md
```diff
@@ -0,0 +1,11 @@
+# tokens
```
YonatanKra, line 1: Is this a good readme?

libs/tokens/src/lib/build.ts
```diff
@@ -0,0 +1,60 @@
+const sourceOnly = require('./filters/source-only');
```
YonatanKra, line 1: Why are we using `require` here?

libs/tokens/src/lib/formatters/constants.ts
```diff
@@ -0,0 +1,10 @@
+const { formatHelpers: { fileHeader } } = require('style-dictionary');
```
YonatanKra, line 1: How does this work? It doesn't look like constants... maybe constants generator?

libs/tokens/src/lib/filters/source-only.ts
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
YonatanKra, line 1: When there's no test, I need to guess as to how this filter should work.

libs/tokens/src/lib/tokens.spec.ts
```diff
@@ -0,0 +1,7 @@
+import { tokens } from './tokens';
```
YonatanKra, line 1: This file disturbs my sleep... :)
Let's talk about it when in office

libs/tokens/src/lib/tokens.ts
```diff
@@ -0,0 +1,3 @@
+export function tokens(): string {
```
YonatanKra, line 1: ?

libs/tokens/src/lib/transforms/shadow-shorthand.ts
```diff
@@ -0,0 +1,21 @@
+const { transform } = require('style-dictionary');
```
YonatanKra, line 1: Why require?
Also, what `shadow-shorthand` stands for?
When there aren't tests, ppl get lost...

YonatanKra (changes requested):
Missing tests. Let's talk about it offline and get it together.

libs/tokens/.babelrc
yinonov: Have no idea. Nx did that

libs/tokens/README.md
yinonov: it's not wrong :D

libs/tokens/src/lib/formatters/constants.ts
yinonov: the correct term according to SD is formatter. thus, within formatters folder

libs/tokens/src/lib/tokens.spec.ts
yinonov: this is generated by nx. not sure I want it. do we have any convention calling the entry point file as the lib itself?

libs/tokens/src/lib/tokens.ts
yinonov: same as https://github.com/Vonage/vivid-3/pull/535#discussion_r934022250

libs/tokens/src/lib/transforms/shadow-shorthand.ts
yinonov: SD is documented at its core as commonjs.
yes this can sure get itself tested

libs/tokens/src/lib/build.ts
yinonov: SD is a commonjs lib

yinonov, line 1: @YonatanKra having and this file is the main entry for tokens package, therefore is self explanatory, and is responsible for generating scss constants & css variables that components rely on, I found creating a test for it fails a DRY principal where the test file was simply copying this same context.

failure to run will fail build / visual regression tests.

still think of any reasonable sanity tests here?

libs/tokens/src/lib/formatters/constants.ts
yinonov: should filename itself explicitly state it being a formatter?

libs/tokens/src/lib/formatters/constants.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+					"attributes": {},
+					"path": ["color", "canvas"]
+				}
+			]
+		},
+		"file": { "destination": "_constants.scss", "options": {} }
+	};
+
+	it('should generate scss constants from token', () => {
+		expect(formatter(token)).toContain(`$vvd-color-canvas: --vvd-color-canvas;`);
```
YonatanKra, line 28: Why is it called `formatter` then? Wouldn't a name like `tokenToScssConstant` be clearer?

libs/tokens/src/lib/build.ts
```diff
@@ -0,0 +1,60 @@
...
+			files: [{
+				destination: `_${theme}.mixin.scss`,
+				format: "css/variables",
+				filter: "sourceOnly",
+				options: {
+					selector: "@mixin variables"
+				}
+			}]
+		},
+		dev: {
```
YonatanKra, line 36: What's the `dev` platform for? How do I use it?

libs/tokens/src/lib/build.ts
YonatanKra: It's not test as much as documentation.
How do I use this file? Can I set env variables that will make it behave differently?

libs/tokens/src/lib/formatters/constants.spec.ts
yinonov: It's a member of an object with specific interface. Same as in connectedCallback

libs/tokens/src/lib/build.ts
yinonov: Web platform build the tokens.
Dev builds constants we use only in development and aren't being used at run-time

YonatanKra: I could tell that from the name but - how do I use it? Why would dev mode need to build constants that aren't used at run-time? And what do you mean by `run-time` - is it the components' runtime or the tokens' runtime (which is the components' build time)?

yinonov: that's what dev outputs for scss variables, it's only for our safety. it may also build types in the future. these are only used in development. they run **once** to provide a mapping for variables.

```css
$vvd-color-on-cta: --vvd-color-on-cta;
$vvd-color-cta-10: --vvd-color-cta-10;
$vvd-color-cta-20: --vvd-color-cta-20;
$vvd-color-cta-30: --vvd-color-cta-30;
$vvd-color-cta-70: --vvd-color-cta-70;
$vvd-color-cta-90: --vvd-color-cta-90;
```

actual tokens (web) run a different formatting and iterate over themes (**more than once**).
they output the following (more extensibly)
```css
@mixin variables {
	--vvd-sizing-4px-scale-3xs: 8px;
	--vvd-sizing-4px-scale-2xs: 12px;
	--vvd-sizing-4px-scale-xs: 16px;
	--vvd-sizing-4px-scale-sm: 20px;
	--vvd-sizing-4px-scale-md: 24px;
	--vvd-sizing-4px-scale-lg: 28px;
	--vvd-sizing-4px-scale-xl: 32px;
	--vvd-sizing-4px-scale-2xl: 36px;
	--vvd-sizing-4px-scale-3xl: 40px;
	--vvd-sizing-8px-scale-3xs: 8px;
	--vvd-sizing-8px-scale-2xs: 16px;
	--vvd-sizing-8px-scale-xs: 24px;
	--vvd-sizing-8px-scale-sm: 32px;
	--vvd-sizing-8px-scale-md: 40px;
	--vvd-sizing-8px-scale-lg: 48px;
	--vvd-sizing-8px-scale-xl: 56px;
	--vvd-sizing-8px-scale-2xl: 64px;
	--vvd-sizing-8px-scale-3xl: 72px;
	--vvd-color-surface-0dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-2dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-4dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-8dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-12dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-16dp: rgba(255, 255, 255, 1);
	--vvd-color-surface-24dp: rgba(255, 255, 255, 1);
}
```

libs/tokens/src/lib/build.ts
yinonov: not sure what u mean. how are env variables related? to affect the file system? I tried creating a test but it resulted in almost the same code. is it worth creating a test just for the sake of its definition? just for having a clear statement of what it is? then let's just have a comment stating the same thing.

if this code breaks or doesn't provide exactly what it needs to, our whole build of components will fail. there's no way this goes under the radar

YonatanKra (approved):
I'll take the test as a side project

PR #538, opened 2022-07-12
chore(docs): drawer modal conditional to desktop or mobile

yinonov:
- [x] initial load should set side drawer `open` conditionally to viewport size


apps/docs/assets/scripts/viewport-handler.js
```diff
@@ -0,0 +1,24 @@
...
+
+	requestAnimationFrame(() => {
+		pendingUpdate = false;
+
+		sideDrawer.modal = width < mobileWidth;
+	});
+
+}
+
+(function () {
```
rachelbt, line 18: this doesn't work - the drawer stays open :(

rachelbt (changes requested):

apps/docs/assets/scripts/viewport-handler.js
yinonov: This only occurs on init

rachelbt (approved):
LGTM

PR #553, opened 2022-07-18
fix(header): docs and slots

rinaok:


libs/components/src/lib/header/README.md
```diff
@@ -51,19 +51,19 @@ The default slot sets assigned nodes to the start of the header.
...
+  <vwc-button slot="action-items" icon="twitter-mono"></vwc-button>
+  <vwc-button slot="action-items" icon="facebook-mono"></vwc-button>
+  <vwc-button slot="action-items" icon="heart-solid"></vwc-button>
 </vwc-header>
 ```
 
 ### App Content
 
-An optional approach to using header in application is to assign application context directly to the header's `app-content` slot, which will set content to follow, vertically, the header itself.
+It is also possible to assign application context directly to the header's `app-content` slot, which will allow content to follow, vertically, the header.
```
rachelbt, line 66:
```suggestion
It is also possible to assign application context directly to the header's `app-content` slot, which will allow content to follow, the header vertically.
```
It still sounds somewhat complicated....
Is it a must or only if the header is fixed?

libs/components/src/lib/header/README.md
```diff
@@ -82,10 +82,7 @@ An optional approach to using header in application is to assign application con
 
 ### Block Size
 
-Header block size is set definitively.
-Often, it is needed to use the block size value of a header in accordance with other elements in the application.
-`--_header-block-size` is a private custom property holding the block size value and is applied to header style internally.
-This property isn't customizable by authors but does pierce in and can be set to apply style on assigned content.
+The size of the header block is set definitively. A header's block size value is often used in conjunction with other elements in the application. The `--_header-block-size` private custom property is applied internally to header styles and holds the block size value. This property isn't customizable by authors but does pierce in and can be set to apply style to assigned content.
```
YonatanKra, line 85: Why change from bullets?

YonatanKra (approved):
I approve. See my comments. Nothing serious, so if you think it's good, go ahead and merge.

PR #556, opened 2022-07-20
fix(dialog): min & max sizes

rachelbt:


libs/components/src/lib/dialog/README.md
```diff
@@ -6,8 +6,8 @@
 
 ```html preview
 <style>
-  html { 
-    block-size: 200px; 
```
YonatanKra, line 9: I think we need to solve this. We said users need only copy paste, but this thing is not part of the copy-pasting stuff.
Not for this PR - just adding this to track in its own issue.

YonatanKra (approved):
LGTM

libs/components/src/lib/dialog/README.md
rachelbt: https://github.com/Vonage/vivid-3/issues/559

PR #573, opened 2022-07-27
fix(layout): inline & block gutters

rachelbt:


rinaok (changes requested):
Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?

libs/components/src/lib/layout/README.md
```diff
@@ -37,7 +37,8 @@ Using layout, elements can be arranged easily in a variety of sizes and spaces t
 
 Use the `gutters` attribute to set the initial preferred `margin`.
 
-- Type: `'small'` | `'medium'` | `'large'`
+- Type: `'small'` | `'small-inline'` | `'small-block'` | `'medium'` | `'medium-inline'` | `'medium-block'` | `'large'` | `'large-inline'` | `'large-block'` 
+
 - Default: `undefined`
```
yinonov, line 42: Can this be define as none in docs?

yinonov (changes requested):

libs/components/src/lib/layout/README.md
rachelbt: done

rachelbt:
> Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?

@yinonov @rinaok 
I added dividers, I think now its better, no?

rinaok:
> > Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?
> 
> @yinonov @rinaok I added dividers, I think now its better, no?

@rachelbt I still think that it would be better with two different examples

rinaok:
> > > Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?
> > 
> > 
> > @yinonov @rinaok I added dividers, I think now its better, no?
> 
> @rachelbt Looks better, but I still think that it would be better with two different examples

@rachelbt Added a change, what do you think? I think that it's more clear now

rachelbt:
> > > > Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?
> > > 
> > > 
> > > @yinonov @rinaok I added dividers, I think now its better, no?
> > 
> > 
> > @rachelbt Looks better, but I still think that it would be better with two different examples
> 
> @rachelbt Added a change, what do you think? I think that it's more clear now

its more clear but it makes the preview code very big. 
@yinonov what do you say?

rinaok:
> > > > > Didn't understand the example. Maybe it should be two examples? One with `inline` sizes and one with `block` sizes?
> > > > 
> > > > 
> > > > @yinonov @rinaok I added dividers, I think now its better, no?
> > > 
> > > 
> > > @rachelbt Looks better, but I still think that it would be better with two different examples
> > 
> > 
> > @rachelbt Added a change, what do you think? I think that it's more clear now
> 
> its more clear but it makes the preview code very big. @yinonov what do you say?

@rachelbt Maybe you can move the gutters examples below the Column Basis example and then it will look better?

yinonov (approved):

rinaok (dismissed):

rinaok (approved):

PR #577, opened 2022-07-28
fix(sidenav-disclosure): chevrons

rinaok:


libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.ts
```diff
@@ -24,6 +26,31 @@ export class SidenavDisclosure extends FoundationElement {
 	 */
 	@attr({ mode: 'boolean' }) open = false;
 
+	/**
+	* @internal
+	*/
+	override connectedCallback(): void {
+		super.connectedCallback();
+		this.onToggle = this.onToggle.bind(this);
```
YonatanKra, line 34: This can be written better like this:
```
class SomeClass() {
  boundFunction = () => {

  }
}
```
It is cleaner than using `.bind` IMO.

YonatanKra: Have you tried this?

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.ts
```diff
@@ -24,6 +26,30 @@ export class SidenavDisclosure extends FoundationElement {
 	 */
 	@attr({ mode: 'boolean' }) open = false;
 
+	/**
+	* @internal
+	*/
+	override connectedCallback(): void {
+		super.connectedCallback();
+		this.details.addEventListener('toggle', () => this.onToggle(this));
```
YonatanKra, line 34: Why do you need to send `this` to `onToggle`?

libs/components/src/lib/sidenav-disclosure/sidenav-disclosure.ts
```diff
@@ -24,6 +26,30 @@ export class SidenavDisclosure extends FoundationElement {
...
+		this.details.addEventListener('toggle', () => this.onToggle());
+		this.details.open = this.open;
+	}
+
+	/**
+	 * @internal
+	 */
+	override disconnectedCallback(): void {
+		super.disconnectedCallback();
+		this.details.removeEventListener('toggle',  () => this.onToggle());
```
YonatanKra, line 43: Have you tested this?

YonatanKra (approved):

PR #583, opened 2022-08-01
fix(docs): onload readme files

rinaok:


rinaok:
@yinonov @rachelbt What do you think about the loading of the page now? I think that it looks better.

apps/docs/_layouts/base.njk
```diff
@@ -61,6 +61,12 @@
     <script type="module" src="{{ '/assets/modules/components/action-group/index.js' | url}}"></script>
 
     {% include "prefetches.njk" %}
+
+    <script>
+      window.onload = (event) => {
```
yinonov, line 66: `window.onload` doesn't necessarily mean custom elements are ready to display

see -

1. https://developer.mozilla.org/en-US/docs/Web/CSS/:defined
2. https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined

yinonov (changes requested):
you might also find this interesting
https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust

rachelbt:
> @yinonov @rachelbt What do you think about the loading of the page now? I think that it looks better.

it still a bit jumpy but much better!!

apps/docs/_layouts/base.njk
```diff
@@ -63,40 +63,28 @@
     {% include "prefetches.njk" %}
```
yinonov, line 63: has anything changed in this file or just formatting?

apps/docs/assets/styles/main.css
```diff
@@ -28,13 +28,22 @@ a:visited {
...
-#menu-button {
+#menu-button:not(:defined) {
+  display: none;
+}
+
+#menu-button:defined {
+  display: block;
   position: fixed;
-  inset: auto auto 8px 4px;
+  inset: auto auto 8px 8px;
```
yinonov, line 38: is this reverting @rachelbt's recent change?

apps/docs/assets/styles/main.css
```diff
@@ -67,6 +76,14 @@ vwc-side-drawer::part(base) {
   }
 }
 
+vwc-header:not(:defined) {
+  display: none;
+}
+
+vwc-header:defined {
+  display: block;
+}
```
yinonov, line 85: guessing this is not required(?)

apps/docs/assets/styles/main.css
```diff
@@ -67,6 +76,14 @@ vwc-side-drawer::part(base) {
   }
 }
 
+vwc-header:not(:defined) {
```
yinonov, line 79: only stating that targeting this specific component can still result in CLS. if it loads first but others don't load yet...

anyway, it works smoothly now

yinonov (changes requested):

apps/docs/assets/styles/main.css
rachelbt: no, it doesn't. 
I'll move the fab to the header, so maybe this will not be needed though

apps/docs/_layouts/base.njk
rinaok: just formatting

apps/docs/assets/styles/main.css
rinaok: removed

apps/docs/_layouts/base.njk
yinonov: than better to revert

yinonov (changes requested):

apps/docs/_layouts/base.njk
rinaok: reverted

yinonov (approved):

PR #584, opened 2022-08-02
feat(number-field): add number-field component

YonatanKra:
Note there's an issue with FAST's form association - you cannot use `required` with form submit


libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -0,0 +1,158 @@
...
+             aria-live="${x => x.ariaLive}"
+             aria-owns="${x => x.ariaOwns}"
+             aria-relevant="${x => x.ariaRelevant}"
+             aria-roledescription="${x => x.ariaRoledescription}"
+             ${ref('control')}
+      />
+      ${() => focusTemplate}
+      ${() => numberControlButtons()}
+    </div>
+	  ${when(x => !x.errorValidationMessage && x.helperText?.length, renderHelperText())}
```
yinonov, line 154: is this our implementation? if so, should be reference by the input element `describedBy`?

yinonov (changes requested):

libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -0,0 +1,158 @@
...
+
+function setControlButtonDensity(numberField: NumberField) {
+	return numberField.density === Density.Extended ? Density.Normal : Density.Condensed;
+}
+
+
+function numberControlButtons() {
+	return html<NumberField>`
+			<div class="control-buttons">
+				<vwc-button id="subtract" icon="minus-line"
```
yinonov, line 80: use context to retrieve actual tag as in
https://github.com/Vonage/vivid-3/blob/1080172e759f42f3a73e2138e23a52043efcc439/libs/components/src/lib/checkbox/checkbox.template.ts#L37-L38

yinonov (changes requested):

libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -0,0 +1,158 @@
...
+function setControlButtonShape(numberField: NumberField) {
+	return numberField.shape === Shape.Pill ? Shape.Pill : null;
+}
+
+function setControlButtonDensity(numberField: NumberField) {
+	return numberField.density === Density.Extended ? Density.Normal : Density.Condensed;
+}
+
+
+function numberControlButtons() {
```
yinonov, line 77: not sure whether this is a valid pattern but worth examination - buttons are controlling the number input
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

yinonov (changes requested):

libs/components/src/lib/number-field/README.md
```diff
@@ -0,0 +1,119 @@
...
+### Disabled
+
+Add the `disabled` attribute to disable the number field.
+
+- Type: `boolean`
+- Default: `false`
+
+```html preview blocks
+<vwc-number-field disabled value="disabled" label='fieldset' appearance='fieldset'></vwc-number-field>
+```
```
yinonov, line 108: when disabled, should buttons be disabled as well? or at least set to `pointer-events: none`? should examine from an a11y perspective.

image

yinonov (changes requested):

libs/components/src/lib/number-field/README.md
```diff
@@ -0,0 +1,119 @@
...
+### Readonly
+
+Add the `readonly` attribute to restrict user from changing the number field's value.
+
+- Type: `boolean`
+- Default: `false`
+
+```html preview blocks
+<vwc-number-field readonly value="readonly text" label='fieldset' appearance='fieldset'></vwc-number-field>
+```
```
yinonov, line 119: was still able to modify value
image

yinonov (changes requested):

libs/components/src/lib/number-field/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { NumberField } from './number-field';
+import { NumberFieldTemplate as template } from './number-field.template';
+
+export const vividNumberField =
+	NumberField.compose<FoundationElementDefinition>({
+		baseName: 'number-field',
+		template: template as any,
+		styles,
+	});
```
yinonov, line 15: it's either we set focus delegation here or remove from text field & text area

yinonov (changes requested):

AyalaBu:
@YonatanKra 
Looks great!

libs/components/src/lib/number-field/number-field.template.ts
YonatanKra: Not relevant - as buttons are not input elements IMO.

libs/components/src/lib/number-field/README.md
```diff
@@ -0,0 +1,119 @@
+# number-field
+
+Represents a number-field custom element. Follows the [`HTMLInput` of `type=text` specifications](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text).
```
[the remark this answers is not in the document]
YonatanKra: Well... we follow two things:
1. Fast
2. https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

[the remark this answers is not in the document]
YonatanKra: Added to readme

libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -0,0 +1,164 @@
...
+
+function numberControlButtons(context: ElementDefinitionContext) {
+	const buttonTag = context.tagFor(Button);
+	const dividerTag = context.tagFor(Divider);
+	return html<NumberField>`
+			<div class="control-buttons">
+				<${buttonTag} id="subtract" icon="minus-line" 
+				            shape="${ setControlButtonShape }"
+				            density="${ setControlButtonDensity }" 
+				            @click="${x => adjustValueByStep(x, SUBTRACT)}"></${buttonTag}>
```
yinonov, line 89: native number controls feature a behavior that isn't supported here. holding mousedown should increment / decrement numbers repeatedly

libs/components/src/lib/number-field/number-field.template.ts
yinonov: they control the value of the input. is there no recommended aria pattern to reflect that relationship?

yinonov (changes requested):
still -
image

yinonov (approved):

yinonov:
I noticed on mobile that when clicking on the controls it blurs out of control focus

yinonov:
Question to ask ourselves. Is helper text equivalent to tooltip?

PR #592, opened 2022-08-02
fix(popup): add remove keydown listener test

rinaok:


libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -243,6 +243,18 @@ describe('vwc-popup', () => {
...
+			const spy = jest.fn();
+			element.addEventListener('keydown', spy);
+	
+			element.disconnectedCallback();
+			element.connectedCallback();
+	
+			element.dispatchEvent(new Event('keydown'));
+	
+			expect((spy as any).mock.calls.length).toEqual(1);
+		});
```
YonatanKra, line 257: I don't think this is a good test here.
What is your code supposed to do?
It `should remove keydown listener after disconnection`. But in this case, it does not remove the listener on the `element` itself - the listener is on the `anchor`.
So what you should test is that the listener is removed from the anchor and not from the element.

```suggestion
		it('should remove keydown listener after disconnection', async function() {
			const spy = jest.fn();
			anchor.addEventListener('keydown', spy);
	
			element.disconnectedCallback();
	
			anchor.dispatchEvent(new Event('keydown'));
	
			expect((spy as any).mock.calls.length).toEqual(0);
		});
```
In addition, you also need to test the case in which you change the `anchor`.

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -228,20 +228,44 @@ describe('vwc-popup', () => {
...
 				.toEqual(true);
 
+			await elementUpdated(element);
+			element.disconnectedCallback();
 			await elementUpdated(element);
 			anchor.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' }));
+
+			expect(element.open)
+				.toEqual(true);
+		});
```
YonatanKra, line 255:
```suggestion
		it('should remove keydown listener after disconnection', async function() {
			const anchor = await setupPopupToOpenWithAnchor();
			const openStateBeforeEsc = element.open;

			await elementUpdated(element);
			element.disconnectedCallback();
			await elementUpdated(element);
			anchor.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' }));

      expect(openStateBeforeEsc)
				.toEqual(true);
			expect(element.open)
				.toEqual(true);
		});
```

Convention is to set the expectations at the end of the test. Make sure it is the same for all other tests plz.

PR #594, opened 2022-08-02
feat(components): new component avatar

rachelbt:
STILL IN WORK

- [x] density class
- [x] icon size in density
- [x] initials - font size of and density
- [ ] initials - [separate component for this](https://github.com/Vonage/vivid-3/issues/613)
- [ ] block-size [separate-issue](https://github.com/Vonage/vivid-3/issues/645)
- [x] image and outline
- [x] do we need ghost?! - only in image
- [x] block size - min-width 128px max - 100%
- [x] ui-tests
- [x] unit-test


libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+```html preview
+<vwc-avatar density='condensed' shape="pill"></vwc-avatar>
+<vwc-avatar density='normal' shape="pill"></vwc-avatar>
+<vwc-avatar density='extended' shape="pill"></vwc-avatar>
+```
+
+
+## Slots
+### Default
+Use the `default` slot to have an image avatar.
```
yinonov, line 83:
```suggestion
Use the `default` slot to set a graphic media of any kind (e.g. image, illustration etc'.)
```

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+</script>
+```
+
+```html preview
+<vwc-avatar></vwc-avatar>
+```
+
+## Members
+
+### Icon
```
yinonov, line 17: what about?
```js
- Type: `string`
- Default: `''`
```

yinonov: maybe something like
https://github.com/Vonage/vivid-3/blob/a8a8db32b29ae377e8b34188e06ef0265266ffb4/libs/components/src/lib/nav-item/README.md#L27-L35

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+Use the `icon` to set any icon for the avatar.  
+Default: 'user-line'
+```html preview
+<vwc-avatar icon="user-solid"></vwc-avatar>
+```
+
+
+### Acronym
+still in progress
```
yinonov, line 26:
```suggestion
<!-- 
### Acronym
still in progress
-->
```

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+- Default: `'filled'`
+
+```html preview
+<vwc-avatar appearance="filled"></vwc-avatar>
+<vwc-avatar appearance="outlined"></vwc-avatar>
+```
+
+
+### Connotation
+The `connotation` attribute sets the colors according to the wanted connotation.
```
yinonov, line 42:
```suggestion

Use the `connotation` attribute to set the avatar's color.
```

yinonov (changes requested):

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Slots
+### Default
+Use the `default` slot to have an image avatar.  
+When using an image make sure `appearance` is not `outlined`
+
+
+```html preview
+<vwc-avatar shape="pill" density="3">
+<img src="https://images.unsplash.com/photo-1567532939604-b6b5b0db2604?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=687&q=80" alt="woman"/>
```
yinonov, line 89: maybe something not as bloated?

yinonov (changes requested):

libs/components/src/lib/avatar/avatar.scss
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	&.shape-pill {
+		border-radius: 50%;
+	}
+
+
+	::slotted(*) {
+		block-size: 100%;
+		inline-size: 100%;
```
yinonov, line 57: wouldn't it damage the proportions? isn't object-fit enough?

yinonov: would a non-square image be ok?

yinonov: yes, checked.

libs/components/src/lib/avatar/README.md
yinonov: must it always be a human name?

libs/components/src/lib/avatar/avatar.scss
rachelbt: that what object-fit is for :)

libs/components/src/lib/avatar/README.md
rachelbt: Don't know... 
The thing is - if we have an image or an icon - it will be nice of us to help the user add the title. 
The only problem that I'm thinking of is with acronym.
Maybe we should leave this for now and decide about it after acronym in added.
WDYT?

libs/components/src/lib/avatar/avatar.scss
```diff
@@ -0,0 +1,61 @@
...
+
+	display: inline-flex;
+	overflow: hidden;
+	align-items: center;
+	justify-content: center;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	block-size: var(#{variables.$avatar-density});
+	color: var(#{appearance.get-appearance-token(text)});
+	inline-size: var(#{variables.$avatar-density});
+	outline: 1px solid var(#{appearance.get-appearance-token(outline)});
```
yinonov, line 31: I know I suggested it but now it feels hacky as outline accees top-layer. We should simply contain the image area to allow the box shadow to appear or simply border with box sizing. Although box Shadow is our convention

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,142 @@
...
+<vwc-avatar density='condensed' shape="pill"></vwc-avatar>
+<vwc-avatar density='normal' shape="pill"></vwc-avatar>
+<vwc-avatar density='extended' shape="pill"></vwc-avatar>
+```
+
+
+## Slots
+### Default
+Use the `default` slot to have an image avatar.  
+When using an image make sure `appearance` is not `outlined`
```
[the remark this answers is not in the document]
AyalaBu: I think it looks ok,
in case that it's outline appearance is consistent with the other avatar types

libs/components/src/lib/avatar/README.md
```diff
@@ -0,0 +1,148 @@
...
+### Acronym
+still in progress
+-->
+
+
+### Appearance
+Set the `appearance` attribute to change the avatar's appearance.
+
+- Type: `'filled'` | `'outlined'`
+- Default: `'filled'`
```
yinonov, line 40: Shouldn't ghost be default?

libs/components/src/lib/avatar/README.md
yinonov: This can fire back. Imagine we add our opinionated representation and automatic checks doesn't throw any violations due to it. Authors wouldn't be aware of the problem and that they need to add their context. We had that same issue with icon button. I think the best alternative is us handling docs with that in mind and provide snippets with aria context

yinonov (changes requested):

libs/components/src/lib/avatar/README.md
rachelbt: whats the meaning of ghost avatar? it looks strange just an icon or letters without bg or outline....

libs/components/src/lib/avatar/README.md
rachelbt: OK then. we will stay with our text about it in the docs :)

yinonov (approved):

PR #603, opened 2022-08-04
chore(docs): add aria-current to main side-drawer

rinaok:


yinonov (changes requested):
by replicating the same code to both layout pages in a script we ignore a principal of DRY.
perhaps we can create a new js dedicated file for this purpose. I expect it to scale as we might even need to consider some links are disclosed by default and will need to be undisclosed on load

rinaok:
> 

@yinonov 

I added a scrollTop for the sideDrawer, now it saves the last scrollTop in localStorage and know where to scroll.

We don't have disclosures yet so I think that it's fine for now, and when the day will come we can add some more functionalities.

yinonov:
> > 
> 
> 
> 
> @yinonov 
> 
> 
> 
> I've added a scrollTop for the sideDrawer, now it saves the last scrollTop in localStorage and know where to scroll.
> 
> 
> 
> We don't have disclosures yet so I think that it's fine for now, and when the day will come we can add some more functionalities.

Come to think of it, this is a common use case for SPA. Could there be a library servicing this specifically? Maybe we can unload this to a 3rd party

apps/docs/assets/scripts/scroll-handler.js
```diff
@@ -8,6 +8,20 @@ const updateHeaderElevationShadow = (isShadowed) => {
...
+}
+
+const setScrollFromLocalStorage = (target) => {
+	// set sideDrawer's scroll from localStorage
+	let scrollTop = localStorage.getItem("scroll");
+	if (scrollTop != null) {
+		setTimeout(() => {
+			target.scrollTop = scrollTop;
+		}, 1);
+	}
```
yinonov, line 24: what for?

apps/docs/_includes/navigation.njk
```diff
@@ -1,14 +1,14 @@
 {% block content %}
```
yinonov, line 1: does links scroll into view by default with the aria-current? 🤔

rinaok: @yinonov no

apps/docs/assets/scripts/scroll-handler.js
rinaok: it saves the scroll, I need to change it to scrollIntoView

apps/docs/pages/designs.njk
```diff
@@ -13,4 +13,10 @@ permalink: "designs/{{ design.title | slug }}/"
 {% endfor %}
 
 
-{% renderFile design.markdown %}
\ No newline at end of file
+{% renderFile design.markdown %}
+
+<script>
```
yinonov, line 18: repeating the same code is probably anti-pattern.
you can, alternatively, observe [mutations to the DOM](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)

yinonov (changes requested):

apps/docs/pages/designs.njk
yinonov: maybe a less "expensive" implementation would be to simply run query selector whenDefined (`vwc-nav-item`)

rinaok: @yinonov added a navigation handler

rachelbt (approved):
LGTM
Love the fact that we have indication to which component are we seeing in the docs 🥳

PR #604, opened 2022-08-05
chore(docs): consider bottom safe area for aside

yinonov:


apps/docs/assets/styles/main.css
```diff
@@ -28,25 +29,22 @@ a:visited {
...
 }
 
-#menu-button:defined {
-  display: block;
-  position: fixed;
-  inset: auto auto 8px 8px;
-  z-index: 2;
+#menu-button:not(:defined) {
+	display: none;
 }
```
rachelbt, line 40: lines 31-40 all this is not needed anymore - the button is now in the header

apps/docs/assets/styles/main.css
```diff
@@ -68,6 +66,7 @@ vwc-side-drawer::part(base) {
   block-size: calc(100vh - var(--_header-block-size));
   bottom: 0;
   top: var(--_header-block-size);
+	padding-block-end: calc(40px + env(safe-area-inset-bottom));
```
rachelbt, line 69:
```suggestion

```
this is not needed anymore - the fab button is not there anymore...

apps/docs/_layouts/base.njk
```diff
@@ -4,7 +4,7 @@
   <head>
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta name='viewport' content='initial-scale=1, viewport-fit=cover'>
```
rachelbt, line 7: 👍

apps/docs/assets/styles/main.css
yinonov:
```suggestion
	padding-block-end: env(safe-area-inset-bottom);
```

apps/docs/assets/styles/main.css
```diff
@@ -14,9 +14,10 @@ html {
 }
 
 body {
-  margin: 0;
+	margin: 0;
   background-color: var(--vvd-color-canvas);
   color: var(--vvd-color-on-canvas);
+	direction: ltr;
```
yinonov, line 20:
```suggestion
  margin: 0;
  background-color: var(--vvd-color-canvas);
  color: var(--vvd-color-on-canvas);
```

PR #609, opened 2022-08-09
chore(docs): move open close button to header

rachelbt:


rinaok (approved):

yinonov:
@rachelbt button's icon only toggled on button click. Instead it should be toggled on drawer state change.
You'll notice the problem on light dismiss (when scrim is clicked)

PR #612, opened 2022-08-10
refactor(listbox-option): migrate to fast

rinaok:


libs/components/src/lib/list-item/list-item.template.ts
```diff
@@ -27,17 +24,20 @@ export const ListItemTemplate: (
...
 		aria-setsize="${x => x.ariaSetSize}"
 		role="option">
-		${x => affixIconTemplate(x.icon)}
+		<slot name="start">${x => affixIconTemplate(x.icon)}</slot>
 		<div class="text-content">
 			${when(x => x.textPrimary, html`<div class="text-primary">${x => x.textPrimary}</div>`)}
 			${when(x => x.textSecondary, html`<div class="text-secondary">${x => x.textSecondary}</div>`)}
 		</div>
-		<slot name="meta"></slot>
+		<slot name="end"><slot name="meta"></slot></slot>
```
yinonov, line 41: 2 slots?

libs/components/src/lib/list-item/list-item.template.ts
```diff
@@ -27,17 +24,20 @@ export const ListItemTemplate: (
...
+		?checked="${(x) => x.checked}"
+		?disabled="${(x) => x.disabled}"
 		aria-checked="${x => x.ariaChecked}"
 		aria-disabled="${x => x.ariaDisabled}"
 		aria-posinset="${x => x.ariaPosInSet}"
 		aria-selected="${x => x.ariaSelected}"
 		aria-setsize="${x => x.ariaSetSize}"
 		role="option">
-		${x => affixIconTemplate(x.icon)}
+		<slot name="start">${x => affixIconTemplate(x.icon)}</slot>
```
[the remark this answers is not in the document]
rinaok: @yinonov We don't have to use them.
It's from fasts listbox-option api:

Screen Shot 2022-08-17 at 12 21 19

libs/components/src/lib/list-item/list-item.template.ts
rinaok: @yinonov I wanted to use fast's slots from their API but didn't want to use the name "end" for our api.
Will it be better to leave just our slots?

image

rinaok:
@yinonov @jshenkman @AyalaBu 

This is how the button inside a list-item looks in vivid-2, how should it look in vivid-3?

image

libs/components/src/lib/listbox-option/listbox-option.template.ts
```diff
@@ -0,0 +1,47 @@
...
+	return html`
+	<div class="${getClasses}">
+		<div class="option-wrapper"
+			?selected="${(x) => x.selected}"
+			?disabled="${(x) => x.disabled}"
+			aria-disabled="${x => x.ariaDisabled}"
+			aria-posinset="${x => x.ariaPosInSet}"
+			aria-selected="${x => x.ariaSelected}"
+			aria-setsize="${x => x.ariaSetSize}"
+			role="option">
```
rachelbt, line 37: maybe all these one needs to move to the upper div as the css classes moved

libs/components/src/lib/listbox-option/listbox-option.scss
```diff
@@ -0,0 +1,88 @@
...
+	hyphens: auto;
+	min-block-size: $min-block-size;
+	padding-block: 10px;
+	padding-inline: $gap;
+	text-decoration: none;
+	vertical-align: middle;
+	word-break: break-word;
+
+	.option-wrapper {
+		display: inline-flex;
```
rachelbt, line 39:
```suggestion
		display: flex;
```

libs/components/src/lib/listbox-option/listbox-option.template.ts
rinaok: Yes I'll move it

libs/components/src/lib/listbox-option/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import '../icon';
+
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { designSystem } from '../../shared/design-system';
+import { ListboxOption } from './listbox-option';
+import { ListboxOptionTemplate as template } from './listbox-option.template';
+import styles from './listbox-option.scss';
+
+export const vividListboxOption = ListboxOption.compose<FoundationElementDefinition>({
+	baseName: 'listbox-option',
```
yinonov, line 11: should simply be named option as is also used within select component
```suggestion
	baseName: 'option',
```

yinonov (changes requested):

PR #616, opened 2022-08-11
chore(form-element): set the form field pattern

YonatanKra:


yinonov (commented):
LGTM but, I assume additional members aren't covered by _fast_? should we expect this addition by fast anytime soon?

yinonov (dismissed):

libs/components/src/lib/text-field/README.md
```diff
@@ -147,10 +147,12 @@ In addition, we programatically *"dirtied"* the field and then called `validate`
 <vwc-text-field pattern="123"></vwc-text-field>
 
 <script>
+setTimeout(() => {
```
yinonov, line 150: did u experience RC issues here? rendering related? of its esm being async?

yinonov (dismissed):

yinonov (approved):

PR #630, opened 2022-08-25
chore(components): fonts cdn assign to self asset

yinonov:
BREAKING CHANGE: fonts fetched from cdn should now be included as assets in authoring scope


rinaok (approved):

libs/components/project.json
```diff
@@ -131,7 +131,7 @@
   },
   "tags": [],
   "implicitDependencies": [
-		"styles",
-		"tokens"
-	]
+    "styles",
+    "tokens"
+  ]
```
yinonov, line 136:
```suggestion
		"styles",
		"tokens"
	]
```

PR #634, opened 2022-08-28
chore(components): implement new note design

### LINKED ISSUE #581: [note] implement re-design

yinonov:
Closes #581


yinonov:
image

there is an initiative to conduct a convention for how rectangles are designed across the components system (_badge_, _button_, _nav-item_, _text field_, etc').

I call it **appearance model** and it features the following predefined categories -

- filled
- subtle
- outlined
- duotone
- ghost

each of them may require additional subset of states like - _idle_, _disabled_, _hover_, _active_, _selected_ etc’.

This was never thoroughly progressed by design but is implemented properly in development.

Aligning to that convention by setting the note to a _subtle_ appearance value produces the following

image

with its current design, the note doesn't fall under any of the available categories in the set. It isn’t really close to any of the above which should raise a flag or at least a discussion.
an natural action should either be **creating a new value for a rectangle model** (for this new design to fit in), or we **align the note to an existing value** and change it's new appearance to comply with what is already part of the convention.

Big question is - does its new appearance have any common ground with the rest? was it design arbitrarily? was it designed with the above described convention in mind?

@AyalaBu @rachelbt

rachelbt (approved):
LGTM

PR #635, opened 2022-08-28
feat(listbox): add new component

rinaok:
after https://github.com/Vonage/vivid-3/pull/760 will be merged


yinonov (changes requested):
@rinaok there are 2 issues with focus needs fixing
- [ ] clicking on listbox items doesn't activate (focus) listbox
- [ ] when extending listboxElement, listbox element isn't activated (focused). extending listbox enabling it to be activated

yinonov:
listbox options, when are selected / checked, are styled by background color only. does it comply with rules? should we add another indicator?

image

libs/components/src/lib/listbox-option/listbox-option.scss
```diff
@@ -3,14 +3,14 @@
...
-	$shades: primary faint soft,
+	$shades: primary primary-increment faint soft,
 	$default: accent
 );
 @use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
 
 @use "../../../../shared/src/lib/sass/mixins/appearance/config" as appearance-config with (
 	$appearances: ghost,
-	$states: idle hover disabled active selected,
+	$states: idle hover disabled active selected selectedAndHover,
```
rinaok, line 13: There is a new PR for option, it's better to make the changes there. 
https://github.com/Vonage/vivid-3/pull/760

libs/components/src/lib/listbox-option/listbox-option.scss
```diff
@@ -23,6 +23,7 @@
 	@include connotation.connotation;
 	@include appearance.appearance;
 
+	position: relative;
```
rinaok, line 26: @rachelbt @yinonov This had to be returned. Without it, the `Multiple selection` is not working.

libs/components/src/lib/listbox/listbox.scss
```diff
@@ -0,0 +1,43 @@
...
+
+@use '../../../../shared/src/lib/sass/mixins/appearance/config' as appearance-config with (
+	$appearances: fieldset ghost,
+	$states: idle disabled,
+	$default: fieldset,
+);
+@use '../../../../shared/src/lib/sass/mixins/appearance' as appearance;
+
+:host {
+	display: inline-flex;
```
rachelbt, line 17:
```suggestion

```
I don't think this is needed, and this prevent the list-box getting its parent width

libs/components/src/lib/listbox-option/listbox-option.scss
yinonov: and the focus indicator needs it as well

libs/components/src/lib/listbox/listbox.ts
```diff
@@ -0,0 +1,50 @@
...
+	@attr appearance?: LisboxAppearance;
+
+	override slottedOptionsChanged(prev: Element[] | undefined, next: Element[]) {
+		super.slottedOptionsChanged(prev, next);
+		this.#disableSlottedChildren();
+	}
+
+	override attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
+		super.attributeChangedCallback(name, oldValue, newValue);
+		switch (name) {
```
olaf-k, line 35: a simple `if` should be enough for the time being

olaf-k (changes requested):

olaf-k (dismissed):
👍

yinonov (changes requested):
wait for #750

rinaok:
> wait for #750

merged

yinonov (changes requested):
mouse click on an option should set focus on listbox. additional keyboard events like arrow navigation should continue navigating through other options but it doesn't

rinaok:
> mouse click on an option should set focus on listbox. additional keyboard events like arrow navigation should continue navigating through other options but it doesn't

@yinonov @rachelbt If I return [this](https://github.com/Vonage/vivid-3/pull/635/commits/1c0ca1ffcdfda0e5164c722584d57ce957082eda) , it fixes the problem.

rachelbt (approved):
LGTM

rinaok:
@yinonov The [refactor](https://github.com/Vonage/vivid-3/commit/2e55fe2331d0d07bd0a88650a3cbf4586a308499#diff-590cb97e4da80dd05a7a1d55cd2d0a3ab33e35cdc20e565a834b949c622a63e7) in fieldset broke the listbox. Can you take a look?

yinonov (approved):

PR #638, opened 2022-08-30
test(components): deterministic layout size in action group doc

yinonov:


libs/components/src/lib/action-group/README.md
```diff
@@ -71,10 +71,10 @@ enabling the `tight` member will result in a dense style that fits the "normal"
 ```html preview
 <style>
   vwc-layout {
-    --layout-grid-template-columns: 35ch;
+    --layout-grid-template-columns: 350px;
   }
 
-  vwc-action-group > vwc-text-field {
+  vwc-action-group > vwc-text-field
```
rachelbt, line 77:
```suggestion
  vwc-action-group > vwc-text-field {
```

yinonov: right!

PR #646, opened 2022-09-01
feat(styles): modify typography desktop style

yinonov:


libs/styles/src/lib/typography/desktop.scss
```diff
@@ -1,39 +1,39 @@
...
-// 	font-size: 62.5%;
-// }
+$selectors: (
+	bold: 'b, strong',
+	code: 'pre, var, code, kbd, samp',
+);
 
 :root {
 	@include typography.variables;
 	font-size: 62.5%;
```
yinonov, line 15:
```suggestion
```

rachelbt (commented):
the icon in the text-field gets now font-size of 14 and not 16 because nothing is set on it.
All of the snap shor width (and some of the also the height) were change - that strange that it happens - it isn't related to the change you made...

yinonov:
> the icon in the text-field gets now font-size of 14 and not 16 because nothing is set on it. All of the snap shor width (and some of the also the height) were change - that strange that it happens - it isn't related to the change you made...

1. text field's icon size was never set - #655
2. it must have something to do with the change I made here, I ran ui tests in parallel on main

yinonov:
@rachelbt `font-size` and `line-height` both affect block spacing

https://user-images.githubusercontent.com/15140652/188968708-78a2ac71-0d2a-4226-9ade-007d36317e1c.mov

rachelbt (approved):
went over the snapshots again - some were changed by png width & height as you explained in the screen recording because of plain text (with no vivid typography set on it) inside the component  that was 16px  and now its base - 14px.

rachelbt (approved):

PR #651, opened 2022-09-05
refactor(styles): separate concerns from styles to tokens

yinonov:
diffrentiate harmless style tokens from styles that affect the application (side-effects)

this describes the updated interface and the scope of this PR
https://github.com/Vonage/vivid-3/blob/bf36987ea9a95e0a9ce9f55e8c89a3efa2289fd7/libs/components/README.md#L33-L65


yinonov:
That's an issue as now we never apply styles (including `color-scheme`) on root 

Fixed color-scheme to not propagate to the viewport background if set on the <body> ([253565@main](https://commits.webkit.org/253565@main))

change/@vonage-vivid-91a27727-cccf-41bd-816d-5640377c5c22.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "refactor styles to separate concerns from styles and tokens",
```
yinonov, line 3:
```suggestion
  "comment": "refactor styles to separate concerns from styles to tokens",
```

rachelbt (approved):

PR #655, opened 2022-09-06
refactor(components): define icon size in text field

yinonov:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -126,6 +126,7 @@ $low-ink-color: --_low-ink-color;
 	z-index: 1;
 	color: var(#{$low-ink-color});
 	inset-inline-start: 16px;
+	font-size: 20px;
```
rachelbt, line 129:
```suggestion
  font-size: 20px;
	inset-inline-start: 16px;
```

rachelbt: to fix lint issue

rachelbt (approved):

PR #656, opened 2022-09-07
chore(docs): restore heading 1 style

yinonov:


yinonov:
@rachelbt I noticed the heading 1 style didn't apply properly and tweaked it as I didn't understand the original style. mind reviewing?

rachelbt:
> @rachelbt I noticed the heading 1 style didn't apply properly and tweaked it as I didn't understand the original style. mind reviewing?

the 

> @rachelbt I noticed the heading 1 style didn't apply properly and tweaked it as I didn't understand the original style. mind reviewing?

In @jshenkman design the headline is starched like this - this is the reason for the negative margin:
image

Also - now with the change we need a way to target the H1 in homepage.

yinonov:
> > @rachelbt I noticed the heading 1 style didn't apply properly and tweaked it as I didn't understand the original style. mind reviewing?
> 
> the
> 
> > @rachelbt I noticed the heading 1 style didn't apply properly and tweaked it as I didn't understand the original style. mind reviewing?
> 
> In @jshenkman design the headline is starched like this - this is the reason for the negative margin: image
> 
> Also - now with the change we need a way to target the H1 in homepage.

this design isn't realistic, there's no fly-out menu, and no tabs

PR #668, opened 2022-09-08
chore(workspace): add repository to package.json

olaf-k:
Just to stop beachball complaining.
`Valid "repository" key not found in "package.json". Consider adding this info for more accurate git remote detection.`


yinonov:
just out curiosity, what interest does beachball have with the root according to their docs? it should focus on the nested scopes

olaf-k:
> just out curiosity, what interest does beachball have with the root according to their docs? it should focus on the nested scopes

I attributed the message to beachball as it always appears when invoked but couldn't find any trace of it in their code, so maybe it isn't the culprit? (but then what else?) otherwise, since it can be used to publish to npm, maybe it's part of some checklist?

olaf-k:
> > just out curiosity, what interest does beachball have with the root according to their docs? it should focus on the nested scopes
> 
> I couldn't find the error message in beachball's code, so it's probably using a dependency to automatically run some checks on the repo (as it can be used to publish to npm).

it's a check run by `workspace-tools` (I'm aware this doesn't answer the "why" question though).

yinonov:
> > > just out curiosity, what interest does beachball have with the root according to their docs? it should focus on the nested scopes
> > 
> > 
> > I couldn't find the error message in beachball's code, so it's probably using a dependency to automatically run some checks on the repo (as it can be used to publish to npm).
> 
> it's a check run by `workspace-tools` (I'm aware this doesn't answer the "why" question though).

it does no harm anyway and a legitimate property

yinonov (approved):

olaf-k:
> just out curiosity, what interest does beachball have with the root according to their docs? it should focus on the nested scopes

This was bugging me so I had a look and FWIW: beachball lists changes in the repo (git diff), then tracks them back to their package.json (parent component) and decides whether it should take them into account based on the `private` and `beachball.shouldPublish` properties in said package file. Changes to the doc are not taken into account by beachball because the main package.json lists the repo as private. (You probably already know all this so that's more of a note for myself :)

PR #671, opened 2022-09-09
perf(components): add focus delegation to interactive elements

yinonov:


libs/components/src/lib/nav/index.ts
```diff
@@ -7,7 +7,7 @@ import { NavTemplate as template } from './nav.template';
 export const vividNav =
 	Nav.compose<FoundationElementDefinition>({
 		baseName: 'nav',
-		template: template as any,
+		template: template as any
```
yinonov, line 10:
```suggestion
		template: template as any,
```

olaf-k (approved):
LGTM

PR #673, opened 2022-09-10
chore(components): fix text area vertical padding

yinonov:


rachelbt (commented):
👍 
You are missing the beachball file...

rachelbt (commented):
something is strange with the box-shadow "border" in firefox. It doesn't happen  if I open firefox

PR #676, opened 2022-09-12
chore(popup): add light dismiss

rinaok:


libs/components/src/lib/popup/README.md
```diff
@@ -69,6 +69,26 @@ Add the `dismissible` attribute to add a close button to the popup.
 </script>
 ```
 
+### Light Dismiss
+
+Add the `light-dismiss` attribute (or `lightDismiss` property) to close the popup when the user taps on any area other than the popup.
```
yinonov, line 74: There's more to light dismiss than click outside

https://open-ui.org/components/popup.research

(Search page for light dismiss)

yinonov (changes requested):

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -131,6 +131,8 @@ describe('vwc-popup', () => {
 				.toBeFalsy();
 			expect(element.dismissible)
 				.toBeFalsy();
+			expect(element.lightDismiss)
+				.toBeFalsy();
```
yinonov, line 135:
```suggestion
				.toEqual(false);
```

yinonov (changes requested):

yinonov (changes requested):
Finger touch outside the popup doesn't close it

libs/components/src/lib/popup/README.md
rinaok: @yinonov When the anchor is in focus, pressing ESC closes the popup. I am not sure it is part of the light dismiss. I made a research, don't remember all of it, but tooltip should close on ESC. Here for example: https://www.lightningdesignsystem.com/accessibility/patterns/tooltip/

rinaok: It is described in tooltip docs: https://vivid.deno.dev/components/tooltip#accessibility
I can add it also to popup

rinaok:
> Finger touch outside the popup doesn't close it

switched `click` to `mousedown`

libs/components/src/lib/popup/README.md
yinonov: > When the anchor is in focus, pressing ESC closes the popup

that might actually be a bug.

there's more than ESC as well, might be resize / scroll.

I suggest we look for an implementation we can rely on instead of patching it all in. if there's no such implementation, we would gradually develop this

yinonov: [this polyfill](https://github.com/oddbird/popup-polyfill) might already do

PR #679, opened 2022-09-13
fix(documentation): native attributes

rachelbt:


libs/components/src/lib/text-field/README.md
```diff
@@ -1,6 +1,7 @@
-# Text Field
+789# Text Field
```
rinaok, line 1:
```suggestion
# Text Field
```

rinaok (changes requested):

libs/components/src/lib/dialog/README.md
```diff
@@ -1,4 +1,7 @@
 # Dialog
+Represents a part of an application that a user interacts with to perform a task
+All native attributes of `checkbox` are supported as well as some enhancements.
```
rinaok, line 3: checkbox?

rinaok (changes requested):

libs/components/src/lib/dialog/README.md
rachelbt: > checkbox?

```suggestion
All native attributes of `dialog` are supported as well as some enhancements.
```

yinonov (approved):

yinonov (approved):

PR #682, opened 2022-09-15
[initials]: new underlying component

rachelbt:


yinonov (commented):
I don't recall if we discussed this as an independent component. it could rather be a service / utility.

have any other examples for components using initials?

libs/components/src/lib/initials/initials.template.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param root0.text
+ */
+
+
+
+const InitialTrim = ({
+	text}: Initials) => {
+	if (text === undefined) return '';
+	const theName = text;
+	const letterTrimming = theName.split(' ')[0].charAt(0).toUpperCase() + theName.split(' ')[1].charAt(0).toUpperCase();
```
yinonov, line 25: there could probably a stable open source service out-there. let's adopt tooling here rather than try and maintain assertion, casing, formatting etc'.

yinonov (changes requested):

libs/components/src/lib/initials/initials.template.ts
rachelbt: something like this:
https://www.npmjs.com/package/initials
?

yinonov: Exactly, although I don't see support for more than 2 names. Some people have more. Are there any others? If not let's start with it and even contribute

rachelbt:
> I don't recall if we discussed this as an independent component. it could rather be a service / utility.
> 
> have any other examples for components using initials?
we talked about it as a separate component but related to the avatar.
Maybe its more accurate to call it - avatar-initials

libs/components/src/lib/initials/initials.template.ts
rachelbt: I found 2 others but the last update was more that 5 years ago...
I don't think we need to support more than 2 letters. It will break the design.

rachelbt:
no need for a separate component.

PR #692, opened 2022-09-20
chore(components): improve breadcrumb item focus indication style

yinonov:


libs/components/src/lib/breadcrumb-item/breadcrumb-item.scss
```diff
@@ -35,6 +35,9 @@
 }
 
 .focus-indicator {
+	--focus-inset: -4px -8px;
```
rachelbt, line 38:
```suggestion
	--focus-inset: -4px -8px;
	--focus-stroke-gap-color: transparent;
```

that it will be like this:
image

and not:
image

like we have in appearance-outlined

rachelbt (approved):
LGTM :)

PR #696, opened 2022-09-22
feat(components): add radio button

olaf-k:


libs/components/package.json
```diff
@@ -1,50 +1,51 @@
 {
```
yinonov, line 1: can u align with the existing format (space / tabs)? just not to generate irrelevant changes

libs/components/src/lib/radio/radio.template.ts
```diff
@@ -0,0 +1,42 @@
...
+
+	return html<Radio>`
+	<span class="${getClasses}"
+		role="radio"
+		aria-checked="${x => x.checked}"
+		aria-required="${x => x.required}"
+		aria-disabled="${x => x.disabled}"
+		aria-readonly="${x => x.readOnly}"
+		@keypress="${(x, c) => x.keypressHandler(c.event as KeyboardEvent)}"
+		@click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
```
yinonov, line 34: these belong on the control

yinonov (changes requested):

yinonov (changes requested):
should also add exporting from `components.ts`

libs/components/package.json
olaf-k: sure, no idea what happened here.

libs/components/src/lib/radio/radio.template.ts
olaf-k: Shouldn't all this apply to the entire component, control + label included? The checkbox is also designed like this.

yinonov: true 🤔

libs/components/src/lib/radio/README.md
```diff
@@ -0,0 +1,56 @@
+# Radio
+
+Represents a radio custom element.
+
+Radio components on their own do not make sense. They should always be used inside `radio-group` containers.
```
yinonov, line 5: maybe quote an aria definition

libs/components/src/lib/radio/index.ts
```diff
@@ -10,7 +10,10 @@ import { RadioTemplate as template } from './radio.template';
 export const vividRadio = Radio.compose<RadioOptions>({
 	baseName: 'radio',
 	template: template as any,
-	styles
+	styles,
+	shadowOptions: {
```
yinonov, line 14: what made u eventually choose focus delegation?

libs/components/src/lib/radio/radio.scss
```diff
@@ -48,16 +48,21 @@ $selected-border: ($selected-size - 10px) * 0.5;
...
 		height: $selected-size;
+		background-color: var(#{appearance.get-appearance-token(outline)});
 		border-radius: 50%;
 		box-shadow: inset 0 0 0 $selected-border var(#{appearance.get-appearance-token(text)});
 		content: '';
+	}	
+
+	.base:where(:checked, .checked):where(:disabled, .disabled) &::after,
+	.base:where(.readonly):where(:not(:disabled, .disabled, :hover, .hover, :active, .active)) &::after {
+		box-shadow: inset 0 0 0 $selected-border var(#{appearance.get-appearance-token(fill)});
```
yinonov, line 69: this is already handled by the appearance mixin

https://github.com/Vonage/vivid-3/blob/3720d1b346de4c732b72266b2e6520c7a8ea3fb5/libs/shared/src/lib/sass/mixins/appearance/_mixins.scss#L12-L31

yinonov (changes requested):

libs/components/src/lib/radio/index.ts
olaf-k: this avoids the vwc-radio from getting focus (outlining control + label) and keep it around the control

libs/components/src/lib/radio/radio.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { createFormHTML, elementUpdated, fixture, getBaseElement, listenToFormSubmission } from '@vivid-nx/shared';
```
yinonov, line 1: in general, u can review and even extract tests from https://github.com/microsoft/fast/blob/8711e8ad82f13c6c67007b43f910b72c27494343/packages/web-components/fast-foundation/src/radio/radio.pw.spec.ts

libs/components/src/lib/radio/radio.scss
olaf-k: well that's what I thought but I spent a loooong time trying to figure out a combination of appearance tokens that would match all the states to no avail.
this was the least verbose override I could come up with.

libs/components/src/lib/radio/radio.spec.ts
olaf-k: 👍 will have a look!

libs/components/src/lib/radio/radio.scss
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	@supports (user-select: none) {
+		user-select: none;
+	}
+
+	&:disabled {
+		cursor: not-allowed;
+	}
```
yinonov, line 43:
```suggestion

	@supports selector(:focus-visible) {
		&:focus {
			outline: none;
		}
	}

	@supports (user-select: none) {
		user-select: none;
	}

	&:not(.disabled) {
		cursor: pointer;
	}

	&.disabled {
		cursor: not-allowed;
	}
```

libs/components/src/lib/radio/radio.scss
```diff
@@ -0,0 +1,82 @@
...
+@use '../../../../shared/src/lib/sass/mixins/connotation' as connotation;
+
+@use '../../../../shared/src/lib/sass/mixins/appearance/config' as appearance-config with (
+	$appearances: fieldset,
+	$states: idle hover checked disabled readonly,
+	$default: fieldset,
+);
+@use '../../../../shared/src/lib/sass/mixins/appearance' as appearance;
+
+$control-size: 20px;
```
rachelbt, line 19: in Vivid-2 the radio size is 22px and the gap is 10px, in Figam the size is 20px and gap is 8px. 
@yinonov which design are we following here?

yinonov: most consistent form is 20px
https://www.figma.com/file/r8OYLPmRmDqfLKt6KHdJDB/Vivid-UI-Kit---2.0?node-id=1987%3A24620

rachelbt: than gap need to be 8px :)

libs/components/src/lib/radio/radio.scss
```diff
@@ -0,0 +1,82 @@
...
+$control-border: 2px;
+
+.base {
+	@include connotation.connotation;
+	@include appearance.appearance;
+
+	display: inline-flex;
+	align-items: center;
+	cursor: pointer;
+	gap: 10px;
```
rachelbt, line 29:
```suggestion
	gap: 8px;
```

olaf-k:
a11y validation fails because Axe considers that radio components should not be made read-only (but doesn't complain for checkboxes 🤷‍♀️)

libs/components/src/lib/radio/radio.template.ts
```diff
@@ -0,0 +1,43 @@
...
+ * The template for the {@link @microsoft/fast-foundation#Radio} component.
+ *
+ * @param context
+ * @public
+ */
+export const RadioTemplate: (context: ElementDefinitionContext) => ViewTemplate<Radio> = (context: ElementDefinitionContext) => {
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<Radio>`
+	<span class="${getClasses}"
```
yinonov, line 27: mind the pitfall where radio-group sets certain attributes (instead of a better practice of properties) on host that correlate with other aria attributes which here are set on inner part.
if they are not set on the same element, they might not make any sense to assistive technology.
we might wanna add dedicated a11y unit tests on radio group to validate such mutations

libs/components/src/lib/radio/radio.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+			expect(classes.contains('disabled')).toBeTruthy();
+		});
+	});
+
+	describe('readonly', () => {
+		it('should set read only class when readonly is true', async () => {
+			const classes = await setBoolAttributeOn(element, 'readonly');
+			expect(classes.contains('readonly')).toBeTruthy();
+		});
+	});
```
yinonov, line 84: remove?

yinonov (changes requested):

libs/components/src/lib/radio/radio.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+		it('should set disabled class when disabled is true', async () => {
+			const classes = await setBoolAttributeOn(element, 'disabled');
+			expect(classes.contains('disabled')).toBeTruthy();
+		});
+	});
+
+	describe('readonly', () => {
+		it('should set read only class when readonly is true', async () => {
+			const classes = await setBoolAttributeOn(element, 'readonly');
+			expect(classes.contains('readonly')).toBeTruthy();
```
yinonov, line 82: as radio button won't support `readonly`, it doesn't make sense to test it here.

instead, test readonly functionality within radio group. when readio group is set to readonly, any interaction to its radio buttons won't change anything

yinonov (changes requested):

yinonov (approved):

PR #697, opened 2022-09-22
fix(components): consistent size for checkbox and radio

olaf-k:
The current Figma file makes the checkbox and the radio controls 20×20px with a 10px gap from the label.
I can either update [my current PR for the radio](https://github.com/Vonage/vivid-3/pull/696) and align it with the checkbox or merge this to change the checkbox.


libs/components/src/lib/checkbox/checkbox.scss
```diff
@@ -67,9 +72,13 @@ $gutter: 4px;
 }
 
 label {
-	color: var(#{constants.$vvd-color-canvas-text});
+	color: var(#{appearance.get-appearance-token(text)});
```
olaf-k, line 75: see https://github.com/Vonage/vivid-3/pull/696#discussion_r983487597

rachelbt: this causes the text to be white when the chackbox is checked:

rachelbt (commented):
you are missing ui test png's :)

libs/components/src/lib/checkbox/checkbox.scss
yinonov: this PR should scope the consistence in sizing. I created another PR to modify checkbox styling Vonage/vivid-3#705

yinonov (changes requested):

yinonov (approved):

PR #699, opened 2022-09-22
refactor(components): migrate tabs to fast

yinonov:


AyalaBu:
@rinaok 
In the vertical tabs component "hover"/"selected" indication must always be the same width,
like in this example:

Screenshot 2022-11-23 at 13 26 39

This is the link to Figma:
https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19097%3A81830&t=pYTzgCh1RB4ypPYV-1

AyalaBu:
@rinaok 
I made some changes to the info connotation in tabs text 
for it to be in line with the other elements with ghost appearance,
Please take a look and change accordingly:

https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19097%3A81791&t=n460dkt4vqkK2LqH-1

https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19097%3A81745&t=n460dkt4vqkK2LqH-1

libs/components/src/lib/tab/tab.scss
```diff
@@ -0,0 +1,69 @@
...
+}
+
+.base {
+	$min-block-size: 48px;
+	$gutter: 8px;
+
+	@include appearance.appearance;
+
+	position: relative;
+	display: inline-flex;
```
rachelbt, line 29:
```suggestion
	display: flex;
```

rinaok:
@rachelbt @yinonov @AyalaBu 

I think that the PR is ready :)

AyalaBu:
@rinaok @yinonov 
The developed tabs don't look like a design I created at all.

These are the links to Figma:
https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19097%3A81745&t=n460dkt4vqkK2LqH-1

https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19097%3A81791&t=n460dkt4vqkK2LqH-1

The appearance is ghost, 
with bg 50 in hover state, and 100 in enabled state.
This pattern should repeat in all the ghost component.

Also the indicator is way under the tab itself,
it's supposed to be 1px in & 1px out.

Also,
the padding is not like in design at all.

rinaok (approved):

rinaok (approved):

libs/components/src/lib/components.ts
```diff
@@ -37,3 +37,4 @@ export * from './switch/definition';
 export * from './text-field/definition';
 export * from './text-area/definition';
 export * from './tooltip/definition';
+export * from './tabs/definition';
```
yinonov, line 40: what about `tab` and `tab-panel`?

rinaok (approved):

libs/components/src/lib/tab-panel/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## CSS Variables
+
+## Events
+
+## Methods
+
+## Accessibility
+
+## Use Cases
```
rachelbt, line 27:
```suggestion

```

rachelbt (commented):
@rinaok - I went over and remembered why this is not done yet :) 
visual things that are still need to be tended:

- the focus ring need to be on only when keyboard navigation
- not sure about the vertical design. Maybe we should remove it for now? in Vivid-2 we do not have it
- the selected underline transition is not fixed yet (get offset width and position and after a split second its OK)

rinaok:
> @rinaok - I went over and remembered why this is not done yet :) visual things that are still need to be tended:
> 
> * the focus ring need to be on only when keyboard navigation
> * not sure about the vertical design. Maybe we should remove it for now? in Vivid-2 we do not have it
> * the selected underline transition is not fixed yet (get offset width and position and after a split second its OK)

@rachelbt @yinonov 

1. Why the focus ring should be only on navigation? In vwc-option for example the focus ring is activated on click.
2. Why should we remove the vertical design? I don't remember that we talked about it.
3. I think that the selected underline should be fixed with css, right?

yinonov (commented):
Tab-panel documentation is missing. There's no recollection of it, what it's for or how to use it

yinonov (commented):
Tab-panel documentation is missing. There's no recollection of it, what it's for or how to use it

yinonov (commented):
there's an issue with tabs foundation enforcing `activeindicator` property which lets authors remove the indicator if they wish to.
we need to reinforce the indicator to always show

https://github.com/microsoft/fast/blob/6ebb5ec7bdb7666f49ddc524bc56f70a2e8aa574/packages/web-components/fast-foundation/src/tabs/tabs.ts#L139-L146

rinaok:
> Tab-panel documentation is missing. There's no recollection of it, what it's for or how to use it

rinaok:
> Tab-panel documentation is missing. There's no recollection of it, what it's for or how to use it

@yinonov I don't think that we should have tab-panel readme file. It cannot live without the tabs and the tab. Therefore, it will look exactly the same as tabs readme.

libs/components/src/lib/tabs/tabs.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.tablist {
+	#{variables.$tabs-active-indicator-stroke-width}: 2px;
+
+	position: relative;
+	display: grid;
+	box-sizing: border-box;
+
+	.base.orientation-vertical & {
+		box-shadow: 1px 0 0 0 var(--vvd-color-neutral-300);
```
yinonov, line 34: magic strings

libs/components/src/lib/tabs/tabs.scss
```diff
@@ -0,0 +1,77 @@
...
+		grid-template-columns: auto 1fr;
+		grid-template-rows: auto;
+		inline-size: 100%;
+		place-self: flex-start end;
+	}
+
+	.base:not(.orientation-vertical) & {
+		align-self: end;
+		justify-content: flex-start;
+		border-bottom: 1px solid var(--vvd-color-neutral-300);
```
yinonov, line 45: magic strings

yinonov:
partially patched up the janky transitioned active-indicator -
width is now determined by css variable assigned with the current active tab `getClientRect().width`.
The tabs container active-indicator is responsible for the smooth animation of transitioning from 1 active tab to another active tab when changed.
this active-indicator element is only visible when transitioned. this requires each of the individual child tabs to set their own active-indicator style when they're active. reason is to support edge case of a tab character data mutation. otherwise the animated active-indicator requires a mutation observer of character data. seem like a complete overkill when there are feasible css hacks around that.

there's still an issue of first change not well baked by fast.
indicator offsets unexpectedly outside horizontal boundaries. `.active-indicator` element seem to require an initial setting of
```css
grid-column: ...;
transform: translateX(...);
```
image

yinonov:
> > Tab-panel documentation is missing. There's no recollection of it, what it's for or how to use it
> 
> @yinonov I don't think that we should have tab-panel readme file. It cannot live without the tabs and the tab. Therefore, it will look exactly the same as tabs readme.

not a readme, but a mention in tabs, at least. on why is it there?

libs/components/src/lib/tabs/tabs.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @public
+ */
+export class Tabs extends FoundationTabs {
+	@observable tablist?: HTMLElement;
+
+	override orientationChanged(): void {
+		super.orientationChanged();
+		this.patchIndicatorStyleTransition();
+		if (this.orientation === 'vertical') {
```
olaf-k, line 17:
```suggestion
		if (this.orientation === TabsOrientation.vertical) {
```
(requires adding `TabsOrientation` to the imports)

libs/components/src/lib/tabs/tabs.scss
yinonov:
```suggestion
		border-bottom: 1px solid var(#{constants.$vvd-color-neutral-300});
```

rinaok:
@YonatanKra Do you have any idea why the test is failing?

olaf-k:
About the UT failure: you can't retrieve a custom CSS var, [it's a jsdom limitation](https://github.com/jsdom/cssstyle/issues/89#issuecomment-1316428531).

yinonov:
> > @YonatanKra Do you have any idea why the test is failing?
> 
> 
> 
> About the UT failure: you can't retrieve a custom CSS var, [it's a jsdom limitation](https://github.com/jsdom/cssstyle/issues/89#issuecomment-1316428531).

Oh. Is there a workaround?

libs/components/src/lib/tab/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Represents a tab custom element.
+
+```js
+<script type="module">
+    import '@vonage/vivid/tab';
+</script>
+```
+
+## Members
```
YonatanKra, line 11: Notice there are members that are not documented. Not here and not in the tests.
These are members that are exposed to the user and are in the [FAST documentation of tabs](https://github.com/microsoft/fast/tree/master/packages/web-components/fast-foundation/src/tabs#fields) like `activetab`.

YonatanKra (changes requested):
Notice my comments regarding missing APIs. Is this on purpose?

olaf-k:
There are two issues with the coverage:
1. Two paths not taken when `activeIndicatorRef?` is not defined, in `orientationChanged()` and `patchIndicatorStyleTransition()`. It seems that `activeIndicatorRef` is kept when `showActiveIndicator` is switched false (which is not gracefully handled by the component it seems.)
2. jsdom does not handle `getClientRects` (another limitation), you need to mock it.
```js
window.HTMLElement.prototype.getClientRects = function () {
    return [ { width: 10 } ] as unknown as DOMRectList;
}
```

libs/components/src/lib/tab/README.md
rinaok: We are missing `hideActiveIndicator`. I think that `activetab` should not be in the docs

rinaok: But I'm not sure we want to expose `hideActiveIndicator` as well, @yinonov what do you think?

olaf-k:
> > > @YonatanKra Do you have any idea why the test is failing?
> > 
> > 
> > About the UT failure: you can't retrieve a custom CSS var, [it's a jsdom limitation](https://github.com/jsdom/cssstyle/issues/89#issuecomment-1316428531).
> 
> Oh. Is there a workaround?

you mean like using Playwright? 😁

libs/components/src/lib/tab/README.md
yinonov: the fast version we use is still using `showActiveIndicator`. `hideActiveIndicator` is introduced in their alpha

yinonov: anyway, we don't want to expose it. correct

YonatanKra:
> > > > @YonatanKra Do you have any idea why the test is failing?
> > > 
> > > 
> > > About the UT failure: you can't retrieve a custom CSS var, [it's a jsdom limitation](https://github.com/jsdom/cssstyle/issues/89#issuecomment-1316428531).
> > 
> > 
> > Oh. Is there a workaround?
> 
> you mean like using Playwright? 😁

The problem with playwright is that it opens a new browser for every spec file.
It is slow and resource consuming (comes hand in hand).
That's what FAST is doing, and while working on the grid, I see a lot of issues with that (besides the fact their tests take a lot of time to run locally on my machine, with lots of flakiness).

olaf-k:
> The problem with playwright is that it opens a new browser for every spec file. It is slow and resource consuming (comes hand in hand). That's what FAST is doing, and while working on the grid, I see a lot of issues with that (besides the fact their tests take a lot of time to run locally on my machine, with lots of flakiness).

it's a tradeoff. by choosing DX over accuracy, we should be aware we'll have unexpected behaviors in our tests (which, in turn, impact DX). I sympathize with [the lack of watch mode](https://github.com/microsoft/playwright/issues/7035) which has been reported two years ago and have made [a pros & cons list of PW vs jsdom here](https://github.com/Vonage/vivid-3/issues/839).

libs/components/src/lib/tab/README.md
rinaok: @yinonov the link is broken

libs/components/src/lib/tabs/tabs.scss
rachelbt:
```suggestion
		box-shadow: 1px 0 0 0 var(#{constants.$vvd-color-neutral-300});
```

libs/components/src/lib/tabs/tabs.scss
```diff
@@ -0,0 +1,87 @@
+@use "../tabs/partials/variables" as variables;
+@use "../../../../shared/src/lib/sass/mixins/connotation/config" with (
+	$connotations: accent,
+	$shades: primary primary-text primary-increment faint soft,
+	$default: accent
+);
+@use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
+
```
rachelbt, line 8:
```suggestion
@use "../../../../../dist/libs/tokens/scss/tokens.constants" as constants;

```

rinaok (approved):

libs/components/src/lib/tab/README.md
yinonov: fixed

yinonov (commented):
tab ui snapshots are broken

rinaok (approved):

PR #701, opened 2022-09-26
feat(components): add radio group

olaf-k:


olaf-k:
WIP, component needed to work with #696

libs/components/src/lib/radio-group/radio-group.template.ts
```diff
@@ -15,7 +14,7 @@ import type { RadioGroup } from './radio-group';
  */
 export const RadioGroupTemplate: (context: ElementDefinitionContext) => ViewTemplate<RadioGroup> = (context: ElementDefinitionContext) => {
 	return html<RadioGroup>`
-		<template
+		<span
```
rachelbt, line 17:
```suggestion
		<div
```

libs/components/src/lib/radio-group/radio-group.template.ts
```diff
@@ -29,6 +28,6 @@ export const RadioGroupTemplate: (context: ElementDefinitionContext) => ViewTemp
 			<div class="positioning-region ${x => x.orientation === Orientation.horizontal ? 'horizontal' : 'vertical'}">
 				<slot ${slotted({ property: 'slottedRadioButtons', filter: elements(context.tagFor(Radio)) })}></slot>
 			</div>
-		</template>
+		</span>
```
rachelbt, line 31:
```suggestion
		</div>
```

libs/components/src/lib/radio-group/radio-group.scss
```diff
@@ -1,6 +1,5 @@
 @use '../../../../../dist/libs/tokens/scss/constants' as constants;
 
-
 .positioning-region {
 	display: flex;
 	flex-wrap: wrap;
```
rachelbt, line 5: in the code foe the horizontal & vertical I suggest changing the column/row-gap to plain `gap: 8px` we need it when the wrap is on:
image

and for horizontal - there's no need for `flex-direction: row; this is the default behaviour of flex.

libs/components/src/lib/radio-group/README.md
```diff
@@ -0,0 +1,73 @@
...
+### Label
+
+Use the `label` member to set the group's label.
+
+- Type: `string`
+- Default: `undefined`
+
+```html preview
+<vwc-radio-group label="Pick a number" name="number">
+    <vwc-radio label="1" value="1" checked></vwc-radio>
```
yinonov, line 23: first show the default unchecked
```suggestion
    <vwc-radio label="1" value="1"></vwc-radio>
```

libs/components/src/lib/radio-group/radio-group.template.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * The template for the {@link @microsoft/fast-foundation#RadioGroup} component.
+ *
+ * @param context
+ * @public
+ */
+export const RadioGroupTemplate: (context: ElementDefinitionContext) => ViewTemplate<RadioGroup> = (context: ElementDefinitionContext) => {
+	return html<RadioGroup>`
+		<div
+			role="radiogroup"
```
yinonov, line 18: the role is set on the base first element of the template but the complimenting attributes are set from within the class to the host itself

these paired aria attributes should probably be set on the same element
image

not sure this makes any sense and what assistive technology actually reads it. can u ensure this is a valid practice or not?

yinonov: same goes for radio component
image

yinonov (changes requested):

libs/components/src/lib/radio-group/radio-group.template.ts
olaf-k: I think we already talked about this, which is the reason for [the additional Axe test](https://github.com/Vonage/vivid-3/blob/3572b135e1a979937066235f692cbfaf8676388e/libs/components/src/lib/radio-group/radio-group.spec.ts#L138).

I ran a check with VoiceOver and the radio-group seemed to behave normally. (On another topic, the `readonly` state is very misleading.)

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+			expect(element.orientation).toEqual('horizontal');
+		});
+	});
+
+	describe('label', () => {
+		it('should display a label when given one', async () => {
+			element.label = 'testlabel';
+			await elementUpdated(element);
+			const label = element.shadowRoot?.querySelector('label');
+			expect(label).not.toBeNull();
```
YonatanKra, line 48: I believe this is not needed, because the next line also makes sure of it. Right?

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+			expect(label?.textContent).toBe(element.label);
+		});
+	});
+
+	describe('orientation', () => {
+		it('should update the positioning region when changing orientation', async () => {
+			const positioningRegion = element.shadowRoot?.querySelector('.positioning-region');
+			expect(positioningRegion?.classList).toContain('horizontal');
```
YonatanKra, line 56: Please move all expectations to the end.

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+			await elementUpdated(element);
+			expect(radios[1].checked && !radios[0].checked && !radios[2].checked).toBeTruthy();
+			expect(radios[0].getAttribute('tabindex')).toBe('-1');
+			expect(radios[1].getAttribute('tabindex')).toBe('0');
+			expect(radios[2].getAttribute('tabindex')).toBe('-1');
+		});
+
+		it('should receive the value of the radio that was clicked', async () => {
+			expect(element.getAttribute('value')).toBeNull();
+			expect(radios.every(r => r.checked)).toBeFalsy();
```
YonatanKra, line 83: Please set all expectation in the end of the `it`.

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+			getBaseElement(radios[1]).dispatchEvent(new MouseEvent('click'));
+			await elementUpdated(element);
+			expect(element.getAttribute('value')).toEqual('1');
+		});
+	});
+	
+	describe('click', () => {
+		it('should update when a radio is clicked', async () => {
+			expect(radios[2].getAttribute('tabindex')).toBe('-1');
```
YonatanKra, line 92: All expects in the end of the `it`

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+	describe('keyboard', () => {
+		async function keyboardCheck(radioToCheck: number, key: string) {
+			expect(radios[radioToCheck].checked).toBeFalsy();
+			expect(element.value).toBeUndefined();
+			radios[0].focus();
+			radios[0].dispatchEvent(new KeyboardEvent('keydown', { key: key, bubbles: true }));
+			await elementUpdated(element);
+			expect(radios[radioToCheck].checked).toBeTruthy();
+			expect(element.value).toBe(radioToCheck.toString());
```
YonatanKra, line 108: Mixing expectations with actions is not a good practice.
Please split them for the Action and Expectation.
You also have a Setup/Preparation phase here hiding inside the `keyboardCheck`.
So one has to read the whole function to understand what's going on instead of:
```
setupFocus(0);
hitKey(key);
expect(checkedRadioValueBeforeKeyStroke).toBeFalsy();
expect(groupValueBeforeKeyStroke).toBeUndefined();
expect(radios[radioToCheck].checked).toBeTruthy();
expect(element.value).toBe(radioToCheck.toString());
```

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+		it('should update when arrows are used', async () => {
+			await keyboardCheck(1, 'ArrowRight');
+		});
+
+		it('should update when Enter is pressed', async () => {
+			await keyboardCheck(0, 'Enter');
+		});
+
+		it('should loop', async () => {
```
YonatanKra, line 119: Should loop over what? and do what? in what condition?

libs/components/src/lib/radio-group/radio-group.spec.ts
olaf-k: In your example, one has to understand what's going on in the additional functions you use too, so I'm not convinced by the gain in efficiency?

YonatanKra: I don't need to enter the functions to understand. I know what they are doing by their names.
For instance - `setupFocus` should be `setFocusOnElementWithValue(1)` or `setFocusOnRadioIndexed(0)`.
I can just read the function's name - I don't care about its implementation.
Same goes for `hitKey` - I don't care how it works. I just know it has the effect of hitting the key `key` :)

libs/components/src/lib/radio-group/radio-group.template.ts
```diff
@@ -0,0 +1,33 @@
...
+			aria-disabled="${x => x.disabled}"
+			aria-readonly="${x => x.readOnly}"
+			@click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
+			@keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
+			@focusout="${(x, c) => x.focusOutHandler(c.event as FocusEvent)}"
```
yinonov, line 23: should [aria-orientation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-orientation) be included here as well?

it's missing in fast https://github.com/microsoft/fast/blob/2536c024febdf8a89300c257805a19a946f8039d/packages/web-components/fast-foundation/src/radio-group/radio-group.template.ts#L11-L17

intentional?

yinonov: I think we should include it in our template. it is inherited to radio-group. probably defaults to horizontal

olaf-k: yeah, it's a c/c from FAST, good catch.

yinonov (approved):

PR #704, opened 2022-09-28
docs(components): fonts as assets responsibility

yinonov:


libs/components/README.md
```diff
@@ -51,6 +51,8 @@ Only one theme is required to be loaded.
 
 - `node_modules/@vonage/vivid/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
 
+Note that font files are not loaded by the css file, and must be loaded separately. This is to allow the author to choose the most appropriate way to load the font files based on their project.
```
yinonov, line 54:
```suggestion
Note that font files are not included within the css file, and must be copied to application assets separately (within the same parsed css folder). This is to allow the author to choose the most appropriate way to load the font files based on their project.
```

libs/components/README.md
```diff
@@ -51,6 +51,8 @@ Only one theme is required to be loaded.
 
 - `node_modules/@vonage/vivid/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
 
+Note that font files are not included within the css file, and must be copied to application assets separately (within the same parsed css folder). This is to allow the author to choose the most appropriate way to load the font files based on their project.
```
YonatanKra, line 54: Don't we want to supply examples?
Not necessarily now, but can be a different issue (supply 3 examples of importing fonts - vanilla, angular, svelete).

YonatanKra (approved):
Approved - just see my suggestion in the comments. If you agree, you can create the issue and assign it to me.

PR #705, opened 2022-09-28
chore(components): refine checkbox style

yinonov:


libs/components/src/lib/checkbox/checkbox.scss
```diff
@@ -34,11 +36,19 @@ $gutter: 4px;
 	@supports (user-select: none) {
 		user-select: none;
 	}
+
+	&:not(.disabled) {
```
rachelbt, line 40: what about readonly?
```suggestion
	&:not(.disabled, .readonly) {
```

libs/components/src/lib/checkbox/checkbox.scss
```diff
@@ -25,6 +25,8 @@ $gutter: 4px;
 	display: inline-flex;
 	align-items: center;
 	gap: 12px;
```
rachelbt, line 27: I know this is not related to the PR but if radio is 8px gap (or 10px) that shouldn't it be the same here?

libs/components/src/lib/checkbox/checkbox.scss
yinonov: Readonly is still interactive

libs/components/src/lib/checkbox/checkbox.scss
yinonov: it's addressed at https://github.com/Vonage/vivid-3/pull/697/files#diff-aa6257cff9e49a629c169c30f42ecac73cfa91357575d9cf13511ac0336eff05R28

rachelbt (approved):
LGTM

PR #707, opened 2022-09-28
refactor(docs): improve ux by window resize

yinonov:
observing window resize rather then visual viewport improves user
experience

### DIFF
--- a/apps/docs/assets/scripts/viewport-handler.js
+++ b/apps/docs/assets/scripts/viewport-handler.js
@@ -3,22 +3,23 @@ const mobileWidth = 880;
 let pendingUpdate = false;
 
 function viewportHandler() {
-	const { visualViewport: { width } } = window;
+	const { innerWidth } = window;
+
 	if (pendingUpdate) return;
+
 	pendingUpdate = true;
 
 	requestAnimationFrame(() => {
 		pendingUpdate = false;
 
-		sideDrawer.modal = width < mobileWidth;
+		sideDrawer.modal = innerWidth < mobileWidth;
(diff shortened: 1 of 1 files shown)


apps/docs/assets/scripts/viewport-handler.js
```diff
@@ -3,22 +3,23 @@ const mobileWidth = 880;
...
 	if (pendingUpdate) return;
+
 	pendingUpdate = true;
 
 	requestAnimationFrame(() => {
 		pendingUpdate = false;
 
-		sideDrawer.modal = width < mobileWidth;
+		sideDrawer.modal = innerWidth < mobileWidth;
+		sideDrawer.open = innerWidth >= mobileWidth;
```
olaf-k, line 16: when zooming but staying out of mobile dimensions, this will reopen the drawer even if the user chose to close it.
maybe with `&&=` ?

yinonov: for large screen that should be a correct use of the "real-estate". should it not? resizing the viewport isn't a something a user do commonly. it's more of a solution for page load in accordance to viewport

olaf-k: I feel the site should not be more opiniated than the user :) but it's nitpicking yes, feel free to ignore.

yinonov: this concern requires keeping a state for user preference and consider it to conditionally open back the aside.
seems like a different scope

yinonov: see #710

olaf-k (approved):

PR #713, opened 2022-09-29
fix(avatar): initials 

rachelbt:


libs/components/src/lib/avatar/avatar.spec.ts
```diff
@@ -91,4 +91,14 @@ describe('vwc-avatar', () => {
 
 	});
 
+	describe('avatar name', () => {
+		it('should not show the icon if name is set', async () => {
```
yinonov, line 95: is there a test where the same _expect_ returns an instance?

libs/components/src/lib/avatar/avatar.template.ts
```diff
@@ -24,11 +47,10 @@ const getClasses = ({appearance, connotation, shape, density}: Avatar) => classN
...
 	definition: FoundationElementDefinition
-) => ViewTemplate<Avatar> = () => html` 
+) => ViewTemplate<Avatar> = () => html`
 	<span class="${getClasses}">
 		<slot>
-			<span class="icon">
-				<vwc-icon type="${(x) => x.icon? `${x.icon}` : 'user-line'}"></vwc-icon>
-			</span>
+			${when(x => x.name, renderInitials())}
+			${when( x => !x.name, renderIcon())}
```
yinonov, line 54: then why render if no icon is set? or if node is assigned to slot?

libs/components/src/lib/avatar/partials/_mixins.scss
```diff
@@ -1,11 +1,11 @@
 @use 'variables' as variables;
 
-
 @mixin get-density-properties($block-size, $font-face) {
 	#{variables.$avatar-density}: $block-size;
 
-	// the font is for acronym:
-	// @include typography.typography-cat-shorthand($font-face);
+	.initials {
+		font: var(#{$font-face});
```
yinonov, line 7: no matter the avatar size, always the same font face?

yinonov (changes requested):

libs/components/src/lib/avatar/README.md
```diff
@@ -82,6 +87,10 @@ Use the `density` attribute/property to set the avatar's to one of the predefine
 <vwc-avatar density='condensed' shape="pill"></vwc-avatar>
 <vwc-avatar density='normal' shape="pill"></vwc-avatar>
 <vwc-avatar density='extended' shape="pill"></vwc-avatar>
+<hr>
```
yinonov, line 90: indeterministic. can result in flaky ui test

yinonov (changes requested):

libs/components/src/lib/avatar/avatar.spec.ts
rachelbt: there was a problem with the test. @YonatanKra [took it on him](https://github.com/Vonage/vivid-3/issues/721)

libs/components/src/lib/avatar/avatar.template.ts
rachelbt: because if there's no icon set - then the avatar is presenting its default view: icon of 'user-line'

libs/components/src/lib/avatar/README.md
rachelbt: then whats the best way to do?

libs/components/src/lib/avatar/partials/_mixins.scss
rachelbt: its a variable - it gets a font size in the scss:
```
&.density-condensed {
		@include mixins.get-density-properties(32px, #{constants.$vvd-font-base-condensed-bold});
	}
```

libs/components/src/lib/avatar/avatar.template.ts
yinonov: true. but anyway, if an image element is assigned, we don't need the icon. the slotted directive can assist verifying if an image was assigned and condition the icon

libs/components/src/lib/avatar/partials/_mixins.scss
yinonov: 🤦‍♂️

libs/components/src/lib/avatar/README.md
yinonov: either vwc-divider or nothing. whatever improves reading of code (nothing)

YonatanKra (approved):

libs/components/src/lib/avatar/README.md
```diff
@@ -82,13 +89,16 @@ Use the `density` attribute/property to set the avatar's to one of the predefine
 <vwc-avatar density='condensed' shape="pill"></vwc-avatar>
 <vwc-avatar density='normal' shape="pill"></vwc-avatar>
 <vwc-avatar density='extended' shape="pill"></vwc-avatar>
+<vwc-avatar density='condensed' shape="pill" name="john doe"></vwc-avatar>
+<vwc-avatar density='normal' shape="pill" name="john doe"></vwc-avatar>
+<vwc-avatar density='extended' shape="pill" name="john doe"></vwc-avatar>
```
yinonov, line 94: let's focus on each member and it's specific responsibility?
```suggestion
```

yinonov (changes requested):
last comment before approval

libs/components/src/lib/avatar/README.md
rachelbt: I think its important to see the sizes of the font as well of the icon changes due to "size/density" member

yinonov: so maybe u should say something about it. I, myself, didn't realize

rachelbt: like what?
Isn't the output is enough?
image

yinonov (approved):

PR #714, opened 2022-09-29
fix(states): readonly with hover

rachelbt:


libs/shared/src/lib/sass/mixins/appearance/_mixins.scss
```diff
@@ -20,9 +20,9 @@ $error: '.error';
 
 $selectors: (
   idle: '&',
-  hover: '&:where(#{$hover}):where(:not(#{$disabled}))',
+  hover: '&:where(#{$hover}):where(:not(#{$disabled}, #{$readonly}))',
```
yinonov, line 23: readonly should interact to user gestures

yinonov (changes requested):
why would we eliminate any response to user interaction when readonly? it is still a reactive state only it should be altered.
it is a valid "landmark" (can be active)

rachelbt:
not needed as [commented](https://github.com/Vonage/vivid-3/pull/714#discussion_r983915546)

PR #723, opened 2022-10-03
feat(components): badge support for icon-only

### LINKED ISSUE #711: [badge] support icon-only
I tried using the badge without text, but what we have right now creates an elliptic shape.

Image

_Originally posted by @AyalaBu in https://github.com/Vonage/vivid-3/issues/702#issuecomment-1262043741_

yinonov:
Closes #711


rachelbt (commented):
I think that we need to enlarge the icon size in size - expended.
Note that when expended or condense is set the padding are being overriden (doesn't effect the look because of the aspect-ration)

yinonov:
> I think that we need to enlarge the icon size in size - expended. Note that when expended or condense is set the padding are being overriden (doesn't effect the look because of the aspect-ration)

will follow design (in the works) https://www.figma.com/file/r8OYLPmRmDqfLKt6KHdJDB/branch/eiSBjHTMsNrioTbA1Kohru/Vivid-UI-Kit---2.0?node-id=16381%3A76273

PR #725, opened 2022-10-03
chore(components): hover and active states precede selected

yinonov:


rachelbt (commented):
Makes sense.
But what about checked? shouldn't it get it as well?
Were did you accoutred the issue - I see that in menu item the selected doesn't have hover running over its colors.

yinonov:
> Makes sense.
> 
> But what about checked? shouldn't it get it as well?
> 
> Were did you accoutred the issue - I see that in menu item the selected doesn't have hover running over its colors.

Can u elaborate please?

rachelbt:
> > Makes sense.
> > But what about checked? shouldn't it get it as well?
> > Were did you accoutred the issue - I see that in menu item the selected doesn't have hover running over its colors.
> 
> Can u elaborate please?

if selected has override upon hover + active, shouldn't checked be too? 
In menu item has selected class it hoesn't have hover state:
https://vivid.deno.dev/components/nav-item
image

rachelbt (approved):
LGTM 
(though, I personally don't like hover on checked or selected items, but U get the need)

PR #726, opened 2022-10-04
fix(size): change density to size

rachelbt:
First stem in moving to using sizing tokens


yinonov:
add this to the list of removals (enlarged number field)

image

change/@vonage-vivid-40f0613f-f6a4-4d06-b516-92c0b52da217.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "change densety to size and extended to expanded",
```
yinonov, line 3:
```suggestion
  "comment": "change density to size and extended to expanded",
```

yinonov (changes requested):

libs/components/src/lib/enums.ts
```diff
@@ -30,19 +30,34 @@ export enum Appearance {
 	Subtle = 'subtle',
 	Ghost = 'ghost',
 }
-
+//need to be removed after remove in text-field
 export enum Density {
```
yinonov, line 34: are we still using it?

libs/components/src/lib/icon/README.md
```diff
@@ -50,13 +50,13 @@ While `type` fetches an icon from the Vivid predefined set of icons, `svg` prope
 
 Use the `size` attribute to apply a predefined size to the icon.
 
-- Type: `'small'` | `'medium'` | `'large'`
-- Default: `'medium'`
+- Type: `'condensed'` | `'normal'` | `'expanded'`
```
yinonov, line 53: Icon should be the same as progress ring. next PR?

yinonov (commented):
few small comments

libs/components/src/lib/icon/README.md
rachelbt: Not sure it does. either way - that sounds like a different PR

libs/components/src/lib/enums.ts
rachelbt: no. you re right - we removed the density there. 
I'll remove it

rachelbt: removed

yinonov (approved):

PR #727, opened 2022-10-04
chore(workspace): update Docker image with Chromium for M1 Macs

olaf-k:
### Problem
The Docker image we use for visual regression tests is (like most) an amd64 one. On M1-based machines (arm64), Docker Desktop will have to run it using the QEMU emulator, which seems to make Chrome crash.

### PR description
The test script had been updated to use another image available in both amd64 and arm64 flavors (automatically selected by `docker run`). As there is no linux/arm64 version of Chrome available, the image also replaces the `Desktop Chrome` Playwright project with a `Desktop Chromium` one. Updated snapshots are also provided (identical in content but with a new filename).


olaf-k:
The PR will be complete once I can push the image to the proper Docker repo, but should work as-is already.

scripts/visual-tests/run.tests.in.container.sh
```diff
@@ -6,5 +6,6 @@ case $1 in
   ./scripts/visual-tests/run.tests.sh ;;
 *)
   echo "Docker test run"
-  docker run -it --name vivid-visual-tests --env task=$1 --rm -v $PWD:/usr/src/app drizzt99/vonage:1.2.0
+  # update with proper image name when moved
+  docker run -it --name vivid-visual-tests --env task=$1 --rm -v $PWD:/usr/src/app olafk/vonage:1.2.0
```
rachelbt, line 10: not that `olafk` is not a good name - but does this mean that the path is connected to your account?
If so - I'm not sure this is so good.

rachelbt (commented):
the action-group png change is strange a bit

olaf-k:
> the action-group png change is strange a bit

this is strange indeed.
if I checkout main and run the e2e tests against the current screenshot, everything is ok (as it should).
but if I delete the Chrome-Stable image and regenerate it, the new one actually looks like the picture from the PR...

docs/ui-test/readme.md
```diff
@@ -74,8 +74,8 @@ The docker image comes with the browsers and playwright ready for action.  If, f
 2. Update the tag and push to the repository.
 
 ```
-docker tag vivid-visual-tests-img drizzt99/vonage:1.x.x
-docker push drizzt99/vonage:1.x.x  
+docker tag vivid-visual-tests-img vivid3/vonage:latest
```
YonatanKra, line 77: Do we need to get the credentials for the `vivid3` repo?

libs/components/playwright.config.ts
```diff
@@ -5,10 +5,9 @@ const config: PlaywrightTestConfig = {
 	outputDir: '../../test-results',
 	projects: [
 		{
-			name: 'Chrome Stable',
+			name: 'Desktop Chromium',
```
YonatanKra, line 8: Are we now using chromium for all machines? Not only on M1?

YonatanKra (approved):
LGTM

PR #728, opened 2022-10-04
chore(workspace): update _deploy-deno.yml

olaf-k:
Can we make the Deno deploy a manual action too? Could be useful to sometimes skip the CI and just deploy to get a shareable URL.


olaf-k:
Speaking of which I forgot to [skip ci] ughhh.

yinonov:
> Can we make the Deno deploy a manual action too? Could be useful to sometimes skip the CI and just deploy to get a shareable URL.

in what scenarios? maybe it's a deeper issue where we want to skip all checks on draft PRs? is that the case?

olaf-k:
> > Can we make the Deno deploy a manual action too? Could be useful to sometimes skip the CI and just deploy to get a shareable URL.
> 
> in what scenarios? maybe it's a deeper issue where we want to skip all checks on draft PRs? is that the case?

we can force skip all workflow actions using `[skip ci]`, but the Deno deployment is exactly a check per se.
the only use case I had was being able to share a reproducer URL, so admittedly not a very frequent occurrence indeed.

yinonov (approved):

PR #730, opened 2022-10-06
fix(components): remove role toolbar from header part

yinonov:
adding a role of toolbar to assigned nodes is an author concern


rachelbt (approved):
LGTM

PR #734, opened 2022-10-06
fix(popup): drop-shadow filter inconsistent view in Safari

rachelbt:


rachelbt:
the issue doesn't happen consistently + do not happen outside the docs:
https://codepen.io/rachelbt/pen/GRdXZbb
also not when trying to mimic the view inside an iframe:
https://codepen.io/rachelbt/pen/ZEoNPNZ
Closing it for now.

PR #737, opened 2022-10-10
fix(components): min-width for radio & checkbox

olaf-k:
(branch based on `radio`)


rachelbt (commented):
why do we need the min-width for?
If its because of it getting shrunk due flex than we can add to it - `flex-shrink: 0;`

PR #745, opened 2022-10-13
chore(form-elements): form elements error message pattern

YonatanKra:


libs/components/src/shared/patterns/form-elements.ts
```diff
@@ -59,3 +62,16 @@ export function formElements<T extends { new (...args: any[]): Record<string, an
...
+
+export const errorMessageTemplateFactory: (context: ElementDefinitionContext) =>
+() => ViewTemplate<FormElement> =
+	(context) => {
+		const iconTag = context.tagFor(Icon);
+		return () => {
+			return html<FormElement>`
+				${when(x => x.errorValidationMessage, html<TextField>`
+					<${iconTag} class="error-message-icon" type="info-negative"></${iconTag}>
+					<span class="error-message">${x => x.errorValidationMessage}</span>`)}
```
rachelbt, line 74:
```suggestion
<div class="base">
					<${iconTag} class="icon" type="info-negative"></${iconTag}>
					<span class="message">${x => x.errorValidationMessage}</span>`)}
</div>
```
and I would add CSS:
```
.base {display: flex; row-gap: 8px; align-items: flex-start;}
.icon {flex-shrink: 0;}
```

rachelbt (commented):
the scss of the error message can also be in the pattern so we can avoid duplicating the code?

olaf-k (changes requested):

libs/components/src/shared/patterns/form-elements.ts
```diff
@@ -59,3 +62,16 @@ export function formElements<T extends { new (...args: any[]): Record<string, an
...
 	return Decorated;
 }
+
+export const errorMessageTemplateFactory: (context: ElementDefinitionContext) =>
+() => ViewTemplate<FormElement> =
+	(context) => {
+		const iconTag = context.tagFor(Icon);
+		return () => {
+			return html<FormElement>`
+				${when(x => x.errorValidationMessage, html<TextField>`
```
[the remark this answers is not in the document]
olaf-k: > In the same spirit of reuse, I'd remove the condition check `${when(x => x.errorValidationMessage` and leave it to the component to call the function (why a factory?) or not (thus making context a parameter).

I can live with the when in the helper until it's needed outside.

yinonov:
> the scss of the error message can also be in the pattern so we can avoid duplicating the code? Or is the pattern css-less?

it's css less, it supports a template & class members. for css we have scss mixins

libs/components/src/shared/patterns/error-message.scss
```diff
@@ -0,0 +1,15 @@
+@use '../../../../../dist/libs/tokens/scss/constants' as constants;
```
yinonov, line 1: hmm I'm not sure if we want this in patterns. we have a sass folder in shared where we maintain mixins
https://github.com/Vonage/vivid-3/blob/d2920294dad9dab5b6c053d2dd7bdf7b54c55f4a/libs/shared/src/lib/sass

yinonov (changes requested):

libs/components/src/shared/patterns/error-message.scss
YonatanKra: This separates the design from the only place that's going to use it. It is not shared with other things. Why should it be outside the components/shared folder?
Maybe give every pattern its own folder?

olaf-k (approved):

yinonov (approved):

PR #748, opened 2022-10-17
fix(shared): small typo

olaf-k:


rachelbt (approved):
😲

yinonov (approved):
failing to provide a valid value resulted in the same output. nice catch!

PR #757, opened 2022-10-23
feat(form-elements): helper text and success message patterns

YonatanKra:
Moved helper text to a pattern and added the success message using a pattern.


libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -155,8 +150,9 @@ export const NumberFieldTemplate: (
       ${() => focusTemplate}
       ${() => numberControlButtons(context)}
     </div>
-	  ${when(x => !x.errorValidationMessage && x.helperText?.length, renderHelperText())}
-	  ${getErrorMessageTemplate(context)}
+	  ${when(x => !x.successText && !x.errorValidationMessage && x.helperText?.length, getFeedbackTemplate('helper', context))}
+	  ${when(x => !x.successText && x.errorValidationMessage, getFeedbackTemplate('error', context))}
+	  ${when(x => x.successText, getFeedbackTemplate('success', context))}
```
olaf-k, line 155: this is really a personal choice but reversing the 3 branches makes more sense to me, do what you want with that ;)

yinonov (commented):
noticed that unlike error message, the success isn't compelled to check validity. it immediately shows success indication. that aligns with a valid state and not _user-valid_ state

PR #758, opened 2022-10-25
chore(workspace): add export to components.ts

olaf-k:


yinonov (commented):
I don't recall, are we currently only export from package.json?

olaf-k:
> I don't recall, are we currently only export from package.json?

yes, the new component is properly added to `package.json` by the generator, but needs to be added manually to `components.ts`. not a big deal, but it's a small change.

yinonov (approved):

yinonov:
I suggest @YonatanKra review as well

yinonov:
> I suggest @YonatanKra review as well

@YonatanKra all good here?

olaf-k:
ping @YonatanKra when you 5min can you have a look?

YonatanKra (dismissed):
LGTM. Consider the suggestions.

YonatanKra (approved):

PR #760, opened 2022-10-26
feat(option): create the option component

rinaok:


YonatanKra (approved):
LGTM. Maybe could use another eye on the scss, but aside from that, looks good.

libs/components/src/lib/listbox-option/listbox-option.scss
```diff
@@ -0,0 +1,72 @@
...
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	color: var(#{appearance.get-appearance-token(text)});
+	font: var(#{constants.$vvd-typography-base});
+	gap: $gap;
+	hyphens: auto;
+	inline-size: 100%;
+	min-block-size: $min-block-size;
+	padding-block: 10px;
+	padding-inline: $gap;
+	text-decoration: none;
```
rachelbt, line 41:
```suggestion
```
no need for that - its not a link inside the option.

libs/components/src/lib/listbox-option/listbox-option.scss
```diff
@@ -0,0 +1,72 @@
...
+
+
+.base {
+	$min-block-size: 40px;
+	$gap: 8px;
+
+	@include connotation.connotation;
+	@include appearance.appearance;
+
+	position: relative;
```
rachelbt, line 26: why do we need this for? is there a `position: absolute` element that will be attached to it?

libs/components/package.json
```diff
@@ -40,6 +40,7 @@
     "./nav-disclosure": "./nav-disclosure",
     "./nav-item": "./nav-item",
     "./note": "./note",
+    "./listbox-option": "./listbox-option",
```
yinonov, line 43: https://github.com/microsoft/fast/blob/87b817259c504fbcae23bf12a6039fc646b2d7a6/packages/web-components/fast-foundation/src/listbox-option/README.md?plain=1#L9

this only affects the class itself and not necessarily the folder structure.
we either export it as option or even rename the path and filenames to plainly _**option**_
```suggestion
    "./option": "./listbox-option",
```

yinonov (changes requested):

libs/components/src/lib/listbox-option/README.md
```diff
@@ -0,0 +1,54 @@
+# Option
```
yinonov, line 1: missing `checked`, but as discussed, we should add hint on pitfalls

as in fast

https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/listbox-option/README.md#fields

yinonov: are there any other members in the fast table that should be included?

yinonov (changes requested):

libs/components/src/lib/listbox-option/listbox-option.scss
yinonov: try the css overview and coverage check
image

libs/components/src/lib/listbox-option/listbox-option.scss
rinaok: @rachelbt The option component will be wrapped with a listbox component. Like this:

```
<vwc-listbox>
	<vwc-option><vwc-option>
	<vwc-option><vwc-option>
	<vwc-option><vwc-option>
</vwc-listbox>

```

libs/components/src/lib/listbox-option/README.md
rinaok: I think it's better not to include it in the docs right now, and only add it when the listbox will be ready./
Then we could add an example with a listbox and write this comment:

`The checked state is used when the parent listbox is in multiple selection mode. To avoid accessibility conflicts, the checked state should not be present in single selection mode.`

rinaok: @yinonov I don't see other members that should be added except of checked

libs/components/src/lib/listbox-option/listbox-option.scss
yinonov: remove it. it doesn't seem to belong. but check visually
```suggestion
```

libs/components/src/lib/listbox-option/README.md
yinonov: sure, just open an issue to have that in our backlog to attend later

image

rinaok: done

https://github.com/Vonage/vivid-3/issues/771

yinonov (approved):

PR #763, opened 2022-10-27
feat(badge): omit badge sizes

rachelbt:


yinonov (approved):
now that sizes are gone. shape is up next 😁

rachelbt:
> now that sizes are gone. shape is up next 😁

what?!

PR #765, opened 2022-10-27
feat(popup): add light dismiss support

rinaok:
- The light dismiss feature will be used in Popup as well in Select and Menu components.

- Light dismiss behaviors: most popups "light dismiss", in that the following will dismiss the popup: pressing the ESC key, invoking elsewhere in the document, invoking another popup (unless it is a child of the currently opened popup).	
	Read more: https://open-ui.org/components/popup.research

- Should be researched if the light dismiss should work also on resize / scroll.

- Should be checked: https://github.com/oddbird/popup-polyfill


yinonov (changes requested):
due to the future support of popover, we might wanna remove any evidence of popup from the API collection. so remove its page. that way we only align with the naming popover introduces to browsers and under the hood we would eventually refactor any floating-UI / popup-polyfill to the native popover. open an issue to exclude popup from our site?

libs/components/src/lib/popup/popup.template.ts
```diff
@@ -24,14 +24,14 @@ export const popupTemplate: (
 	definition: FoundationElementDefinition
 ) => ViewTemplate<Popup> = () => html`
   <vwc-elevation>
-    <!--the popup-wrapper is needed for alternating the inside of the popup nd not its shadow-->
+    <!--the popup-wrapper is needed for alternating the inside of the popup and not it's shadow-->
```
yinonov, line 27: possessive word
```suggestion
    <!--the popup-wrapper is needed for alternating the inside of the popup and not its shadow-->
```

rinaok:
@yinonov There is a new version 0.0.3 for the package. They changed it's name to be `popover-polyfill`.
Should we also change popup's name to popover?

yinonov:
> Member

yes. everyone else do https://carbondesignsystem.com/components/popover/usage/

olaf-k:
just a reminder to [make sure the tooltip benefits from this](https://github.com/Vonage/vivid-3/pull/829#discussion_r1031351356)

rinaok:
> > Member
> 
> yes. everyone else do [carbondesignsystem.com/components/popover/usage](https://carbondesignsystem.com/components/popover/usage/)

@yinonov I see that chrome still calls it pop-up, and it had a popover attribute https://developer.chrome.com/docs/web-platform/pop-up-api/

rinaok:
@yinonov The popover-polyfill still doesn't support in all light-dismiss behaviours. It doesb't suppoer ESC for example..

libs/components/src/lib/popup/README.md
```diff
@@ -69,6 +69,26 @@ Add the `dismissible` attribute to add a close button to the popup.
 </script>
 ```
 
+### Popup
+
+Use the `popup` attribute to add Light dismiss behaviors.
```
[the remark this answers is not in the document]
rinaok: @yinonov I see that chrome calls the element pop-up, and it has a popover attribute [developer.chrome.com/docs/web-platform/pop-up-api/popover-attribute](https://developer.chrome.com/docs/web-platform/pop-up-api/popover-attribute/)

yinonov:
> @yinonov The popover-polyfill still doesn't support in all light-dismiss behaviours. It doesb't suppoer ESC for example..

it don't matter. it will. we don't support any of the light dismiss methods

rinaok:
> > @yinonov The popover-polyfill still doesn't support in all light-dismiss behaviours. It doesb't suppoer ESC for example..
> 
> it don't matter. it will. we don't support any of the light dismiss methods

Ok, so I think the PR is good now, I only have this linting problem

libs/components/src/lib/popup/README.md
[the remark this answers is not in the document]
yinonov: ok. doesn't make sense to do any changes at this point to our interface as that boat rocks

libs/components/src/lib/popup/README.md
```diff
@@ -69,7 +69,27 @@ Add the `dismissible` attribute to add a close button to the popup.
 </script>
 ```
 
-### Arrow
+### Popover
+
+Use the `popover` attribute to add Light dismiss behaviors. 
+- When `'auto'` the following will dismiss the popup: pressing the ESC key, keyboard-navigation, invoking elsewhere in the document, invoking another popup, etc.
```
yinonov, line 75:
```suggestion
- When `'auto'` the following will dismiss the popup: ~~pressing the ESC key~~, keyboard-navigation, invoking elsewhere in the document, invoking another popup, etc.
```

libs/components/src/lib/menu/menu.template.ts
```diff
@@ -35,6 +35,7 @@ export const MenuTemplate: (
 				@open="${x => x.popupOpenChanged()}"
 				@close="${x => x.popupOpenChanged()}"
 				${ref('_popup')}
+				popover
```
yinonov, line 38: test this

libs/components/src/lib/popup/README.md
```diff
@@ -69,7 +69,27 @@ Add the `dismissible` attribute to add a close button to the popup.
...
+- When `'manual'`, popup cannot be light dismissed (it can only be dismissed by an explicit trigger element or by JavaScript), and they don't automatically dismiss previously-shown popups.
+
+- Type: `'auto'` | `'manual'`
```
yinonov, line 78: decide whether we want to support (and test) `manual` choice, which means no-light-dismiss, OR
add `lightDismiss` property to popup custom element to support by default (or not default). keep in mind popup is now an implementation detail for underlying usage and will not reflect (formally) any of it's feature. it's like a constant alpha tagged custom element. as github readme files are still exposed in the repo, we should state in the top of popup readme that it is being used internally to enable components' popup ability

yinonov (changes requested):

rinaok:
Waiting for https://github.com/oddbird/popover-polyfill/pull/49

libs/components/src/lib/popup/README.md
rinaok: @yinonov  added manual test

libs/components/src/lib/popup/popup.ts
```diff
@@ -95,20 +93,24 @@ export class Popup extends FoundationElement {
...
+		super.connectedCallback();
+	}
+
 	override disconnectedCallback(): void {
 		super.disconnectedCallback();
-		this.anchorEl?.removeEventListener('keydown', this.#handleKeydown);
 		this.#cleanup?.();
 	}
 
 	override attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
```
yinonov, line 105: use `anchorChanged` & `openChanged` as in fast element convention

libs/components/src/lib/popup/popup.ts
```diff
@@ -95,20 +93,24 @@ export class Popup extends FoundationElement {
 	 */
 	@attr anchor!: string | HTMLElement;
 
+	override connectedCallback(): void {
```
yinonov, line 96: redundant

libs/components/src/lib/popup/definition.ts
```diff
@@ -5,7 +5,7 @@ import { elevationRegistries } from '../elevation/definition';
 import { Popup } from './popup';
 import styles from './popup.scss';
 import { popupTemplate as template } from './popup.template';
-
+import '@oddbird/popover-polyfill';
```
yinonov, line 8: this still bugs me as it executes side effects from here, the component class file.

should we add the following code in a static method instead?
https://github.com/oddbird/popover-polyfill/blob/fe1958d5b7677a716a8c6b5172fceec83adba56b/src/index.ts#L1-L11

yinonov (changes requested):

yinonov (changes requested):
do action-group snapshot belong in this scope?

rinaok:
> do action-group snapshot belong in this scope?

It's not, here I created a separate PR for that: https://github.com/Vonage/vivid-3/pull/924

libs/components/src/lib/popup/definition.ts
rinaok: But if there will be a change, our tests will fail right?

yinonov: no, it shouldn't. actually I'm doubting the benefit. @olaf-k @YonatanKra WDYT?

olaf-k: what issue do you foresee running their code implicitly, as opposed to doing it ourselves? should we have more control over when `apply` is called? do we plan another implementation of `isSupported`? if the answer is no to both questions, I'm ok with the side-effect import.

yinonov: Ok then

yinonov (approved):

PR #768, opened 2022-10-30
refactor(tokens): locking down baseline size

yinonov:
restrict allowed values for baseline size

- [x] add @property platform to collect all public references. output property at rules based on the value and meta data


libs/tokens/src/lib/transforms/font-size.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+	isSource: false
+};
+
+const token = {
+	value: "16 * 4.125",
+	type: 'fontSizes',
+};
+
+describe('basic', () => {
+	it('should match if is of type fontSizes', () => {
```
YonatanKra, line 20:
```suggestion
	it('should match if token is of type fontSizes', () => {
```

libs/tokens/src/lib/transforms/font-size.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+		expect(matcher({
+			...defaultToken,
+			...token,
```
YonatanKra, line 23: I actually think it is much clearer if you just add `type: "fontSizes"` instead of adding the token here.

libs/tokens/src/lib/transforms/font-size.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+		})).toBeTruthy();
+	});
+
+	it('should not match if not of type fontSizes', () => {
```
YonatanKra, line 27:
```suggestion
	it('should not match if token is not of type fontSizes', () => {
```

libs/tokens/src/lib/transforms/font-size.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+		expect(matcher({
+			...defaultToken,
+			type: 'other'
+		})).toBeFalsy();
+	});
+
+	it('should transform to css calc', () => {
+		expect(transformer({
+			...defaultToken,
+			...token
```
YonatanKra, line 37: Can hard-code the token here for readability.

YonatanKra (changes requested):
Looks good. Much simpler. Added some comments regarding the tests.

libs/tokens/src/lib/transforms/public-css-references.ts
```diff
@@ -0,0 +1,13 @@
+import type { Named, Transform } from "style-dictionary";
```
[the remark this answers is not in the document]
yinonov: Yes. Ideas for a better name?

yinonov:
should document public references. could these be autogenerated by tokens?

libs/tokens/src/lib/transforms/public-css-references.ts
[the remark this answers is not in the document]
YonatanKra: Maybe be more clear these are css variables. For us it is clear but I had to get into the file to learn that. We want to avoid devs needing to get into a file to know what it's supposed to do/hold, no?

libs/tokens/src/lib/transforms/font-weight.spec.ts
```diff
@@ -11,31 +11,27 @@ const defualtToken = {
...
+	it('should match if token is of type fontWeights', () => {
 		expect(matcher({
 			...defualtToken,
-			...token,
+			type: 'fontWeights'
 		})).toBeTruthy();
 	});
 
-	it('should not match if not of type fontWeights', () => {
+	it('should not match if token is not of type fontWeights', () => {
```
YonatanKra, line 34:
```suggestion
	it('should return false if token is not of type fontWeights', () => {
```

YonatanKra (dismissed):
LGTM. Just consider my final comment on the `it` description. Try not to use `not` :)

YonatanKra (approved):

Issue #718, opened 2022-10-02
[text-field]: remove density + add use-case with token value change


rachelbt:
after theses are done:
[[size]: base token](https://github.com/Vonage/vivid-3/issues/716)
[[size]: base token and sizes documentation](https://github.com/Vonage/vivid-3/issues/717)

Issue #641, opened 2022-09-01
[density]: naming convention and sizes


rachelbt:
02/10/2022 tasks related to this issue:

- [x] https://github.com/Vonage/vivid-3/issues/716
- [ ] https://github.com/Vonage/vivid-3/issues/717
- [x] https://github.com/Vonage/vivid-3/issues/718
- [x] https://github.com/Vonage/vivid-3/issues/743
- [ ] https://github.com/Vonage/vivid-3/issues/719
- [ ] https://github.com/Vonage/vivid-3/issues/720
- ~[ ] https://github.com/Vonage/vivid-3/issues/731~

- [ ] refer all size styles in components instead of the hardcoded values
- [ ] consider this: https://github.com/Vonage/vivid-3/issues/822

---------------------------------------------------------------------------------------------------
Currently we have 2 methods of density - names and numbers (in progress-ring).
We want to unify all as suggested here:

```
const sizeMap = {
	ultraCondensed: '20', // required for progress ring,
	superCondensed: '24',
	extraCondensed: '28',
	condensed: '32',
	semiCondensed: '36',
	normal: '40',
	semiExpanded: '44',
	expanded: '48',
	extraExpanded: '52',
	superExpanded: '56',
        ultraExpanded: '60',
};
```

this change will affect all components as we will change the `extended` density to `expanded`, and progress ring will not use numbers but keywords in density.

yinonov:
relates to #337 #571

yinonov:
notice that icon still has a `size` member

<img width="591" alt="image" src="https://user-images.githubusercontent.com/15140652/188305919-39ed7406-0ea7-4bc5-9b62-f64bdaceeed0.png">

yinonov:
layout members are still defined by t-shirt sizes

<img width="974" alt="image" src="https://user-images.githubusercontent.com/15140652/188800315-78a577e4-daa9-4b97-8307-caeb9d28f33b.png">

rachelbt:
Decision summery:
1. we will use only one property - size for all
2. Text-field that has size as native attribute:
2.1 will not support vivid sizes property
2.2 will have an option of changing the base: 40px token to allow dense q extended text-field.we
2.3 Add documentation for a use-case for text-field using token customize.
3. we will still have 2 methods of size units: 
4.1 naming (normal, condensed etc..) - for components that has 3 sizes (of 8px jumps).
[...]

PR #773, opened 2022-10-31
chore(components): pair with scss size tokens

yinonov:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -81,12 +82,8 @@ $low-ink-color: --_low-ink-color;
...
 
 	/* Size */
-	.base:not(.density-extended)>& {
-		block-size: 40px;
-	}
-
-	.base.density-extended>& {
-		block-size: 48px;
+	.base > & {
+		block-size: #{size.$vvd-size-normal};
```
rachelbt, line 86:
```suggestion
	.base:not(.density-extended)>& {
		block-size: #{size.$vvd-size-normal};
	}

	.base.density-extended>& {
		block-size: #{size.$vvd-size-expanded};
```
we have [this ](https://github.com/Vonage/vivid-3/issues/718) and [this](https://github.com/Vonage/vivid-3/issues/641#:~:text=change%C2%A0%23718-,%5Btext%2Dnumber%5D%3A%20remove%20density%20%2B%20add%20use%2Dcase%20with%20token%20value%20change%C2%A0%23743,-%5Bsize%5D%3A%20change%20all) for omitting the size in text-area + number-field.

rachelbt (approved):
LGTM

PR #775, opened 2022-11-01
chore(components): apply selected & hover

yinonov:


rachelbt (approved):
LGTM

rinaok:
> LGTM @rinaok we should add this to option

@rachelbt It is already inside option :)

PR #779, opened 2022-11-02
fix(text-field): omit density

rachelbt:


libs/components/src/lib/number-field/number-field.template.ts
```diff
@@ -70,19 +64,19 @@ function numberControlButtons(context: ElementDefinitionContext) {
...
+					            density="condensed"
 					  					tabindex="${getTabIndex}"
 					            @click="${x => adjustValueByStep(x, SUBTRACT)}"></${buttonTag}>
 				<${dividerTag} class="divider" orientation="vertical"></${dividerTag}>
 				<${buttonTag} id="add" icon="plus-line"
 					  					aria-controls="control"
-					            shape="${ setControlButtonShape }" 
-					            density="${ setControlButtonDensity }"
+					            shape="${ setControlButtonShape }"
+					            density="condensed"
```
YonatanKra, line 79: Why do we need the `density` property here at all? Can't it be removed?

YonatanKra (approved):

PR #782, opened 2022-11-02
fix(components): harden density customization

### LINKED ISSUE #780: size density issues in changing to 1 or -1
- [ ] number-field composition
- [x] text-field composition
- [x] buttons (icon inside needs to be half of button-size)
- [x] banner (~~icons~~ buttons inside it)
- [x] checkbox
- [x] radio-button
- [x] dialog icon
- [x] fab
- [ ] all _tile_ nested elements

a point to think of:
Should typography be aligned with base-size-density and change accordingly?

yinonov:
resolve issues with custom density

Closes #780

- [ ] apply relative padding
- [ ] apply relative gap


rachelbt (commented):
not so sure radio + checkbox need to resize :)
image
image


image

I think Ayala should check the components in density 1 / -1 too

apps/docs/assets/styles/iframe.css
```diff
@@ -1,3 +1,12 @@
+
+/* delete this css block when this #782 is resolved */
+:root {
+	--vvd-size-density: -1;
```
rachelbt, line 4: why -1? it makes all 36px....

rachelbt (approved):
I guess will have more iteration in the density subject, but for now the changes here look OK.

PR #789, opened 2022-11-04
fix(docs): samples height

olaf-k:
This is an attempt at computing the samples height a little bit more accurately.
See the [Card Footer slot example](https://vivid.deno.dev/components/card#footer) for example.


yinonov (commented):
Nice! Works?

olaf-k:
> Nice! Works?

I quickly tested the most visible examples and it seems ok (some of them still have a scrollbar but appear completely on my Windows Chrome though).
Not urgent, so if someone finds a problem, shout at me!

rachelbt (commented):
it fixes the height is Safari 🥳

yinonov (approved):

PR #766, opened 2022-10-30
fix(docs): colors suggestion

rachelbt:

PR #790, opened 2022-11-04
feat(docs): different color for code blocks

olaf-k:
Maybe the font differences are too subtle, maybe I'm too old. A (slight) color shift may help distinguish code parts.
WDYT? (Also, not sure it's the proper place to do it?)


libs/styles/src/lib/core/typography.scss
```diff
@@ -26,6 +26,7 @@ $selectors: (
 
 	#{map.get($selectors, 'code')} {
 		font: var(#{constants.$vvd-typography-base-code});
+		color: var(--vvd-color-neutral-600);
```
yinonov, line 29: Colors to typefaces were not yet defined from a design perspective and this style file is meant for consumers use.
To style our docs use the docs app `main.css`

yinonov (changes requested):

apps/docs/assets/styles/main.css
```diff
@@ -155,6 +155,14 @@ article > .cbd-container {
   max-inline-size: calc(100vw - 32px);
 }
 
+article pre,
+article var,
+article code,
+article kbd,
+article samp {
+  color: var(--vvd-color-neutral-600);
```
yinonov, line 163: See highlight js region at the bottom of the file

yinonov (changes requested):

apps/docs/assets/styles/main.css
olaf-k: not sure what you mean? move it there or?

yinonov: what's the difference of when it's wrapped by anchor?

yinonov: even though these tags are a semantic family, they might have different  types of implementations. if we want to adopt a finite state for style, side effects should be avoided. this styles an implementation which isn't defined yet. style only what exist in the docs?

also notice that this style also applies to the code within highlight js. so the selector might need to be more specific 
image

olaf-k: > what's the difference of when it's wrapped by anchor?

links styling should take precedence

olaf-k: > even though these tags are a semantic family, they might have different types of implementations. if we want to adopt a finite state for style, side effects should be avoided. **this styles an implementation which isn't defined yet**. style only what exist in the docs?
> 
> also notice that this style also applies to the code within highlight js. so the selector might need to be more specific image

not sure what you mean by the bold part.
the idea here is to make the semantic differences in the doc clearer. the difference between the proportional and the fixed-size fonts are subtle: enough to make you think something is not uniform in the text, not strong enough to highlight what this something is.
image
adding a color shift helps parsing the code examples faster
image

and yes, it also impacts the default color in code blocks, this is a desirable side-effect (and adds consistency), but it can be removed if you think it's not useful.

yinonov:
consider https://github.com/Vonage/vivid-3/pull/766#issuecomment-1297015417

PR #792, opened 2022-11-05
fix(docs): minor touch-ups

olaf-k:


libs/components/src/lib/card/README.md
```diff
@@ -131,7 +131,7 @@ The number of lines are controlled by css variable:
 
 - `--headline-line-clamp`
 - `--subtitle-line-clamp`
-
+<!-- -->
```
yinonov, line 134: why not simply a line break?

rachelbt: it was a requested & designed  by @jshenkman:
https://vonage.slack.com/archives/C013F0YKH99/p1634197624060000

olaf-k: > why not simply a line break?
a new line in the middle of a list spaces items evenly
image
to separate two contiguous lists, we could use two `<br>` or one comment, I chose the latter
image

yinonov: Each variable should defined separately. And while at it, these are [css syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/@property/syntax) types. Calling it a string do to css pre parsing should be wrong here. These are integers.

If already refactoring, can we align to css syntax?

olaf-k: sure

libs/components/src/lib/card/README.md
```diff
@@ -124,29 +124,46 @@ Assign nodes to `main` slot to fully override a card's predefined flow and style
...
+  subtitle="here is the card text"
+>
+</vwc-card>
+```
+### Trim subtitle
+
+The card subtitle can be trimmed to your preferable number of lines.
+The number of lines is controlled by css variable `--subtitle-line-clamp`.
+
+- Type: `number`
```
yinonov, line 154: it must be a whole number (not fractioned)
```suggestion
- Type: `integer`
```

libs/components/src/lib/card/README.md
```diff
@@ -124,29 +124,46 @@ Assign nodes to `main` slot to fully override a card's predefined flow and style
...
-The card headline and the subtitle can be trimmed to your preferable number of lines.
-The number of lines are controlled by css variable:
+The card headline can be trimmed to your preferable number of lines.
+The number of lines is controlled by the css variable `--headline-line-clamp`.
 
-- `--headline-line-clamp`
-- `--subtitle-line-clamp`
-
-- Type: `string`
+- Type: `number`
```
yinonov, line 132:
```suggestion
- Type: `integer`
```

libs/components/src/lib/layout/README.md
```diff
@@ -179,8 +179,8 @@ Use the `gutters` attribute to set the initial preferred `margin`.
 Controls the grid-template-columns of the layout.
 
 - `--layout-grid-template-columns`
```
yinonov, line 181:
```suggestion
Control the grid-template-columns of the layout by setting `--layout-grid-template-columns`.

```

yinonov (changes requested):

yinonov (approved):

PR #800, opened 2022-11-07
fix(docs): horizontal scroll in mobile main page

rinaok:
now:
Screenshot 2022-11-07 at 15 44 01

after fix:
Screenshot 2022-11-07 at 15 46 20

### DIFF
--- a/libs/components/README.md
+++ b/libs/components/README.md
@@ -49,7 +49,7 @@ Only one theme is required to be loaded.
 
 ### Fonts (Prerequisite)
 
-- `node_modules/@vonage/vivid` `/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
+- `node_modules/@vonage/vivid/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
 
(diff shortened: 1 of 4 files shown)


olaf-k (dismissed):
LGTM (iPhone/iPad)

libs/components/README.md
```diff
@@ -49,7 +49,7 @@ Only one theme is required to be loaded.
 
 ### Fonts (Prerequisite)
 
-- `node_modules/@vonage/vivid/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
+- `node_modules/@vonage/vivid` `/styles/fonts/spezia.css` - Loads the *Spezia* variable font and defines its font face values. *Spezia* is Vonage's branded font and is required by most Vivid components. folder also contains the font files.
```
yinonov, line 52: I don't get it

yinonov (changes requested):

libs/components/README.md
yinonov: ok I now get it. code tags don't break lines. something like `word-break: break-word;` applied to `code` would resolve the issue for good even if occurs in other docs.

question is, should we add it to the core styles package or the docs website

yinonov (changes requested):

libs/components/README.md
rinaok: I added it to the article.

yinonov (approved):

Issue #770, opened 2022-10-31
[docs] document css public references


yinonov:
some of the Vivid design identities are referred to by template styles.
to customize components style in a systematic level (rather than per-mutating individual components) tokens are provided in run-time (e.g. typography base, size-density etc'.)

document and provide usage examples in docs

_Originally posted by @yinonov in https://github.com/Vonage/vivid-3/issues/768#issuecomment-1296700785_

yinonov:
featured options

```css
--vvd-font-family-upright: arial; // customize the default font
--vvd-font-family-monospace: monospace; // customize the code font
--vvd-size-font-scale-base: 1rem; // customize the typography scale's base font size
--vvd-size-density: -1; global density integer is used to scale all components sizes. allowed values are 0, 1, -1 (0 being default)
```

yinonov:
In addition to documenting density, the site or code blocks demos can feature a control to choose density value

yinonov:
should add typography scale to the scope of this issue

![image](https://user-images.githubusercontent.com/15140652/199458874-b3982a0a-dc5b-4553-865d-f23bb0b7f50c.png)

yinonov:
regarding `--vvd-size-font-scale-base: 1rem;` -
Vivid default base is 16px.
the idea behind setting the base to `1rem` is to align back with browser's default font size where all other typefaces descend from. this is an accessibility issue where if a user enlarge their browser's font size, the component should follow
![image](https://user-images.githubusercontent.com/15140652/199531177-05738ba7-34e3-4c73-91f3-ae57116d1b81.png)

olaf-k:
> In addition to documenting density, the site or code blocks demos can feature a control to choose density value

I'll address that in a separate PR.

PR #791, opened 2022-11-04
feat(docs): add doc for public CSS references

olaf-k:
Will close #770
Works better with #789
Looks nicer with #790

PR #801, opened 2022-11-07
feat(docs): add density button to code block demo

olaf-k:
[as discussed here](https://github.com/Vonage/vivid-3/issues/770#issuecomment-1303916622)

### DIFF
--- a/apps/docs/_layouts/base.njk
+++ b/apps/docs/_layouts/base.njk
@@ -82,7 +82,7 @@
     <script type="module" src="{{ '/assets/modules/components/icon/index.js' | url}}"></script>
     <script type="module" src="{{ '/assets/modules/components/card/index.js' | url}}"></script>
     <script type="module" src="{{ '/assets/modules/components/header/index.js' | url}}"></script>
-    <script type="module" src="{{ '/assets/modules/components/action-group/index.js' | url}}"></script>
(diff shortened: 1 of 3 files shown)


yinonov:
nice!🚀  we still need to ensure #780 resolves properly. and probably have this reviewed from a design perspective @AyalaBu

rachelbt (commented):
@yinonov de we want to expose it?
I find it a bit confusing with no explanation what is the density for. 
User may use it instead of the sizes that we offer.
Also, if this is to be merged ,we need to first merge the change of "density" property to Size before we add this.

olaf-k:
> I find it a bit confusing with no explanation what is the density for.

[The doc PR](https://github.com/Vonage/vivid-3/pull/791/files#diff-2a2b78a5afbea89fab9c9618e96a4f23baa23713dd777583346a589c5ffab3c1R45) would be a prerequisite indeed.

rachelbt:
> is for it to be a low profile interface and not be exposed at all?

I think `--vvd-size-density` is for high-level users and it should be a low profile interface and not be exposed as a button to play with. 
We should document it as in the other PR, but not present it and cause our user to use it wrongly (no one will read the doc unless they will need it....) they will just set the density instead-of using "size", and this will be wrong usage

PR #802, opened 2022-11-07
refactor(components): stabilize appearance

yinonov:
fixes many discrepancies accross components and re-unifies deviations where outline-fill-text model is applied


yinonov:
in relevance to #805

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -93,6 +93,7 @@
 	}
 
 	.icon {
+		color: var(#{constants.$vvd-color-neutral-600});
```
olaf-k, line 96: the icon has now a different color to the item's text, is it on purpose?
image

olaf-k: same for `nav-disclosure` / `nav-item`

olaf-k (commented):
The placeholders in the number|text fields are the same color as the user input, it probably should not.
image

olaf-k (commented):
Labels (for checkbox, number/text field, etc.) also don't change color when the widget is disabled. Is that on purpose?

libs/components/src/lib/accordion-item/accordion-item.scss
yinonov: yes. on purpose. that's the original design and still need to be addressed again as the question was raised. it's by design spec and has been deviated from accidentally

yinonov:
> The placeholders in the number|text fields / text-area are the same color as the user input, it probably should not. image

nice catch! resolved

yinonov:
> 

resolved. readonly should still be interactive

yinonov:
> Labels (for checkbox, number/text field, etc.) also don't change color when the widget is disabled. Is that on purpose?

they shouldn't. the change is deliberate

olaf-k (approved):
We can discuss the radio/cursors/readonly state outside of this PR. I'll have a look at the latter.

rachelbt:
> Not related to your changes but in scope of appearance consistency:
> 
> * number/text fields & text-area lack the `not-allowed` cursor in disabled state - https://github.com/Vonage/vivid-3/issues/811
> * the readonly state for checkbox/radio/switch should probably not keep the `pointer` cursor - not sure - need to be checked.

olaf-k (approved):

PR #806, opened 2022-11-08
fix(components): fixes radio-group focus when clicked

### LINKED ISSUE #803: [radio-group] mouse click fail to activate
mouse click on radio button within a radio-group does not activate the radio group. It does work with the keyboard when tabbed onto.

the expectation is that the mouse click will set focus on the group to allow the radio button to be navigated with the keyboard.

this could be caused by a css property. listbox had a similar issue and was fixed by setting a css property. ask @rinaoka @rachelbt for details.

olaf-k:
fixes #803 

note: the irony is that [this test](https://github.com/Vonage/vivid-3/blob/919a1b6f5c34e6983515f4c6d3319df8c953ebd7/libs/components/src/lib/radio-group/radio-group.spec.ts#L132) should have failed but seems to work with jsdom... this should be tested with PW.


libs/components/src/lib/radio/index.ts
```diff
@@ -10,10 +10,7 @@ import { RadioTemplate as template } from './radio.template';
 export const vividRadio = Radio.compose<RadioOptions>({
 	baseName: 'radio',
 	template: template as any,
-	styles,
-	shadowOptions: {
```
yinonov, line 14: can't we still, at least, delegate focus to the base wrapper? then continue handling focus indicator and style within template?

olaf-k: that's what was done before and didn't work. delegating focus breaks keyboard handling.

yinonov (approved):

PR #810, opened 2022-11-09
fix(components): radio.isInsideRadioGroup()

olaf-k:
small fix, long explanation ;)

- the `radio` component code relies on a private method called `isInsideRadioGroup` (the name is self-explanatory).
- this method works by checking the `.closest("[role=radiogroup]")` which doesn't work in our implementation since the role is not directly exposed on the host but in the shadow dom.
- since it's private, the choice was either to override it in a messy way or to just omit the intermediate `.base` element.

this PR is the result of choosing the latter.


yinonov:
> small fix, long explanation ;)
> 
> * the `radio` component code relies on a private method called `isInsideRadioGroup` (the name is self-explanatory).
> * this method works by checking the `.closest("[role=radiogroup]")` which doesn't work in our implementation since the role is not directly exposed on the host but in the shadow dom.
> * since it's private, the choice was either to override it in a messy way or to just omit the intermediate `.base` element.
> 
> this PR is the result of choosing the latter.

you know, I wonder why closest argument isn't passing a class getter. than it would be quite easy to customize by simply overriding it

yinonov (approved):
> small fix, long explanation ;)
> 
> * the `radio` component code relies on a private method called `isInsideRadioGroup` (the name is self-explanatory).
> * this method works by checking the `.closest("[role=radiogroup]")` which doesn't work in our implementation since the role is not directly exposed on the host but in the shadow dom.
> * since it's private, the choice was either to override it in a messy way or to just omit the intermediate `.base` element.
> 
> this PR is the result of choosing the latter.

you know, I wonder why closest argument isn't passing a class getter. than it would be quite easy to customize by simply overriding it.
Worth a fast issue / PR? it's a convention they should probably adopt

olaf-k:
> you know, I wonder why closest argument isn't passing a class getter. than it would be quite easy to customize by simply overriding it. Worth a fast issue / PR? it's a convention they should probably adopt

forgot about this PR. yes, good point, added to my todo.

PR #815, opened 2022-11-10
fix(listbox): fix disable class

rinaok:
The fix is as in fast: https://github.com/microsoft/fast/blob/6fae6f01978c40a52121aed6700f4429e12e69af/packages/web-components/fast-foundation/src/listbox/stories/listbox.register.ts


yinonov:
> @AyalaBu Is this a bug? (the listbox is disabled but the option inside is selected)
> 
> 
> 
> Screenshot 2022-11-10 at 18 24 15
> 
> 
> 
> ```
> 
> <fast-listbox disabled>
> 
>     <fast-option value="1">Option 1</fast-option>
> 
>     <fast-option selected value="2">Option 2</fast-option>
> 
>     <fast-option value="3">Option 3</fast-option>
> 
> </fast-listbox>
> 
> 
> 
> ```

I think we need to address this question to ourselves on the legibility of such permutation from the pattern perspective

AyalaBu:
@yinonov @rinaok 
I think so
Disabled can't be selected

rinaok:
@yinonov @AyalaBu 

I tried to check if option can be selected and disabled at the same time, and it looks like it can:

Screenshot 2022-11-13 at 11 25 09

libs/components/src/lib/listbox/listbox.scss
```diff
@@ -31,6 +31,13 @@
 	border-radius: 6px;
 	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
 	contain: paint;
+	gap: 4px;
```
yinonov, line 34: handled in #816

libs/components/src/lib/listbox/listbox.scss
```diff
@@ -31,6 +31,13 @@
...
+
+	&.disabled ::slotted(*) {
+		cursor: not-allowed;
+		opacity: 0.3;
```
yinonov, line 38: by design?

libs/components/src/lib/listbox/listbox.scss
```diff
@@ -31,6 +31,13 @@
...
+		pointer-events: none;
```
yinonov, line 39: we don't want to rely on css to disable the child options

yinonov (changes requested):

yinonov:
> @yinonov @AyalaBu
> 
> I tried to check if option can be selected and disabled at the same time, and it looks like it can:
> 
> Screenshot 2022-11-13 at 11 25 09
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

ok then we should not add any logic not backed up by spec

libs/components/src/lib/listbox/listbox.scss
yinonov: ok, noticed that's what fast implement as well. It's not sufficient as the styling alone cannot be implemented on the warpping listbox. each individual option should apply its own disabled state for the `disabled` class to be set and apply disabled appearance. I believe @olaf-k is facing the same issue with radio-group and radio-buttons. can you please discuss and propose a feasible solution?

rinaok: @yinonov This is how I implemented it. We can now see that it causes a lot of problems. If you want to remove the disabled class from the listbox, you need to remember the previous disable state each option had. Fast's solution seems easier and more logical to me. What do you think?

libs/components/src/lib/listbox/listbox.scss
rinaok: @AyalaBu the design is still in progress, right?

libs/components/src/lib/listbox/listbox.scss
yinonov: I agree, but it's not a sufficient solution due to the above reasons. as mentioned, there is a similar issue in radio should be resolved in a similar manner. let's discuss it live

yinonov: What's needed here is an observable for a parent disabled state `parentDisabled`. Such property can reflect disabled class on options. The css implementation is kind of hacky. Be advised by the fast team for workarounds

olaf-k: the issue seems to boil down to "should the parent component keep its children's state?"
if yes, we need to add logic to keep track of that (in `listbox`, in `radio-group`, elsewhere?)
if no, it's the application responsibility to do so

AyalaBu:
@yinonov 
About option selected and disabled at the same time:
this is kind of the same scenario like with the radio button:
it is an option that was chosen previously in a settings or something like that,
but again, it doesn't happen a lot

libs/components/src/lib/listbox/listbox.ts
```diff
@@ -23,24 +23,4 @@ export class Listbox extends FoundationListboxElement {
 	 * HTML Attribute: appearance
 	 */
 	@attr appearance?: LisboxAppearance;
```
olaf-k, line 25: not related to this PR but `LisboxAppearance` should be `ListboxAppearance`

rinaok:
@yinonov @olaf-k So what have we decided about it? Should we have a meeting?

olaf-k:
> @yinonov @olaf-k So what have we decided about it? Should we have a meeting?

I'm in favor of the parent keeping track of its children's state when changing its own (as opposed to make the children aware of the parent) but I'm open to discussion. the CSS solution seems too brittle.

acelletti:
Add JIRA ticket [VIV-866 - Investigate disable state bug in Fast](https://jira.vonage.com/browse/VIV-866)

PR #816, opened 2022-11-10
chore(components): align with design spec

### LINKED ISSUE #805: [Development check up]: things that need attention
- [x] _Disabled text color_ must be chosen based on fieldset - 
  - **text/border neutral 400, background 200**

- [x] Ghost disabled appearance model - 
  - **inner content (text, icon) - neutral 400**

- [x] What happens to the _field label_ in different states? -  
  - **I think it stays the same, still would like to check**

- [x] _Read only field_ - 
 - **background: neutral 100**
 - **border: neutral 400**
 - **icons, label, helper text: neutral 600**
 - **Input text: neutral 1000**

  @yinonov 

- [ ] _Chosen state_ - **The background will be changed to neutral 200** 
**since 1000 creates kind of a dark mode situation with inner components**  @yinonov 

- [ ] _Listbox with options_ (multiple select) - is it problematic? How should be handled? - 
  - @yinonov  The a11y proposes the solution of checkboxes with options that create a checklist:

https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-select-lists

But I also saw this really cute solution with tags in a few places:

https://ant.design/components/select/

https://atlassian.design/components/select/examples

But the issue that needs to be considered is what happens when the field is full?
What I saw is that it just becomes higher.
Probably it is the same treatment with the checkboxes option.

Honestly, I prefer the checkbox solution,
But basically what are your thoughts about this?

- [x] Should there be an appearance difference between _"selected"_ and _"checked" states_ - 

@yinonov 
  - Based on what a saw in ARIA:
    https://w3c.github.io/aria-practices/examples/#examples_by_props_label

   _"Checked"_ state applies only to a handful of components that need special treatment:
1. checkbox
2. radio button
3. menubar (the checked option we discussed for the menu-item)
4. switch
5. toolbar

Each and every one of them is a completely separate component in which the "checked" state looks different.

If we give the _"selected"_ state to the entire box (the moment the choice is being made, and a ui component with a text have a certain background color for example) then it is something different, as it gives the indication about a choice being made.

- [x] Add _distance_ (padding) between options - 
  - **4px distance between the option items**
  - **2px padding-top + padding-bottom** 
  Just like here:

  https://www.figma.com/file/QG5hvuw5qsfS0XYMVJ6V3L/Vivid-UI-Kit---Exploration?node-id=4278%3A61829
  
  @yinonov 

- [ ] Check what happens to _padding_ when a component has icon + text / icon only - 
  - For the most part when there is only icon left in the component (like a button or a badge) for example the padding becomes the same on all sides, which creates a square/circle shape (depending on the border-radius).
  The only different component is nav-item, when there is another padding additional the the padding around the icon, in order to let it fit when the nav-items are open and have a text.
  @yinonov 

- [ ] _Clickable/unclickable icon_ - what color should it receive? - 
  - **After some research and meditation:**
     I agree with Joella's convention,
     by which unclickable  (decoration) icon will be 600,
     & clickable icon will be 1000.
     
     Although, I wonder, maybe we can consider allowing cta color to clickable icon?
     
     @yinonov @rachelbt 

- [x] _Note component_ - check connotations needed, and remove the unnecessary ones - 
  - **Based on the research I did the connotations we have appear in the most of design systems, the only change I would propose is to change announcement (magenta) connotation to confirmation (accent)** @yinonov 

- [x] Should _outline appearance_ receive a background color? - 
  - @yinonov So this is the situation:
    **Ant design:** _No bg_
    **Carbon:** _Present but with opacity 0_
    **Fluent:** _Present_
    **Fast:** _No bg_
    **Semantic UI:** _No bg_
    **Bootstrap:** _No bg_
    **Lightning:** _No bg_

So basically we can see that there isn't one single practice that is always present,
although what I saw in more cases was the transparent background,
Which is the option I, personally, like muck more

yinonov:
Closes #805


libs/components/src/lib/listbox/listbox.scss
```diff
@@ -26,11 +26,14 @@
...
 
 	position: relative;
+	display: flex;
+	flex-direction: column;
 	padding: 4px;
 	background-color: var(#{appearance.get-appearance-token(fill)});
 	border-radius: 6px;
 	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
 	contain: paint;
+	gap: 4px;
```
rachelbt, line 36:
```suggestion
	gap: 2px;
```
why did you change it to 4?

libs/components/src/lib/nav-disclosure/nav-disclosure.scss
```diff
@@ -73,7 +73,11 @@
 }
 
 .content {
+	display: flex;
+	flex-direction: column;
 	border-inline-start: 1px solid var(#{constants.$vvd-color-neutral-200});
+	gap: 4px;
```
rachelbt, line 79: same.
Is the gap was supposed to be 4 to start with?

libs/components/src/lib/note/note.template.ts
```diff
@@ -11,7 +11,7 @@ import type { Note } from './note';
 
 const connotationIconMap = new Map([
 	[Connotation.Information, 'info-solid'],
-	[Connotation.Announcement, 'megaphone-solid'],
```
rachelbt, line 14: no pink note :(

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
```diff
@@ -104,9 +104,9 @@ $states-mapping: (
       outline: transparent,
     ),
     fieldset: (
-      text: var(#{constants.$vvd-color-neutral-800}),
-      fill: var(#{constants.$vvd-color-neutral-200}),
-      outline: var(#{constants.$vvd-color-neutral-500}),
+      text: var(#{constants.$vvd-color-canvas-text}),
```
rachelbt, line 107: how come this change didn't failed the text-field, textarea, numberfield ui tests?

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
```diff
@@ -119,23 +119,6 @@ $states-mapping: (
       outline: var(#{constants.$vvd-color-neutral-400}),
     )
   ),
-  // readonlyAndHover: (
```
rachelbt, line 122: do we still need this?

libs/components/src/lib/listbox/listbox.scss
yinonov: Consist with our unit

libs/components/src/lib/nav-disclosure/nav-disclosure.scss
yinonov: I believe so

libs/shared/src/lib/sass/mixins/appearance/_variables.scss
yinonov: good catch. apparently it won't catch those slight differences. updated them anyway now

libs/components/src/lib/listbox/listbox.scss
```diff
@@ -33,7 +33,7 @@
 	border-radius: 6px;
 	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
 	contain: paint;
-	gap: 4px;
+	gap: 2px;
```
yinonov, line 36: @rinaok @rachelbt why 2?

rachelbt (approved):
For the record - I'm sad that the note pink variation is removed.
Apart from that - LGTM

PR #817, opened 2022-11-10
feat(components): direction support for listbox

yinonov:
this maneuver easily features a variant of listbox similar to the design [button-toggle-group](https://vivid.vonage.com/?path=/story/components-button-toggle-group--pill-shape) had. listbox, most likely, provides better semantics.

Disclaimer:

1. as block sizing could only be affected by the listbox child items, and as child items are, in practice, options, this may conflict. an option should not feature sizing. [density increment](https://vivid.deno.dev/designs/design-tokens#sizing) could be used instead

2. same goes for shape

- [x] add keydown handling for horizontal arrows interaction

This enablement can also be provided by a dedicated component or action group usage example.

should certainly be investigated to choose best role and whether keyboard handling is required and how


libs/components/src/lib/listbox/listbox.scss
```diff
@@ -26,11 +26,23 @@
...
+	gap: 2px;
+	inline-size: 100%;
+
+	&:not(.direction-horizontal) {
+		flex-direction: column;
+	}
+
+	&.direction-horizontal {
+		flex-direction: row;
+	}
```
rachelbt, line 45:
```suggestion

```
this is the default of flex you don't need to specify it

rachelbt (commented):
so what if one wants to have a toogle-icon-buttons group?
I think that the shape is also needed...no?

libs/components/src/lib/listbox/listbox.scss
yinonov: 10x

yinonov:
> so what if one wants to have a toogle-icon-buttons group? I think that the shape is also needed...no?

that's an issue yeah

yinonov:
> so what if one wants to have a toogle-icon-buttons group? I think that the shape is also needed...no?

@rachelbt listbox can set styles on its child options
```css
border-radius: 24px;
overflow: hidden;
```

yinonov:
> > so what if one wants to have a toogle-icon-buttons group? I think that the shape is also needed...no?
> 
> @rachelbt listbox can set styles on its child options
> 
> ```css
> border-radius: 24px;
> overflow: hidden;
> ```
> 
> within a select, options will need to be sharpened. it's more of a reason to apply shaping style from the wrapper image

@rachelbt applied

libs/components/src/lib/listbox/listbox.ts
```diff
@@ -24,6 +32,67 @@ export class Listbox extends FoundationListboxElement {
...
+	 * @internal
+	 */
+	private horizontalKeydownHandler(e: KeyboardEvent): boolean | void {
+		if (this.disabled) {
+			return true;
+		}
+
+		this.shouldSkipFocus = false;
+
+		const key = e.key;
```
olaf-k, line 64: not needed

libs/components/src/lib/listbox/README.md
```diff
@@ -56,6 +56,39 @@ Set the `appearance` attribute to change the listbox's appearance.
...
+  <vwc-option value="1" text="Option"></vwc-option>
+  <vwc-option value="2" text="Option"></vwc-option>
+  <vwc-option value="3" text="Option"></vwc-option>
+</vwc-listbox>
+```
+
+### Shape
+
+Use the `shape` attribute to set the action-Group's edges.  
+When using shape - pay in mind setting the slotted elements with the same shape property.
```
olaf-k, line 77: "When using this property, remember to set to slotted elements to the same value."

didn't play with the code yet, but couldn't this be done from the listbox directly instead of leaving the responsibility to the user?

libs/components/src/lib/listbox/README.md
```diff
@@ -1,6 +1,6 @@
 # Listbox
 
-Listbox displays list of [vwc-option](../../components/option) and allow users to select from them. 
+Listbox displays list of [vwc-option](../../components/option) and allow users to select from them.
```
olaf-k, line 3: ^allows

olaf-k (changes requested):
the change in orientation should update the keyboard handler: in the current deployment, changing a vertical listbox to a horizontal one leaves the up/down keys responsible for moving

yinonov:
> there's something weird with the keyboard handlers: in the current deployment, changing a vertical listbox to a horizontal one leaves the up/down keys responsible for moving

it's the same for radio-group and it actually feels right. I wonder why horizontal arrow keydown don't trigger vertical navigation OOTB as well

libs/components/src/lib/listbox/README.md
yinonov: checked with a grammer checker. both versions seem to be satisfactory 🤷‍♂️

yinonov, line 3:
```suggestion
Listbox displays a list of [vwc-option](../../components/option) and allow users to select from them.
```

libs/components/src/lib/listbox/README.md
yinonov: no. it can easily be done by authors. component intervention is a last resort

yinonov: actually, in this particular case, is IS done by the listbox

libs/components/src/lib/listbox/listbox.ts
yinonov: how did TS not catch that? 🤔

libs/components/src/lib/listbox/README.md
olaf-k: > checked with a grammer checker. both versions seem to be satisfactory 🤷‍♂️

^grammar 😁

olaf-k:
> it's the same for radio-group and it actually feels right. I wonder why horizontal arrow keydown don't trigger vertical navigation OOTB as well

`radio-group` doesn't update its keyboard handler on direction change and uses up/down along with left/right indifferently.
the current behavior of `listbox` does not feel right (try to add `direction="horizontal"` in the first doc example)

libs/components/src/lib/listbox/README.md
yinonov: I'll have to trust you on that one

libs/components/src/lib/listbox/README.md
```diff
@@ -1,6 +1,6 @@
 # Listbox
 
-Listbox displays list of [vwc-option](../../components/option) and allow users to select from them. 
+Listbox displays a list of [vwc-option](../../components/option) and allow users to select from them.
```
yinonov, line 3:
```suggestion
Listbox displays a list of [vwc-option](../../components/option) and allows users to select from them.
```

acelletti:
Moved to [VIV-863 - Investigate direction support for listbox](https://jira.vonage.com/browse/VIV-863)

PR #818, opened 2022-11-12
build(components): generate custom.elements.json

yinonov:
custom elements manifest is a schema published as a JSON and allows tooling and IDEs to give rich information about the custom elements in a given project.

for more - https://github.com/webcomponents/custom-elements-manifest

required steps -

- [ ] run `cem analyze --config "custom-elements-manifest.config.mjs"` pre-packaging step
- [ ] generate readme / .md from base classes. code snippets could be included in JSDocs comments and rendered by plugins (as [plugins: [jsdocLinkFix()]](https://github.com/yinonov/fast/blob/e0df1e06ae6674b0bf9adbd2a346b720f83c6b6c/packages/web-components/fast-foundation/custom-elements-manifest.config.mjs#L23-L24)) (can be extracted to a following scope)

assistive tools
https://github.com/open-wc/custom-elements-manifest

PR #821, opened 2022-11-13
feat(docs): adopt `@microsoft/api-extractor` & `@custom-elements-manifest/analyzer`

k-paxian:
In order to ship components meta data in JSON format alongside the NPM package
- Generate `customElements` manifest
- Generate vivid components api model, to access Enum constants in the first place


libs/components/project.json
```diff
@@ -3,6 +3,16 @@
   "sourceRoot": "libs/components/src",
   "projectType": "library",
   "targets": {
+    "gen": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [
```
yinonov, line 9: @YonatanKra should run these commands pre-packaging step

libs/components/project.json
```diff
@@ -3,6 +3,16 @@
...
+          "cem analyze --config cem.config.mjs",
+          "api-extractor run -c ./libs/components/api-extractor.json"
```
yinonov, line 11: should evaluate as we do not follow TSDoc guidelines completely

yinonov (commented):
@k-paxian thanks! better starting point than #818

.github/workflows/_publish-package.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         run: |
           npm ci --ignore-scripts
           npx nx run components:build
+          npx nx run components:gen
```
yinonov, line 48: should this be ran manually or a "prestart" implementation in project.json

cem.config.mjs
```diff
@@ -0,0 +1,31 @@
...
+    /** Run in dev mode, provides extra logging */
+    dev: false,
+    /** Enable special handling for fast */
+    fast: true,
+    plugins: [
+        jsdocFunctionPlugin(),
+        jsdocExamplePlugin(),
+        asyncFunctionPlugin(),
+        readonlyPlugin()
+    ]
```
yinonov, line 30: following steps, create plugin to render code block within comments. eventually READMEs should be autogenerated

libs/components/project.json
k-paxian: no worries, there is a config ignoring inconsistency for now, it can be un-ignored gradually later
image

.github/workflows/_publish-package.yml
k-paxian: it's more like `prepublish` action and for now the only spot for publishing is in the workflow area 🤷

libs/components/project.json
k-paxian: these needs to be run anywhere after the build process, since api-extractor relying on `index.d.ts` produced during the build

yinonov (changes requested):

README.md
```diff
@@ -75,7 +75,7 @@ This project is [Apache 2.0][license-url] licensed.
 
 <hr>
 
-##### Built With [Typescript](https://www.typescriptlang.org), [Lit](https://lit.dev), [Sass](https://sass-lang.com) and ❤️
+##### Built With [Typescript](https://www.typescriptlang.org), [Fast](https://fast.design), [Sass](https://sass-lang.com) and ❤️
```
yinonov, line 78: good one

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
yinonov, line 1: how does it affect installations?

package.json
```diff
@@ -84,6 +90,7 @@
 		"prettier": "2.7.1",
 		"rollup-plugin-copy": "^3.4.0",
 		"rollup-plugin-delete": "^2.0.0",
+		"rollup-plugin-postcss": "^4.0.2",
```
yinonov, line 93: why? I vaguely recall it was removed intentionally. need it here?

libs/components/project.json
```diff
@@ -3,6 +3,16 @@
   "sourceRoot": "libs/components/src",
   "projectType": "library",
   "targets": {
+    "gen": {
```
yinonov, line 6: maybe more explicit?

```suggestion
    "generateMeta": {
```

yinonov (changes requested):

libs/components/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './lib/components';
+export * from './lib/enums';
```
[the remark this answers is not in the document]
k-paxian: while specifying the attributes we are using string literals for certain values like 
image

But we would like to use const references there like
```jsx
import { Connotation, Layout } from '@vonage/vivid-react'

          <VwcButton
            connotation={Connotation.CTA}
            layout={Layout.Outlined}
          />
```

.npmrc
k-paxian: controversial one, we usually use yarn to install packages, just to have both `yarn.lock` & `package.lock` side by side looks a bit odd, even worse, yarn is making use of `package.lock` if it exists. Don't mind to discard this change if you feel it's not useful.

package.json
k-paxian: I don't remember exactly, but at some point during the `build` script run on a windows box it desperately asks for this dependency 🤷 worth checking at least why is it happening?

libs/components/project.json
```diff
@@ -3,6 +3,16 @@
...
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [
+          "cem analyze --config cem.config.mjs",
```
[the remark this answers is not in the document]
k-paxian: yes, works fine by me on a windows machine

package.json
k-paxian: if it shouldn't be there feel free to discard it

.npmrc
yinonov: if it doesn't have any affect on the result of generated files of this scope's target, let's revert this one.
you may have a point here but I recall having trouble with packages not aligning to semver in the past, can extract it this to another tiny PR?

libs/components/src/index.ts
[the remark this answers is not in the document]
yinonov: sure - we thought of refactoring to const be there was no such requirement until now.

- [ ] refactor enums to consts
- [ ] export enums / consts subset types
- [ ] remove [libs/components/src/index.ts](https://github.com/Vonage/vivid-3/pull/821/files/cdc7a88ad75cbaf307dffaf02e8a4e70ef246f4b#diff-9068f8a41f226453682b9874c9951a886671b0552e4d8baf2447404f1091d6b9) once done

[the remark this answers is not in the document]
yinonov: Vonage/vivid-3#844

yinonov (dismissed):

yinonov:
can you please run `npx bachball change` to generate change file?

yinonov (approved):

PR #825, opened 2022-11-13
feat(components): combobox new component

### LINKED ISSUE #824: [combobox] create a new component

yinonov:
Closes #824

- [x] add support for popup width (or more specifically its assigned node)
- [ ] enable text field icon trailing, or compose a dedicated text field with alternate style (can be customized in element definition). or just use plain input in template


libs/components/src/lib/option/option.scss
```diff
@@ -30,7 +30,6 @@
 	box-sizing: border-box;
 	align-items: center;
 	background-color: var(#{appearance.get-appearance-token(fill)});
-	border-radius: 6px;
```
yinonov, line 33: requested by design?

libs/components/src/lib/combobox/ui.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+test('should show the component', async ({ page }: { page: Page }) => {
+	const template = extractHTMLBlocksFromReadme(
+		path.join(new URL('.', import.meta.url).pathname, 'README.md')
+	).reduce(
+		(htmlString: string, block: string) =>
+			`${htmlString} <div style="margin: 5px;">${block}</div>`,
+		''
+	);
+
```
yinonov, line 20: set the viewport width to not waste extra space and a large image

libs/components/src/lib/option/option.scss
rinaok: @yinonov IDK... It is your change: https://github.com/Vonage/vivid-3/pull/825/commits/418d54dbe942418ac7a935e76efba7b6bc3a87f4
I can remove it if not needed.

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -0,0 +1,47 @@
...
+	padding: 4px;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	border-radius: 6px;
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	contain: paint;
+}
+
+.popup {
+	--_popup-width: 209px;
+}
```
rinaok, line 47: @rachelbt How I make the popup be the same width as the combobox?

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -0,0 +1,47 @@
...
+
+	padding: 4px;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	border-radius: 6px;
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	contain: paint;
+}
+
+.popup {
+	--_popup-width: 209px;
```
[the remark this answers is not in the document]
rinaok: @rachelbt Tried that but it didn't work, that's why I put 209px

libs/components/src/lib/popup/popup.scss
```diff
@@ -1,26 +1,31 @@
...
 	}
 }
 
 .popup-wrapper {
-	position: fixed;
+	&:not(.absolute) {
+		position: fixed;
+	}
+	&.absolute {
+		position: absolute;
```
rachelbt, line 20:
```suggestion
		position: absolute;
		inline-size: var(#{variables.$popup-width}
```

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -0,0 +1,47 @@
...
+);
+@use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
+
+@use "../../../../shared/src/lib/sass/mixins/appearance/config" as appearance-config with (
+	$appearances: fieldset ghost,
+	$states: idle disabled,
+	$default: fieldset
+);
+@use "../../../../shared/src/lib/sass/mixins/appearance" as appearance;
+
```
rachelbt, line 15:
```suggestion
:host {
  position: relative;
  }
```

rinaok:
> Closes #824
> 
> * [x]  add support for popup width (or more specifically its assigned node)
> * [ ]  enable text field icon trailing, or compose a dedicated text field with alternate style (can be customized in element definition). or just use plain input in template

@yinonov Are you sure we need icon inside the combobox? I think that chevron is enough .

libs/components/src/lib/combobox/README.md
```diff
@@ -66,10 +66,10 @@ Add a `placeholder` attribute to add placeholder text to the input.
 ```
 ### Autocomplete
 
-- Type: `'inline'` | `'list'` | `'both'` | `undefined`
+- Type: `'inline'` | `'list'` | `'both'`
```
yinonov, line 69: good fix. should we be adding `none`? https://github.com/microsoft/fast/tree/6ebb5ec7bdb7666f49ddc524bc56f70a2e8aa574/packages/web-components/fast-foundation/src/combobox#variables

rachelbt (commented):
when opening the listbox - should the user navigate with the up & down arrows?

rinaok:
> when opening the listbox - shouldn't the user navigate with the up & down arrows?

@rachelbt Yes I know, I'm working on it

rinaok:
> > when opening the listbox - shouldn't the user navigate with the up & down arrows?
> 
> @rachelbt Yes I know, I'm working on it

@rachelbt It's working now

libs/components/src/lib/combobox/README.md
yinonov: It's a native spec. not our call

rinaok (approved):

yinonov (commented):
how does triggering the focus on an option work? on `keypress`? would it help to delegate focus in popup definition? so that focus on host will automatically decide which option to focus on (instead of manual)

rinaok:
> how does triggering the focus on an option work? on `keypress`? would it help to delegate focus in popup definition? so that focus on host will automatically decide which option to focus on (instead of manual)

@yinonov You want us to remove all Fast's key handlers from the combobox template and that the popup will handle it?

yinonov:
> rs from the combobox template and that the po

no! so it's a hindered knowledge to this scope? handled by fast? I'd still add tests to it. WDYT?

rinaok:
> 

@yinonov Yes, all the behaviour is handled by Fast

yinonov (commented):
why does the `open` section rendered snippet autofocus? I didn't see any explicit code for it (it might be intentional. just making sure)

image

change/@vonage-vivid-8865978f-8468-4691-a216-12a13c91a92f.json
```diff
@@ -0,0 +1,7 @@
+{
```
yinonov, line 1: is this redundant?

libs/components/src/lib/combobox/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+```js
+<script type="module">
+  import '@vonage/vivid/combobox';
+</script>
+```
+
+## Slots
+
+### Option
```
yinonov, line 13: it should be `default`. as in
https://github.com/Vonage/vivid-3/blob/9bafb96d81556e9cff522046acd2ca166da29b01/libs/components/src/lib/menu/README.md#L78-L81

rinaok: @yinonov updated but we need to update it also in listbox readme

libs/components/src/lib/combobox/README.md
```diff
@@ -0,0 +1,136 @@
...
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+Set the `value` attribute to set the default value for the input field. Setting the property on the element will not change the default value, but will change the value shown in the view as well as the submitted value in a form (imitating the native behavior).
+
+```html preview
+<vwc-combobox value="value">
+ <vwc-option text="First Option"></vwc-option>
+ <vwc-option text="Second Option"></vwc-option>
+</vwc-combobox>
```
olaf-k, line 51: this example does not demonstrate the `value` usage.
```suggestion
<vwc-combobox value="First Option">
 <vwc-option text="First Option"></vwc-option>
 <vwc-option text="Second Option"></vwc-option>
</vwc-combobox>
```

also, and this is beyond the scope of your PR, but why are `option` text and value the same thing? they should be be two separate props/attrs

olaf-k (changes requested):

olaf-k:
clicking on the combobox button opens the list of options, but clicking again does not close it.

olaf-k:
if you have a page with a combobox followed by a listbox, the combobox list appears under the listbox when opened:
chrome_DdqtA2DSml

libs/components/src/lib/combobox/README.md
yinonov: I think options here should have a matching value. Don't they?

yinonov:
U mean it doesn't go on top? Top-layer access issue. Referenced in our docs in multiple caveat sections

libs/components/src/lib/combobox/README.md
olaf-k: > I think options here should have a matching value. Don't they?

the FAST and native implementations both distinguish the displayed text from the form value, I was expecting our `option` to behave in the same manner.

yinonov: of course. sorry, didn't catch ur previous comment. I mean combobox value should match an option value

libs/components/src/lib/popup/partials/variables.scss
```diff
@@ -1 +1 @@
-$popup-width: --_popup-width;
+$popup-width: --popup-width;
```
rachelbt:
```suggestion
.popup {
	$popup-width: --_popup-width;
	#{$popup-width}: 50%;
}
```

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -0,0 +1,51 @@
...
+	padding: 4px;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	border-radius: 6px;
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	contain: paint;
+}
+
+.popup {
+	--popup-width: 100%;
```
rachelbt:
```suggestion
	$popup-width: --_popup-width;
	#{$popup-width}: 100%;
```

libs/components/src/lib/popup/partials/variables.scss
rachelbt:
```suggestion
$popup-width: --_popup-width;
```

rinaok:
> 

@olaf-k @yinonov It's the same behaviour as in FAST https://explore.fast.design/components/fast-combobox, do you think it's a bug?

olaf-k:
> @olaf-k @yinonov It's the same behaviour as in FAST [explore.fast.design/components/fast-combobox](https://explore.fast.design/components/fast-combobox), do you think it's a bug?

good point, I checked and [they already have an issue on their side about it](https://github.com/microsoft/fast/issues/5721).

yinonov:
> why does the `open` section rendered snippet autofocus? I didn't see any explicit code for it (it might be intentional. just making sure)
> 
> image

https://github.com/microsoft/fast/blob/6ebb5ec7bdb7666f49ddc524bc56f70a2e8aa574/packages/web-components/fast-foundation/src/combobox/combobox.ts#L336-L352

rinaok:
> U mean it doesn't go on top? Top-layer access issue. Referenced in our docs in multiple caveat sections (e.g. [tooltip#caveat](https://vivid.deno.dev/components/tooltip#caveat)). perhaps a similar caveat mention should be added for combobox

@yinonov Added Caveat

rinaok (approved):

rinaok (dismissed):

libs/components/src/lib/combobox/README.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+## Caveat
+
+Document elements display precedence is formed by the imaginary z-axis [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), commonly by order of which elements are rendered and special properties (e.g. _z-index_).
+Combobox component is a low level element, unaware of its document context, but is, in most cases, required to overlay on top of all elements.
+
+A common practice used in apps / frameworks to promote a Combobox component to top other elements z-axis, is to utilise a service that dynamically appends a Combobox component to the end of the body element, when called for.
```
olaf-k, line 143: hmm. this probably is a valid advice for modal components that appear on their own but this is hardly applicable to form components such as the combobox. I'm ok to leave the widget as it is until someone complains but we need to be prepared for another solution than "dynamically append it to body".

olaf-k (changes requested):
I'll open an issue for this option.value thingy.

libs/components/src/lib/combobox/README.md
rinaok: ok

rinaok (dismissed):

olaf-k (approved):
can you put back the rest of the disclaimer?

```
## Caveat

Document elements display precedence is formed by the imaginary z-axis [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), commonly by order of which elements are rendered and special properties (e.g. _z-index_).
Combobox component is a low level element, unaware of its document context, but is, in most cases, required to overlay on top of all elements.
```

rinaok (approved):

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -0,0 +1,53 @@
...
+
+	padding: 4px;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	border-radius: 6px;
+	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
+	contain: paint;
+}
+
+.popup {
+	$popup-width: --_popup-width;
```
yinonov, line 50: how did `--_popup-width` pass stylelint? it should have thrown an error exception.

extract to, and import from `partials/_variables.scss`

libs/components/src/lib/combobox/ui.test.ts
yinonov: @rinaok

libs/components/src/lib/combobox/combobox.scss
rinaok: done

rinaok: @rachelbt @yinonov I think that we should add default value for popup-width inside popup css, right?

Screenshot 2022-12-21 at 16 15 14

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -47,7 +48,5 @@
 }
 
 .popup {
-	$popup-width: --_popup-width;
-	
-	#{$popup-width}: 100%;
+	#{popup-variables.$popup-width}: 100%;
```
yinonov, line 51: can u explain this again? where is this variable used?

rinaok: @yinonov When popup is strategy="absolute"

yinonov: and?

yinonov: never mind followed the code

rinaok (approved):

libs/components/src/lib/combobox/combobox.template.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { Combobox } from './combobox';
+import { focusTemplateFactory } from './../../shared/patterns/focus';
+
+/**
+ *
+ */
+function renderLabel() {
+	return html<Combobox>`
+	  <label for="control" class="label">
+		  ${x => x.label}
```
olaf-k, line 16: mixed spaces with tabs here and in `comboboxTemplate` (should've been caught by linter though?)

yinonov: something's off with eslint

PR #827, opened 2022-11-14
feat(tooltip): add inert

rinaok:


libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -21,7 +21,7 @@ export const TooltipTemplate: (
 ) => ViewTemplate<Tooltip> = () => html`
 <vwc-popup class="${getClasses}" arrow alternate
 placement=${(x) => x.placement} open=${(x) => x.open} anchor=${(x) => x.anchor}
-  exportparts="vvd-theme-alternate" role="tooltip">
+  exportparts="vvd-theme-alternate" role="tooltip" inert>
```
yinonov, line 24: I know it's unrelated to the change, but why is the popup getting the _tooltip_ role? why not the _tooltip_ classed element?

yinonov (changes requested):

libs/components/src/lib/tooltip/tooltip.template.ts
yinonov: The `role` attribute is set on the popup instead of on the the dedicated tooltip template

rinaok: yeah you are right

yinonov (approved):

PR #829, opened 2022-11-14
refactor(tooltip): disable manual handling of open state

rinaok:


libs/components/src/lib/popup/popup.ts
```diff
@@ -23,6 +23,9 @@ export class Popup extends FoundationElement {
 	#cleanup?: () => void; // cleans the autoupdate
 
 	#anchorEl: Element | null | undefined;
+	get anchorElement(): Element | null | undefined {
```
yinonov, line 26: why? then this accessor might as well be a property and be assigned to the actual element

libs/components/src/lib/tooltip/README.md
```diff
@@ -21,7 +21,7 @@ Do not target non-interactive controls as a tooltip's anchor (such as non-focusa
```
yinonov, line 21: should explain what triggers the tooltip, and why (by recommended pattern) -

> A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out.

https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -1,8 +1,8 @@
 import { html } from '@microsoft/fast-element';
```
yinonov, line 1: no real change in this file, can be reduced from the PR

yinonov (changes requested):

yinonov (commented):
we should go over microsoft/fast#6468

libs/components/src/lib/tooltip/tooltip.template.ts
rinaok: moved the role inside

libs/components/src/lib/tooltip/README.md
```diff
@@ -12,45 +14,25 @@ A tooltip is a brief, informative message or descriptions or explanations for th
 
 ### Anchor
 
-The tooltip can be placed on interactive controls (things that can be hovered or focusable) such as: button, checkbox, input text.
+The tooltip shoul be placed on interactive controls (things that can be hovered or focusable) such as: button, checkbox, input text.
```
yinonov, line 17:
```suggestion
The tooltip should be placed on interactive controls (things that can be hovered or focusable) such as: button, checkbox, input text.
```

libs/components/src/lib/tooltip/README.md
```diff
@@ -1,6 +1,8 @@
 # Tooltip
 
-A tooltip is a brief, informative message or descriptions or explanations for their paired element. Tooltips in general are less accessible so be sure to follow our [accessibility recommendation](#accessibility) in the docs.
+A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out.
```
yinonov, line 3:
```suggestion
A tooltip is a popup that displays informative message, description or explanation related to an element receiving keyboard focus or when the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out. 
```

libs/components/src/lib/tooltip/README.md
```diff
@@ -1,6 +1,8 @@
 # Tooltip
 
-A tooltip is a brief, informative message or descriptions or explanations for their paired element. Tooltips in general are less accessible so be sure to follow our [accessibility recommendation](#accessibility) in the docs.
+A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out. 
+
+The tooltip should be referenced through the use of aria-describedby. The aria-describedby attribute is on the anchor element, not on the tooltip. It should be the same as tooltip's id. To change aria-describedby, change tooltip's id (the default id is 'tooltip').
```
yinonov, line 5: are we planning to feature this? if so, why document as instructions?

libs/components/src/lib/tooltip/README.md
yinonov, line 17:
```suggestion
The tooltip should be placed on interactive controls, elements that can be hovered over or focusable (e.g. button, checkbox, input text etc'.).
```

libs/components/src/lib/tooltip/README.md
```diff
@@ -12,45 +14,25 @@ A tooltip is a brief, informative message or descriptions or explanations for th
...
-<vwc-button id="anchor" icon="help-line" shape="pill" aria-describedby="tooltip"></vwc-button>
-<vwc-tooltip id="tooltip" anchor="anchor" text="I'm a tooltip" open></vwc-tooltip>
+<vwc-button id="anchor" icon="help-line" shape="pill"></vwc-button>
+<vwc-tooltip anchor="anchor" text="I'm a tooltip"></vwc-tooltip>
 ```
 
 ### Text
 
-The tooltip is a description and therefor, the tooltip itself can not be interactive and contains only text.
+The tooltip is a description and therefor, the tooltip itself can not be interactive or focusable and contains only text.
```
yinonov, line 31: first describe the text as we do in other component.

next paragraph should note -

> Because the tooltip itself never receives focus and is not in the tabbing order, a tooltip can not contain interactive elements like links, inputs, or buttons

_[mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#description)_

yinonov (changes requested):
fast mentioned something regarding control over delay via css. what's our delay now?

libs/components/src/lib/tooltip/README.md
rinaok: So you mean that the user should not know about it at all?

rinaok:
> fast mentioned something regarding control over delay via css. what's our delay now?

We don't have a delay. They remove it because they now control tooltip transitions via CSS.

I added 
`.open {
		transition: visibility 0s 0s linear, opacity 50ms 0.5s linear;
		opacity: 1;
		visibility: visible;
	}

	&:not(.open) {
		transition: visibility 0s 0.5s, opacity 50ms linear;
		opacity: 0;
		visibility: hidden;
	}`
	
	what do you think?

libs/components/src/lib/tooltip/README.md
```diff
@@ -1,6 +1,6 @@
 # Tooltip
 
-A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out. 
+A tooltip is a popup that displays informative message, description or explanation related to an element receiving keyboard focus or when the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out.
```
yinonov, line 3:
```suggestion
A tooltip displays additional, non-essential, informative message, description or explanation to a focusable element upon hover or focus. It typically appears after a short delay and disappears when `Escape` key is pressed, element looses focus or hovered off. 
```

yinonov (changes requested):

yinonov (commented):
shouldn't the visual regression fail?

libs/components/src/lib/tooltip/README.md
```diff
@@ -87,40 +71,22 @@ Use the `--tooltip-inline-size` variable to set the tooltip's inline size.
...
-
 ## Use Cases
 
 ### Usage inside text
 
 ```html preview
 <p>
   Text with tooltip - press the question mark
-  <vwc-button id="button" icon="help-line" shape="pill" aria-describedby="tooltip"></vwc-button>
+  <vwc-button id="button" icon="help-line" shape="pill"></vwc-button>
```
yinonov, line 85: this shouldn't be a button as it doesn't trigger any action.
it can be a plain icon with `tabindex="0"` and `aria-describedby`.

this is more of a toggletip implememtation. perhaps it's better to avoid this use-case and resolve to actual controls (button, text-field etc')

yinonov (changes requested):

yinonov:
> > fast mentioned something regarding control over delay via css. what's our delay now?
> 
> @yinonov We don't have a delay. Fast removed it because they now control tooltip transitions via CSS.
> 
> I added it like fast:
> 
> ```
> .open {
> 		opacity: 1;
> 		transition: visibility 0s 0s linear, opacity 50ms 0.5s linear;
> 		visibility: visible;
> 	}
> 
> 	&:not(.open) {
> 		opacity: 0;
> 		transition: visibility 0s 0.5s, opacity 50ms linear;
> 		visibility: hidden;
> 	}
> ```
> 
> ```
> what do you think?
> ```

yes that's good to reflect. we might want a default value here? maybe a css variable to control the duration? or better off, no default but a usage example of adding it via css (as you did)

rinaok:
> 

you mean that we should remove it from our css, and let the user add it?

rinaok:
> shouldn't the visual regression fail?

why?

libs/components/src/lib/tooltip/README.md
rinaok: "The tooltip can be placed on interactive controls (things that can be hovered or focusable) such as: button, checkbox, input text." 
What is the problem with the button? You can't place it on an icon.

rinaok:
> > > fast mentioned something regarding control over delay via css. what's our delay now?
> > 
> > 
> > @yinonov We don't have a delay. Fast removed it because they now control tooltip transitions via CSS.
> > I added it like fast:
> > ```
> > .open {
> > 		opacity: 1;
> > 		transition: visibility 0s 0s linear, opacity 50ms 0.5s linear;
> > 		visibility: visible;
> > 	}
> > 
> > 	&:not(.open) {
> > 		opacity: 0;
> > 		transition: visibility 0s 0.5s, opacity 50ms linear;
> > 		visibility: hidden;
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > what do you think?
> > ```
> 
> yes that's good to reflect. we might want a default value here? maybe a css variable to control the duration? or better off, no default but a usage example of adding it via css (as you did)

I think that the user should not control on the transition at all.

libs/components/src/lib/popup/popup.scss
```diff
@@ -6,10 +6,6 @@
 	border-radius: inherit;
 	contain: layout;
 	inline-size: fit-content;
-
-	&:not(.open) {
-		display: none;
-	}
```
rinaok, line 12: @yinonov For the tooltip's transition to work, I had to remove the `display:none` from the popup. What do you think about it?

libs/components/src/lib/tooltip/README.md
yinonov: but this is a pointless button. button triggers action on click. what does this one trigger?

rinaok: @yinonov Do you want to remove this usage example?

yinonov: I think it is best

libs/components/src/lib/popup/popup.scss
yinonov: Can u elaborate?

rinaok: @yinonov If the popup has the display:none when not open. We can't see the transition.

libs/components/src/lib/tooltip/README.md
```diff
@@ -1,6 +1,6 @@
 # Tooltip
 
-A tooltip is a brief, informative message or descriptions or explanations for their paired element. Tooltips in general are less accessible so be sure to follow our [accessibility recommendation](#accessibility) in the docs.
+A tooltip displays additional, non-essential, informative message, description or explanation to a focusable element upon hover or focus. It typically appears after a short delay and disappears when `Escape` key is pressed, element looses focus or hovered off.
```
olaf-k, line 3: `element loses focus or hover state.`

libs/components/src/lib/tooltip/README.md
```diff
@@ -12,45 +12,29 @@ A tooltip is a brief, informative message or descriptions or explanations for th
 
 ### Anchor
 
-The tooltip can be placed on interactive controls (things that can be hovered or focusable) such as: button, checkbox, input text.
+The tooltip should have an `anchor` attribute with the anchor's id.
+
+The tooltip should be placed on interactive controls, elements that can be hovered over or focusable (e.g. button, checkbox, input text etc'.).
```
olaf-k, line 17: (e.g., button, checkbox, input text, etc.).

libs/components/src/lib/tooltip/tooltip.ts
```diff
@@ -16,29 +15,55 @@ export class Tooltip extends FoundationElement {
...
-	 *
-	 * @public
-	 * HTML Attribute: open
-	 */
-	@attr({
-		mode: 'boolean',
-	}) open = false;
+	constructor() {
+		super();
+		this.id = 'tooltip';
```
olaf-k, line 20: this means that several tooltips on the same page will share the same `id`!

libs/components/src/lib/popup/popup.scss
yinonov: this might affect SR reading content which is supposedly doesn't display, thus shouldn't be read by any ATs

olaf-k (changes requested):

libs/components/src/lib/popup/popup.ts
```diff
@@ -123,11 +119,11 @@ export class Popup extends FoundationElement {
 	 * @public
 	 */
 	async updatePosition() {
-		if (!this.open || !this.#anchorEl) {
+		if (!this.open || !this.anchorEl) {
```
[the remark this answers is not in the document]
rinaok: it's protected not public

libs/components/src/lib/popup/popup.scss
rinaok: I'm not sure I understood. The tooltip is visible false, isn't it affect SR reading content as well?

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -1,28 +1,32 @@
...
 export const TooltipTemplate: (
 	context: ElementDefinitionContext,
 	definition: FoundationElementDefinition
 ) => ViewTemplate<Tooltip> = () => html`
 <vwc-popup class="${getClasses}" arrow alternate
 placement=${(x) => x.placement} open=${(x) => x.open} anchor=${(x) => x.anchor}
-  exportparts="vvd-theme-alternate" role="tooltip">
-  <div class="tooltip">
+  exportparts="vvd-theme-alternate">
+  <div class="tooltip" role="tooltip" inert>
```
olaf-k, line 29: `inert` is an issue here as it "makes the browser "ignore" user input events for the element, including focus events and events from assistive technologies" ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert))
which means that `aria-describedby` will not take the inert element into account (confirmed by a quick test with NVDA and VoiceOver)

olaf-k (changes requested):
I don't think setting the `aria-describedby` dynamically to the anchor was a bad strategy, maybe just using unique ids?

libs/components/src/lib/tooltip/tooltip.template.ts
olaf-k: (this is less important, but apparently `role=tooltip` seems to be useless from what I read: screen readers just ignore it)

rinaok: @olaf-k @yinonov I removed the inert from this PR, we have a separate PR for that: https://github.com/Vonage/vivid-3/pull/827

libs/components/src/lib/tooltip/README.md
```diff
@@ -1,6 +1,6 @@
 # Tooltip
 
-A tooltip is a brief, informative message or descriptions or explanations for their paired element. Tooltips in general are less accessible so be sure to follow our [accessibility recommendation](#accessibility) in the docs.
+A tooltip displays additional, non-essential, informative message, description or explanation to a focusable element upon hover or focus. It typically appears after a short delay and disappears when `Escape` key is pressed, element loses focus or hovered off.
```
olaf-k, line 3: remove `typically` since we're talking specifically about this implementation
+
it actually only disappears on Esc when it is focused, not when it is hovered, is it the wanted behavior?

olaf-k (changes requested):

libs/components/src/lib/tooltip/README.md
olaf-k: sure, it makes sense to track the behavior there, but the current doc should reflect the current behavior

rinaok:
@olaf-k @yinonov 

What do you think? Can we merge the pr?

olaf-k:
> @olaf-k @yinonov
> 
> What do you think? Can we merge the pr?

I'm ok with the PR as it is, but we'll have to dedicate some time going over the widgets behavior when using a screen reader.
as it is (without `describedby`) the tooltip is not read when the button is focused. out of this particular scope though.

olaf-k (approved):

libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -1,28 +1,32 @@
...
+	* The template for the {@link @microsoft/fast-foundation#Tooltip} component.
+	*
+	* @param context
+	* @public
+	*/
 export const TooltipTemplate: (
 	context: ElementDefinitionContext,
 	definition: FoundationElementDefinition
 ) => ViewTemplate<Tooltip> = () => html`
 <vwc-popup class="${getClasses}" arrow alternate
```
olaf-k, line 26: better to use `context.tagFor` than hardcode references to other widgets

olaf-k (changes requested):

olaf-k (approved):

PR #830, opened 2022-11-15
[Select]: migrate new component

rachelbt:


libs/components/src/lib/select/select.template.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export const SelectTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Select> = (context: ElementDefinitionContext) => {
+	const popupTag = context.tagFor(Popup);
+
+	return html<Select>`
+	  <div class="base ${getStateClasses}"
+            aria-activedescendant="${x => x.ariaActiveDescendant}"
```
yinonov, line 76: I'm not sure this will connect with its active descendant due to shadow & light tree boundaries. that's why it was set on template (host)

yinonov (changes requested):

libs/components/src/lib/select/select.template.ts
rachelbt: So should I move it to the host?
I know that we usually prefer putting all on a wrapper div, rather on the host

yinonov: we prefer but this implementation + shadow boundaries are constraints. that actually not a 100% verified info. it should be researched whether such references work cross light & shadow trees

rinaok:
@yinonov @rachelbt 

There are several reasons why popup positioning does not work:

1. A popup looks for the anchor element using `document.getElementById(this.anchor)`. The anchor in select is inside the shadow root, so popup cannot find it.

2. As the popup is rendered first, even if you put the first option as popup's anchor, it cannot find it since it hasn't been rendered.

yinonov:
> @yinonov @rachelbt
> 
> There are several reasons why popup positioning does not work:
> 
> 1. A popup looks for the anchor element using `document.getElementById(this.anchor)`. The anchor in select is inside the shadow root, so popup cannot find it.
> 2. As the popup is rendered first, even if you put the first option as popup's anchor, it cannot find it since it hasn't been rendered.

@rinaok
1. if you recall, in the past, we talked about supporting both id string and element as anchors. can this resolve the issue? are there any better more feasible solutions? I suggest reviewing [FAST and their implementation for submenu](https://github.com/microsoft/fast/blob/6ebb5ec7bdb7666f49ddc524bc56f70a2e8aa574/packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts)

yinonov:
@rachelbt as discussed, 1 of the main issues, and a blocker, for this PR is the enablement to align with anchor's width. popup / popover native element will enable this via css but until such element is supported by all browsers, we might just add a patch to, at least, observe `clientRects` and delegate requirements to template where needed

rinaok:
> 
@yinonov We have the same issue here as in the [light dismiss PR](https://github.com/Vonage/vivid-3/pull/765#issuecomment-1336396955).

libs/components/src/lib/components.ts
```diff
@@ -31,6 +31,7 @@ export * from './progress/definition';
 export * from './progress-ring/definition';
 export * from './radio/definition';
 export * from './radio-group/definition';
+export * from './select';
```
rinaok, line 34:
```suggestion
export * from './select/definition';
```

rinaok (changes requested):
A snapshot update of accordion should be removed.

yinonov (changes requested):
1. remove accordion snapshot
2. style select to be inlined as in the #923

libs/components/src/lib/select/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## Members
+
+### Label
+
+- Type: `string` | `undefined`
+- Default: `undefined`
+
+Add a `label` attribute to add label to the Select.
```
rinaok, line 32: Should appear before the type

rinaok, line 32:
```suggestion
Add a `label` attribute to add label to the Select.

- Type: `string` | `undefined`
- Default: `undefined`
```

rachelbt:
> 1. remove accordion snapshot
> 2. style select to be inlined as in the #923

done.
Added focus but it's not behaving as it should :woman-facepalming::skin-tone-2:

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+describe('vwc-select', () => {
+	let element: Select;
+
+	global.ResizeObserver = jest.fn()
+		.mockImplementation(() => ({
+			observe: jest.fn(),
+			unobserve: jest.fn(),
+			disconnect: jest.fn()
+		}));
```
YonatanKra, line 15: This repeats in other tests. Should make it a global definition or a shared test component like `elementUpdated`. Not a must for now, but if you decide not to do it in this PR, plz open an issue for this (you can assign me ;) ).

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+		element = (await fixture(
+			`<${COMPONENT_TAG}></${COMPONENT_TAG}>`
+		)) as Select;
+	});
+
+	describe('basic', () => {
+		it('should be initialized as a vwc-select', async () => {
+			expect(element).toBeInstanceOf(Select);
+			expect(element.open).toEqual(false);
+			expect(element.disabled).toEqual(false);
```
YonatanKra, line 27: What's the default value of label? appearance? shape?

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+	});
+
+	describe('label', function () {
+		it('should set a select label if label is set', async function () {
+			const labelText = 'label';
+			element.label = labelText;
+			await elementUpdated(element);
+			const labelElement = element.shadowRoot?.querySelector('label');
+			expect(labelElement)
+				.toBeTruthy();
```
YonatanKra, line 38: Not needed. The next expect already implicitly checks this.

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+			expect(labelElement?.textContent?.trim())
+				.toEqual(labelText);
+		});
+
+		it('should show label only if label is set', async function () {
```
YonatanKra, line 43:
```suggestion
		it('should show label element only if label is set', async function () {
```

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+				.toBeNull();
+		});
+	});
+
+	describe('disabled', function () {
+		it('should set disabled class for select  when disabled is true', async () => {
+			expect(element.shadowRoot?.querySelector('.disabled')).toBeFalsy();
+			element.toggleAttribute('disabled', true);
+			await elementUpdated(element);
+			expect(element.shadowRoot?.querySelector('.disabled')).toBeTruthy();
```
YonatanKra, line 55: Expectations in the end of the use-case:
```suggestion
      const disableClassExistsWithDisabledFalse = Boolean(element.shadowRoot?.querySelector('.disabled'));
      
			element.toggleAttribute('disabled', true);
			await elementUpdated(element);
			
			expect(disableClassExistsWithDisabledFalse).toBeFalsy();
			expect(element.shadowRoot?.querySelector('.disabled')).toBeTruthy();
```

libs/components/src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+		});
+	});
+
+	describe('open', function () {
+		it('should set open when clicked', async () => {
+			expect(element.open).toEqual(false);
```
YonatanKra, line 61: See my comment for disabled.

libs/components/src/lib/select/select.template.ts
YonatanKra: It doesn't cross the shadow. I checked it in the `data-grid`. That's why I reverted to `template`.

YonatanKra (changes requested):

libs/components/src/lib/select/select.spec.ts
rachelbt: label is `undefined` unless consumer added string.
appearance = `fieldset`
shape = `rounded`
do I need to add them?
If yes - where? :)

libs/components/src/lib/select/select.spec.ts
rachelbt: looked at it - not sure what you mean... :)

libs/components/src/lib/select/select.template.ts
rachelbt: @YonatanKra @yinonov I changed .base to be the host.
Not sure about it because this is usually not what we do in our component.
Having problem with the open test & focus :-(
Appreciate you help (again...)

nx.json
```diff
@@ -1,55 +1,55 @@
 {
```
yinonov, line 1: revert?

tsconfig.base.json
```diff
@@ -1,26 +1,26 @@
 {
```
yinonov, line 1: revert?

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -427,7 +427,7 @@ describe('vwc-text-field', () => {
 				textContent?.trim()).toEqual(errorMessageTwo);
```
yinonov, line 427: relate?

yinonov (changes requested):

rinaok (changes requested):
Screenshot 2023-01-08 at 12 59 40

The code blocks don't look good

libs/components/src/lib/select/README.md
```diff
@@ -0,0 +1,108 @@
...
+  <vwc-option value="2" text="Option 2"></vwc-option>
+  <vwc-option value="3" text="Option 3"></vwc-option>
+</vwc-select>
+```
+
+
+## Caveat
+
+Document elements display precedence is formed by the imaginary z-axis [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), commonly by order of which elements are rendered and special properties (e.g. _z-index_).
+Combobox component is a low level element, unaware of its document context, but is, in most cases, required to overlay on top of all elements.
```
yinonov, line 108:
```suggestion
Select component is a low level element, unaware of its document context, but is, in most cases, required to overlay on top of all elements.
```

yinonov: isn't it missing the rest of the explanation? readers would ask "so what should we do then?"

yinonov (changes requested):

yinonov (commented):
noticed a layout shift caused by the change of the select value. toggle between the options to experience it. should this be resolved in this scope?

rachelbt:
> lib/text-field/text-field.spec.ts

fixed :)

rachelbt:
> noticed a layout shift caused by the change of the select value. toggle between the options to experience it. should this be resolved in this scope?

that happened because there was no enough space in the iframe. I set it to be bigger than 150px;

libs/components/src/lib/select/README.md
rachelbt: it exactly what is in combobox but you're right - I added an explanation line of what to do if

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -427,7 +427,7 @@ describe('vwc-text-field', () => {
...
 		});
 
-		it('should change the error message only when not in focus', async function() {
+		it('should change the error message only when already not valid', async function() {
```
yinonov, line 430: this too

yinonov (changes requested):

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -458,7 +458,7 @@ describe('vwc-text-field', () => {
 				.toBeNull();
 		});
 
-		it('should not show error message when success is shown', async function () {
+		it('should remove error message when success is shown', async function () {
```
rachelbt, line 461:
```suggestion
		it('should not show error message when success is shown', async function () {
```

yinonov:
> > noticed a layout shift caused by the change of the select value. toggle between the options to experience it. should this be resolved in this scope?
> 
> that happened because there was no enough space in the iframe. I set it to be bigger than 150px;

I mean the inline size of the selectmenu changes according to its changed value

rachelbt:
> > > noticed a layout shift caused by the change of the select value. toggle between the options to experience it. should this be resolved in this scope?
> > 
> > 
> > that happened because there was no enough space in the iframe. I set it to be bigger than 150px;
> 
> I mean the inline size of the selectmenu changes according to its changed value

this is because we dont have a min-width or fixed width to the select. 
Should we have one? should we leave it to our consumes? 
Material sets a min-width

yinonov:
> > > > noticed a layout shift caused by the change of the select value. toggle between the options to experience it. should this be resolved in this scope?
> 
> > > 
> 
> > > 
> 
> > > that happened because there was no enough space in the iframe. I set it to be bigger than 150px;
> 
> > 
> 
> > I mean the inline size of the selectmenu changes according to its changed value
> 
> 
> 
> this is because we dont have a min-width or fixed width to the select. 
> 
> Should we have one? should we leave it to our consumes? 
> 
> Material sets a min-width

How does native select / selectmenu is expected to behave?

libs/components/src/lib/select/select.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type SelectAppearance = Extract<Appearance, Appearance.Outlined | Appearance.Ghost>;
+type SelectShape = Extract<Shape, Shape.Rounded | Shape.Pill>;
+
+/**
+ * Base class for select
+ *
+ * @public
+ */
+@formElements
```
olaf-k, line 17: I'm assuming the only thing we use from this is the validation, but then I have 2 questions:
1. what kind of validation? the select always seems to have a value, so `required` seems strange.
2. why is `error-text` not displayed in the template?

yinonov: 1. it doesn't necessarily initiated with a value set
2. it should. are we not implementing the [form-elements pattern](https://github.com/Vonage/vivid-3/blob/ea937655378e4a338f6be3121888d92a6505af8e/libs/components/src/shared/patterns/form-elements/form-elements.ts#L1)?

olaf-k: 1. the select can be empty indeed, but then a `required` validation doesn't make sense:
image
2. we inherit from `FASTSelect` which already is a `FormAssociatedSelect`, so I'm not really clear about why we use `@formElements` which seems to bring a lot of unused stuff?

rachelbt: @YonatanKra ?

libs/components/src/lib/select/select.template.ts
```diff
@@ -0,0 +1,120 @@
...
+}: Select) => classNames(
+	['disabled', disabled],
+	[`appearance-${appearance}`, Boolean(appearance)],
+	[`shape-${shape}`, Boolean(shape)],
+);
+
+/**
+ *
+ */
+function renderLabel() {
```
olaf-k, line 25: any reason for extracting this to a function?

libs/components/src/lib/select/select.ts
```diff
@@ -0,0 +1,44 @@
...
+export class Select extends FoundationSelect {
+	/**
+	 * Indicates the select's label.
+	 *
+	 * @public
+	 * @remarks
+	 * HTML Attribute: label
+	 */
```
olaf-k, line 25: copy/paste: update or remove

libs/components/src/lib/select/select.ts
```diff
@@ -0,0 +1,44 @@
...
+	@attr appearance?: SelectAppearance;
+	@attr shape?: SelectShape;
+
+	/**
+	 *
+	 * HTML Attribute: string
+	 */
```
olaf-k, line 32: copy/paste: update or remove

olaf-k (changes requested):

olaf-k (dismissed):

libs/components/src/lib/select/select.ts
YonatanKra: We need to look at it. Do we really need it?
The `formElement` decorator adds a lot of validation stuff to an element.
Its built with the text area and field in mind and can be refactored to be further modularized.

The validation @olaf-k presents in this pic is the native validation. Required should not be active unless we add `required` to the list. Sometimes, select is a required field. Does this happen even if no `required` is added? If so, it is a bug in how the field works.

YonatanKra: @olaf-k 
I do not understand your objection. Is there anything wrong with the select's behavior?

olaf-k: > @olaf-k I do not understand your objection. Is there anything wrong with the select's behavior?

there are two things:
1. I'm asking what kind of validation we expect from a component that always has a value. `required` doesn't make sense as I illustrated, so what else?
2. my code review input is: what is the decorator for? if it is not used, it should be removed.

> The formElement decorator adds a lot of validation stuff to an element.

precisely. none of its helpers are used. it seems to *only* add `label` and I'm sure we don't need it just for that. so back to 1: what kind of validation do we expect?

YonatanKra: Actually, one can null the value of select and thus we'd expect required to work.
Native select works with require (and other validation functionality). IMO it would be in the spirit of consistency to follow the native spec here too, no?
image

olaf-k: the native element offers features ours doesn't, so consistency is not the prevalent point here.
the nullable value is a valid use case though and that answers 1, but leaves 2 unanswered.

YonatanKra: It mostly solves this: https://github.com/microsoft/fast/issues/5963

olaf-k: right. and since none of the error/success helpers are used in select, my point is that it should be removed. (+ it also adds a `charCount` prop that doesn't fit).

libs/components/src/lib/select/select.ts
```diff
@@ -0,0 +1,44 @@
+import { applyMixins, Select as FoundationSelect } from '@microsoft/fast-foundation';
+import { attr } from '@microsoft/fast-element';
+import type { Popup } from '../popup/popup';
+import { FormElement, formElements } from '../../shared/patterns';
+import { AffixIcon } from '../../shared/patterns';
+import type { Appearance, Shape } from '../enums';
+
+
+type SelectAppearance = Extract<Appearance, Appearance.Outlined | Appearance.Ghost>;
+type SelectShape = Extract<Shape, Shape.Rounded | Shape.Pill>;
```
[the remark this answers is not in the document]
yinonov: what's the difference in result? from a readability POV it's very explicit, subset portion of... instead of using `or` which can lead to arbitrary content. it is also already the convention we practiced so that's another readability concern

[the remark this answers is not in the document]
olaf-k:
```
type SelectAppearance = Extract<Appearance, Appearance.Outlined | Appearance.Ghost>;
```
is less readable than
```
type SelectAppearance = Appearance.Outlined | Appearance.Ghost;
```
because it makes the reader wonder what Extract does, find out what it does, realize it's not needed and wonder why it is used. The second is idiomatic TS and can be read with minimum TS knowledge.
I can open a PR to refactor the rest of the code, no worries.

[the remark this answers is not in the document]
yinonov: I disagree. it's anyway breaking convention so I would pick on that here, but rather in a short [discussion](https://github.com/Vonage/vivid-3/discussions)?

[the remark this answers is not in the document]
rachelbt: done: https://github.com/Vonage/vivid-3/discussions/961

yinonov (approved):

yinonov (approved):

YonatanKra (approved):

PR #832, opened 2022-11-15
fix(side-drawer): set inline-size with css var

rachelbt:


rachelbt:
After talking to @yinonov  - the css variable is for other use and the content need to be adaptive to its width.

PR #836, opened 2022-11-16
refactor(styles): unset root's font size to resolve to 16

yinonov:


libs/styles/src/lib/core/typography.scss
```diff
@@ -10,14 +10,22 @@ $selectors: (
...
+
+	&:not(:scope) {
+		font: var(#{constants.$vvd-typography-base});
 	}
-	--vvd-size-font-scale-base: 1rem;
+
 	font-feature-settings: 'kern';       /* turns on kerning */
 	text-rendering: optimizeLegibility;  /* emphasise on legibility when rendering, turns on ligatures and kerning */
 	-webkit-font-smoothing: antialiased; /* apply font anti-aliasing for Webkit on OSX */
 	-moz-osx-font-smoothing: grayscale;  /* apply font anti-aliasing for Firefox on OSX */
```
rachelbt, line 28: why not set these properties on body like the font itself and not on the HTML?

libs/styles/src/lib/core/typography.scss
```diff
@@ -8,12 +8,27 @@ $selectors: (
...
-	font: var(#{constants.$vvd-typography-base});
+}
+
+.vvd-root {
+	&:root { // <--- When 'vvd-root' is set on the `:root` (html element), typeface sizes are able to descend from the root font-size, thus comply with the [WCAG 1.4.4](https://www.w3.org/WAI/WCAG21/Understanding/resize-text)
+		--vvd-size-font-scale-base: 1rem;
+		font-size: unset; // resolve to 16px
```
olaf-k, line 22: `resolve to the user-agent default font size (generally 16px)`
(yes I'm reviewing comments I know)

rachelbt (approved):
LGTM

libs/styles/src/lib/core/typography.scss
yinonov:
```suggestion
		font-size: unset; // resolve to the user-agent default font size (generally 16px)
```

olaf-k (approved):

PR #837, opened 2022-11-16
perf(components): action group role as group (VIV-897)

yinonov:
see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role


rachelbt:
_The group role should be used to form a logical collection of items with related functionality_

it seems right to add the role="group" but it may be problematic since we do not know what our consumers are using the action group for.

however we have this:
_The group role should not be used for major perceivable sections of a page. If a section is significant enough that it should be included in the page's table of contents, use the region role or a standard [landmark role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#landmark_roles)._

I'm wondering whether we should add it in the component or suggest adding it with a usage example in the docs.

yinonov:
> _The group role should be used to form a logical collection of items with related functionality_
> 
> it seems right to add the role="group" but it may be problematic since we do not know what our consumers are using the action group for.
> 
> however we have this: _The group role should not be used for major perceivable sections of a page. If a section is significant enough that it should be included in the page's table of contents, use the region role or a standard [landmark role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#landmark_roles)._
> 
> I'm wondering whether we should add it in the component or suggest adding it with a usage example in the docs.

it can be set by default and be opt-out or even allow author-custom role (where future toolbar might require). it should be documented

libs/components/src/lib/action-group/README.md
```diff
@@ -132,18 +132,10 @@ Use `<vwc-divider>` for adding separator between the action elements
...
- }
-</style>
-
-<vwc-action-group>
-  <div role="radiogroup" aria-label="Text Alignment">
-   <vwc-button type="button" role="radio" icon="align-left-line" aria-checked="false" tabindex="0" aria-label="Text Align Left"></vwc-button>
-   <vwc-button type="button" role="radio" icon="align-center-line" aria-checked="true" tabindex="-1" aria-label="Text Align Center" appearance="filled"></vwc-button>
-   <vwc-button type="button" role="radio" icon="align-right-line" aria-checked="false" tabindex="-1" aria-label="Text Align Right"></vwc-button>
-  </div>
+<vwc-action-group role="radiogroup" aria-label="Text Alignment">
```
rachelbt, line 135: 👍

libs/components/src/lib/action-group/action-group.ts
```diff
@@ -59,6 +59,7 @@ export class ActionGroup extends FoundationElement {
 		mode: 'boolean',
 	}) tight = false;
 
+	@attr() role: string | undefined;
```
rachelbt, line 62: is there a way to restrict it to rote types?

rachelbt (commented):
look look great!
Left on comment

libs/components/src/lib/action-group/action-group.template.ts
```diff
@@ -28,6 +28,8 @@ export const ActionGroupTemplate: (
 	definition: FoundationElementDefinition
 
 ) => ViewTemplate<ActionGroup> = () => html`
-<div class="${getClasses}">
+<div class="${getClasses}" 
+	role="${x => x.role ? x.role : 'group'}" 
+	aria-label="${x => x.ariaLabel}">
```
olaf-k, line 33: if you want the `aria-label` to be updated when it changes on the host, you need to declare it as an `@attr` otherwise it'll just keep the initial value.

olaf-k (changes requested):

olaf-k (approved):

Issue #838, opened 2022-11-16
[toggletip]: create a new component


yinonov:
### The problem

we’re often asked for vivid tooltip to support interactive content (e.g. links).
The reason tooltip hinders any content other than plain text is to improve a11y concerns which are often overlooked in development, and tooltip in that sense require a little chat every time it's raised.
a tooltip - toggletip comparison [nice summary](https://web.dev/building-a-tooltip-component/#toggletip-vs-tooltip) 

### The solution

create a dedicated component for toggletip and reduce the wrong use of tooltip

### Application/website

null

### Business priority

Medium Priority = upcoming release but is not pressing

### Examples

https://vonage.slack.com/archives/C013F0YKH99/p1667547664898289

following thread have some helpful information
https://web-a11y.slack.com/archives/C042TSFGN/p1668458681197569

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/vonage/vivid-3/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/vonage/vivid-3/issues) for duplicate issues

rinaok:
https://jira.vonage.com/browse/VIV-865

PR #843, opened 2022-11-17
feat(components): toggletip

olaf-k:
working proto, takes care of #838

needs more work +
- [ ] a11y (aria-live)
- [ ] tests
- [ ] changes made to `popup` should be moved to their own PR

Open questions:
- do `disabled` and `connotation` make sense?
- atm, the toggletip container displays pretty much everything you throw at it, should it be constrained?
- button size condensed or normal by default?


yinonov:
@rinaok - seems like @olaf-k implemented basic light dismiss in this one. you should probably exchange information

olaf-k:
> @rinaok - seems like @olaf-k implemented basic light dismiss in this one. you should probably exchange information
> 
> related #765

"implemented" is flattering :D it's more of a quick hack for the proto to look better. I'll have a look at what @rinaok did!

libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @public
+ */
+export const ToggletipTemplate:
+	(context: ElementDefinitionContext, definition: FoundationElementDefinition) => ViewTemplate<Toggletip> = (context: ElementDefinitionContext) => {
+		const button = context.tagFor(Button);
+		const popup = context.tagFor(Popup);
+		// const layout = context.tagFor(Layout);
+
+		return html`
+			<span class="${getClasses}">
```
rachelbt, line 25:
```suggestion
			<div class="${getClasses}">
```

libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -0,0 +1,47 @@
...
+				<${popup}
+					arrow
+					placement="${x => x.placement ? x.placement : 'right'}"
+					${ref('popup')}
+				>
+					<span style="padding: 12px;">
+						<slot></slot>
+					</span>
+				</${popup}>
+			</span>
```
rachelbt, line 45:
```suggestion
			</div>
```

libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const ToggletipTemplate:
+	(context: ElementDefinitionContext, definition: FoundationElementDefinition) => ViewTemplate<Toggletip> = (context: ElementDefinitionContext) => {
+		const button = context.tagFor(Button);
+		const popup = context.tagFor(Popup);
+		// const layout = context.tagFor(Layout);
+
+		return html`
+			<span class="${getClasses}">
+				<${button}
```
rachelbt, line 26: I'm not sure that the toggle-tip template should contain the button inside. It makes the use too complicated to my opinion. shape, size, appearance are being set but applied only to the button.
Why cant we do the toggletip like tooltip and in the docs have an example for usage with button?

libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -0,0 +1,47 @@
...
+					icon="${x => x.icon ? x.icon : 'info-line'}"
+					${ref('button')}
+				>
+				</${button}>
+				<${popup}
+					arrow
+					placement="${x => x.placement ? x.placement : 'right'}"
+					${ref('popup')}
+				>
+					<span style="padding: 12px;">
```
rachelbt, line 41:
```suggestion
					<span>
```
style="padding: 12px;" - why not in the CSS? :)

rachelbt (commented):
Looks like a good start of the toggletip :)
- what about close button adding it with member same as modal?
- Do we have a heading alternative via member (again - same as modal)? 
- I'm wondering also whether we have one slot or are we dividing it ti content + footer, and a content slot to override our opinionated structure.

rinaok:
> > @rinaok - seems like @olaf-k implemented basic light dismiss in this one. you should probably exchange information
> > related #765
> 
> "implemented" is flattering :D it's more of a quick hack for the proto to look better. I'll have a look at what @rinaok did!

This is the light-dismiss pr, still in progress and we still have many question https://github.com/Vonage/vivid-3/pull/765.

rinaok (commented):
@olaf-k @yinonov 

What about an example of the dismissible attribute?
When you add dismissible to popup you have a dismiss button:

Screenshot 2022-11-20 at 14 16 17

libs/components/src/lib/popup/popup.ts
```diff
@@ -15,14 +15,14 @@ export class Popup extends FoundationElement {
 	get #distance(): number { return 12; }
 	get #strategy(): Strategy { return 'fixed'; }
 	get #middleware(): Array<any> {
-		const middleware = [flip(), hide(), inline()];
+		const middleware = [inline(), flip(), hide()];
```
rinaok, line 18: Why the order matters?

libs/components/src/lib/popup/popup.ts
```diff
@@ -165,7 +165,7 @@ export class Popup extends FoundationElement {
 		return document.getElementById(this.anchor);
 	}
 
-	#handleKeydown = (event: Event) => {
+	handleKeydown = (event: Event) => {
```
rinaok, line 168: handleKeydown should remain private. you can add a `hide` method

yinonov: if it's referenced from template, it must be public or at least, maybe, protected

libs/components/src/lib/toggletip/README.md
```diff
@@ -0,0 +1,128 @@
+# toggletip
```
rinaok, line 1:
```suggestion
# Toggletip
```

libs/components/src/lib/toggletip/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+Represents a toggletip custom element.
+
+```js
+<script type="module">import '@vonage/vivid/toggletip';</script>
```
rinaok, line 6:
```suggestion
<script type="module">
	import '@vonage/vivid/toggletip';
</script>
```

libs/components/src/lib/toggletip/README.md
```diff
@@ -0,0 +1,128 @@
...
+- Default: `'accent'`
+
+#### Ghost toggletip with connotation
+
+```html preview
+<p><vwc-toggletip connotation='accent'>more information</vwc-toggletip></p>
+<p><vwc-toggletip connotation='cta'>more information</vwc-toggletip></p>
+<p><vwc-toggletip connotation='success'>more information</vwc-toggletip></p>
+<p><vwc-toggletip connotation='alert'>more information</vwc-toggletip></p>
+```
```
rinaok, line 97: Can you use the <vwc-layout> component instead of <p>?

rinaok:
> @olaf-k @yinonov
> 
> What about an example of the dismissible attribute? When you add dismissible to popup you have a dismiss button:
> 
> Screenshot 2022-11-20 at 14 16 17

Also missing the `alternate` attribute example.

libs/components/src/lib/popup/popup.ts
rinaok: @yinonov it's not referenced from template here. It's referenced from toggletip.ts
In tooltip [PR](https://github.com/Vonage/vivid-3/pull/829), the tooltip extends the popup, so it has all the methods.
I made the anchorEl be protected instead of private. Why not do the same here?

rinaok:
@olaf-k @yinonov @AyalaBu If the user wants a different button, what should they do? This is not the most optimal implementation of the toggletip, in my opinion.

AyalaBu:
@rinaok 
Information gathering is in process

libs/components/src/lib/toggletip/toggletip.template.ts
olaf-k: it's a work in progress rhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

libs/components/src/lib/popup/popup.ts
olaf-k: https://floating-ui.com/docs/middleware#ordering

image
image

olaf-k:
group answer:

- **it is missing the alternate look**
fair enough, I'll add it.

- **what should be the content API?**
I don't know! I'm pushing everything the user throws at the widget inside a stupid span because that is easy and flexible. I'm OK with a more constraining layout but it would have to be defined first.

- **close button/dismissible**
the first iteration actually used dismissible but I realized most toggletip examples don't have a close button and rely on light dismiss (see [here](https://ltdesign.azurewebsites.net/pattern-library/messaging/toggletips.html), [here](https://carbondesignsystem.com/components/toggletip/usage) or [here](https://inclusive-components.design/tooltips-toggletips/) for instance). same as above: there's no specs atm (but I know @AyalaBu is looking into it).

- **extracting the button part**
we don't need a component to mimic the toggletip behavior, [as shown in this CodePen](https://codepen.io/olaf-k/pen/eYKGJxg). there are two ideas behind this PR though:
1. provide an opiniated component that looks and behaves consistently across applications.
2. take care of internal details, from basic event handling to more complex a11y attributes (this is probably an ARIA live region and users don't want to deal with that).

AyalaBu:
I created a toggletip component example on this branch:
https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/branch/C9VZeRowJGzDiC4NLFezTn/Vivid-UI-Kit---3.0-WIP?node-id=20623%3A129100&t=y8uM758UHHYz7hAu-1

It is mostly based on Carbon example:
https://carbondesignsystem.com/components/toggletip/usage/

After some thought I think it will be more correct to go for the light dismiss,
since toggletip is an actionable component by its core, perhaps there is no need to oversaturate it with different kinds of ctas.

@yinonov @YonatanKra @rinaok @rachelbt @olaf-k 
So what are your thought about this?

olaf-k:
> I created a toggletip component example on this branch:

Thanks @AyalaBu !

> After some thought I think it will be more correct to go for the light dismiss, since toggletip is an actionable component by its core, perhaps there is no need to oversaturate it with different kinds of ctas.
> (...) 
> @yinonov @YonatanKra @rinaok @rachelbt @olaf-k So what are your thought about this?

I agree with the light dismiss, which is the way all toggletips seem to work.
Question about your design: does it mean there should always be two buttons present at the bottom? Shouldn't they be optional and give the author the possibility to have none, one or two?

As far as design systems go, here are two more I had a look at for this PR, aside from Carbon you already mentioned:
https://ltdesign.azurewebsites.net/pattern-library/messaging/toggletips.html
https://inclusive-components.design/tooltips-toggletips/

AyalaBu:
@olaf-k 
If there are no buttons present at all then it is called tooltip (based on my research), which is not exactly the same.

Aside of that, there are two ctas present in the design in order to give designers the options to have two actions.
Toggletip absolutely doesn't have to include two actions at all.

acelletti:
Closing this for now, created a ticket to review [VIV-865 - Investigate Toggletip](https://jira.vonage.com/browse/VIV-865)

PR #847, opened 2022-11-18
ci(publish): actions/checkout resolve to default github.token

yinonov:
actions/checkout@v3 [by default uses ${{ github.token }}](https://github.com/actions/checkout#usage).
as all other _checkout_ steps in our workflows use the default token, is there any specific reason this action provides one explicitly?
providing the token anyway failed in [that actions](https://github.com/Vonage/vivid-3/actions/runs/3496963659)

is this paramater redundant here?


olaf-k (approved):
I'll leave the last word to @YonatanKra but it looks like the extra permissions of `VNG_VVD_PAT` are indeed not necessary here

package.json
```diff
@@ -54,7 +54,7 @@
 		"@typescript-eslint/eslint-plugin": "5.41.0",
 		"@typescript-eslint/parser": "5.41.0",
 		"@typescript-eslint/utils": "^5.36.1",
-		"@vonage/vivid-figma-tokens": "github:vonage/vivid-figma-tokens#v0.2.0",
+		"@vonage/vivid-figma-tokens": "github:vonage/vivid-figma-tokens#v0.1.1",
```
yinonov, line 57:
```suggestion
		"@vonage/vivid-figma-tokens": "github:vonage/vivid-figma-tokens#v0.2.0",
```

.github/workflows/_publish-package.yml
```diff
@@ -1,4 +1,4 @@
-name: '📦 Publish Vivid'
+name: "📦 Publish Vivid"
```
yinonov, line 1:
```suggestion
name: '📦 Publish Vivid'
```

YonatanKra:
They are necessary for this step:
```
 - name: Bump and Changelog
        id: bump
        run: |
          npx beachball bump -m "Publishing a new version" --no-push -y > tmp.txt
          echo "::set-output name=version_bumped::$(if grep -q "\- @vonage-vivid-" "./tmp.txt"; then echo true; else echo false; fi)"
          cat tmp.txt
          cat libs/components/package.json
          git add .
          git commit -m "Version Bump"
          git push
```
Otherwise, you cannot commit to the `main` branch.
Have you checked that?

yinonov:
> They are necessary for this step:
> 
> ```
>  - name: Bump and Changelog
>         id: bump
>         run: |
>           npx beachball bump -m "Publishing a new version" --no-push -y > tmp.txt
>           echo "::set-output name=version_bumped::$(if grep -q "\- @vonage-vivid-" "./tmp.txt"; then echo true; else echo false; fi)"
>           cat tmp.txt
>           cat libs/components/package.json
>           git add .
>           git commit -m "Version Bump"
>           git push
> ```
> 
> Otherwise, you cannot commit to the `main` branch. Have you checked that?

no. isn't it risky to check? not sure dry-run are available. anyway, I now understand what that token is for. however, it does seem like [GITHUB_TOKEN permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) can be set.
will close if it makes no difference

PR #850, opened 2022-11-20
refactor(components): style gutters and gaps in child items

yinonov:
align gutters and gaps to vivid-2


olaf-k (commented):
it's a detail but it might be worth increasing the viewport width for the nav-disclosure ui tests.

rachelbt (approved):
LGTM
In nav-item - although they are not interfering with design - we should consider removing the inline-padding from icon-only option.

yinonov:
> it's a detail but it might be worth increasing the viewport width for the nav-disclosure ui tests.

it doesn't trim anything. so why?

rachelbt:
> > it's a detail but it might be worth increasing the viewport width for the nav-disclosure ui tests.
> 
> it doesn't trim anything. so why?

because it doesn't do anything :)

olaf-k (approved):

olaf-k:
> > > it's a detail but it might be worth increasing the viewport width for the nav-disclosure ui tests.
> > 
> > 
> > it doesn't trim anything. so why?
> 
> because it doesn't do anything :)

voilà ;)

it just makes it easier to spot differences when the text is not split, but as a I said, it's a detail, feel free to ignore, I won't be offended.

yinonov:
> > > > it's a detail but it might be worth increasing the viewport width for the nav-disclosure ui tests.
> > > 
> > > 
> > > it doesn't trim anything. so why?
> > 
> > 
> > because it doesn't do anything :)
> 
> voilà ;)
> 
> it just makes it easier to spot differences when the text is not split, but as a I said, it's a detail, feel free to ignore, I won't be offended.

I think this is actually a desired visual regression test. lines are designed to break in nav item / disclosure. might require a bit of block padding but that's another scope of iterating design, unrelated to this PR

PR #852, opened 2022-11-21
refactor(components): warning connotation design patching

### LINKED ISSUE #755: [Warning connotation token change]: changing the token in warning elements
Components that have filled appearance like buttons, badges, banners etc. have a color token 500 in all the connotations.

This presents a problem with the warning connotation, since the yellow turns to orange mustard color.
It is a known practice to treat the warning connotation differently, as explained in this article:

https://uxdesign.cc/the-dark-yellow-problem-in-design-system-color-palettes-a0db1eedc99d

In the Vivid 3 documentation the warning component receives the same token number as any other connotation,
while in the original design (Vivi UI Kit 2.0 Figma file) it is 300, and was originally designed this way.

There came a request to change the 300 to 200, but since in the documentation it has a 500 token, this case needs special treatment.

yinonov:
Closes #755


rachelbt (commented):
- not sure that it's relevant to this PR but the note background color is not aligned to the one in the Figma - i needs to be lighter
- in dark mode the banner looks a bit too brownish - @AyalaBu  is it suppose to be like this?

AyalaBu:
@rachelbt 
I'm not a 100% sure because I don't see the code,
but it seems that in the note (new one with soft background and an outline)
bg color is 100, while it's supposed to be 50.

yinonov:
> @rachelbt I'm not a 100% sure because I don't see the code, but it seems that in the note (new one with soft background and an outline) bg color is 100, while it's supposed to be 50.

changed to 50

rachelbt (approved):
LGTM

PR #862, opened 2022-11-25
fix(components): remove hardcoded tags

olaf-k:
necessary if we want to be prefix-agnostic


yinonov:
Bravo!

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -56,8 +58,8 @@ const renderHeaderButton = (context: ElementDefinitionContext) => {
 		${when(x => x.meta, html`<span class="meta">${x => x.meta}</span>`)}
 		<span class="indicator">
 			${when(x => !x.noIndicator && !x.iconTrailing, html`
-				${when(x => !x.open, html`<vwc-icon type='chevron-down-solid'></vwc-icon>`)}
-				${when(x => x.open, html`<vwc-icon type='chevron-up-solid'></vwc-icon>`)}
+				${when(x => !x.open, html`<${iconTag} type='chevron-down-solid'></${iconTag}>`)}
```
yinonov, line 61: Might as well replace with the `affix` pattern

yinonov (changes requested):

olaf-k:
TODO: address CSS too
image

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -86,8 +86,7 @@
 
 	.indicator {
 		display: flex;
-		vwc-icon,
-		slot[name="indicator"i]::slotted(vwc-icon) {
+		:first-child {
 			margin-inline-start: 16px;
```
rachelbt, line 90: we actually do not need this at all - this is not a slot... 
we can keep only this:
`margin-inline-start: 16px;`
no need the `display: flex` at all :)

rachelbt (commented):
👍

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -46,20 +46,22 @@ const renderHeaderButton = (context: ElementDefinitionContext) => {
...
 		${x => affixIconTemplate(x.icon)}
 		<span class="heading-text">${x => x.heading}</span>
 		${when(x => x.meta, html`<span class="meta">${x => x.meta}</span>`)}
-		<span class="indicator">
-			${when(x => !x.noIndicator && !x.iconTrailing, html`
-				${when(x => !x.open, html`<vwc-icon name='chevron-down-solid'></vwc-icon>`)}
-				${when(x => x.open, html`<vwc-icon name='chevron-up-solid'></vwc-icon>`)}
-			`)}
-		</span>
+
```
yinonov, line 58: reason for 2 elements is to eagerly load both icons. even if 1 isn't used initially

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -46,20 +46,22 @@ const renderHeaderButton = (context: ElementDefinitionContext) => {
 	const affixIconTemplate = affixIconTemplateFactory(context);
 	const focusTemplate = focusTemplateFactory(context);
 
+	/* eslint-disable @typescript-eslint/indent */
```
yinonov, line 49: why?

libs/components/src/lib/accordion/accordion.scss
```diff
@@ -5,6 +5,6 @@
 	flex-direction: column;
 }
 
-::slotted(vwc-accordion-item:not(:only-of-type)) {
+::slotted(:not(:only-of-type)) {
```
yinonov, line 8: hmm anything then? should we create an issue to propagate role to host as in other widgets?

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.template.ts
olaf-k: because eslint's idea of how the template should be indented makes it less readable.

libs/components/src/lib/accordion/accordion.scss
olaf-k: yes, I thought about it, but accordion-items don't have particular roles. ours are buttons, but I tried to stay generic.
I'm open to suggestions!

libs/components/src/lib/accordion-item/accordion-item.template.ts
yinonov: this is a local patch and problem will persist in other templates. maybe less relevant to this scope?

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.template.ts
yinonov: `<vwc-icon name="chevron-up-solid">` already fetches the svg file from CDN

yinonov (changes requested):

olaf-k:
@yinonov eslint exception removed + svg preloading in #909

yinonov (approved):

PR #867, opened 2022-11-28
feat(components): icon enable svg nodes assignment to default slot

yinonov:
BREAKING CHANGE: direct assign to svg property is now discouraged


apps/docs/_data/components.json
```diff
@@ -280,7 +280,7 @@
 		]
 	},
 	{
-		"title": "option",
+		"title": "listbox-option",
```
rinaok, line 283: How does it relate to this PR?

yinonov: it doesn't. caught red handed
```suggestion
		"title": "option",
```

rachelbt (approved):
LGTM.
Make so much more sense to use it as a slot

PR #880, opened 2022-12-01
[DO NOT MERGE] feeding ESM to Jest

olaf-k:
[The PR to allow custom prefix](https://github.com/Vonage/vivid-3/pull/861) for the Vivid elements relies on `meta.import.url` to propagate said prefix. As this is an ESM feature and Jest works with CJS modules, the PR mocks `meta.import` using [`ts-jest-mock-import-meta`](https://github.com/ThomZz/ts-jest-mock-import-meta).

This PR is an attempt at feeding Jest ESM directly. This has been done by updating the config files, using Jest native ESM support and updating a few things in the tests.

### Config files

- Removed `module: commonjs` from `tsconfig.spec.json` to output ES modules.
- Updated `jest.config.cjs` to use the [`ts-jest/presets/default-esm` preset](https://kulshekhar.github.io/ts-jest/docs/getting-started/presets).
- Fine-tuning: ignore CSS styles files imported in components with [`jest-transform-stub`](https://github.com/eddyerburgh/jest-transform-stub), make sure `@vivid-nx/shared` is found using `moduleNameMapper`.

### Jest ESM

[As explained here](https://jestjs.io/docs/ecmascript-modules), the support is currently experimental and only works when the proper `--experimental-vm-modules` node flag is used. Do not forget it or nothing will work.

### Updating the tests

- For tests that use them, Jest globals need to be explicitly imported with `import {jest} from '@jest/globals';` (could also have used the jest object in `import.meta`).
- The icon test uses `require('promise')` and several tests use `jest-axe` [which is only available as a CJS module](https://github.com/NickColley/jest-axe/issues/230). These tests were patched using the method documented in [Jest's site](https://jestjs.io/docs/ecmascript-modules) (i.e., by [importing node's require](https://nodejs.org/api/module.html#modulecreaterequirefilename)).
- The rest are just tiny irrelevant TS fixes to make ts-jest stop complaining.

### Remaining issues

- The dialog test has been excluded from the suite because Jest insists on treating the CSS from the polyfill it uses as a module. I didn't investigate further. Should be fixable given more time? (And why doesn't it fall into `jest-transform-stub`?)
- The popup test imports the floating-ui ES module and uses Jest to mock part of it, but since ESM are immutable this raises an error. This should be [fixable using a dynamic import apparently](https://github.com/facebook/jest/issues/12145) but I haven't tested.
- More annoyingly, the `import.meta.url` does not seem to carry over query parameters? See the breadcrumb test which logs the url without params when called with `import './index?prefix=xxx';` which is quite ironic since that would be a major reason to switch to ESM. Jest limitation or I may have missed something here.

### Additional reading

- [Jest ESM support documentation page](https://jestjs.io/docs/ecmascript-modules)
- [Jest main GitHub issue about ESM support](https://github.com/facebook/jest/issues/9430)
- [ts-jest ESM support documentation page](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support)

### DIFF
--- a/libs/components/jest.config.cjs
+++ b/libs/components/jest.config.cjs
@@ -2,18 +2,11 @@
 
 module.exports = {
   displayName: 'components',
-//   preset: '../../jest.preset.js',
   preset: 'ts-jest/presets/default-esm',
-	// resolver: '@nrwl/jest/plugins/resolver',
   testEnvironment: 'jsdom',
   moduleNameMapper: {
 	  "@vivid-nx/shared": "<rootDir>/../../libs/shared/src/index.ts",
-    // "dialog-polyfill": "<rootDir>/../../node_modules/dialog-polyfill/dist/dialog-polyfill.js"
   },
-  // roots: ['<rootDir>'],
-  // modulePaths: ['<rootDir>', '/dist'],
-  // moduleDirectories: ['node_modules', 'dist'],
-
   globals: {
     'ts-jest': {
   		useESM: true,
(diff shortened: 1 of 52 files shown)


olaf-k:
Keeping for the future, closing for now.

PR #881, opened 2022-12-01
feat(doc): a more interactive `--vvd-size-density` example

olaf-k:


docs/design/design-tokens.md
```diff
@@ -137,21 +137,36 @@ Due to a11y and design constraints, the density can only be set to one of the fo
...
-  --vvd-size-density: -1;
+ #scoped-region {
+  --vvd-size-density: 1;
  }
 </style>
 
-<vwc-header>
-  This header is condensed
-</vwc-header>
+<vwc-number-field label="Choose density" helper-text="Change the density value" min="-1" max="2" value="2" style="justify-self: flex-start"></vwc-number-field>
```
rachelbt, line 147:
```suggestion
<vwc-number-field label="Choose density" helper-text="Change density value" min="-1" max="2" value="2" style="justify-self: flex-start; width: 105px;"></vwc-number-field>
```

rachelbt (commented):
just style :)

yinonov:
for some reason iframe height is smaller than source document content

yinonov:
> for some reason iframe height is smaller than source document content

I'm not sure but this might occur in other examples occasionally too

yinonov:
> for some reason iframe height is smaller than source document content

I'm not sure but this might occur in other examples occasionally too

olaf-k:
> > for some reason iframe height is smaller than source document content
> 
> I'm not sure but this might occur in other examples occasionally too

yes, there doesn't appear to be a failsafe way to resize them precisely.
it's also annoying to fix as I can't reproduce the behavior locally, even when throttling cpu/connection.

olaf-k:
I've added a resize observer that self-destruct after 3s of inactivity, this seems to do the trick...

yinonov (approved):

PR #883, opened 2022-12-01
perf(components): fix button non-responsive display flow

yinonov:


rachelbt (commented):
can't believe it didn't came up earlier.
Wondering if it needs to be limited to not stacked not icon.
WDYT?

rachelbt (approved):
LGTM

PR #888, opened 2022-12-05
fix(components): listbox and radio-group disabled state

olaf-k:
This aims at fixing [this comment](https://github.com/Vonage/vivid-3/pull/815#discussion_r1020900567) from #815 

When they are set to disabled, `listbox` and `radio-group` now record the state of their children then set them as disabled.
In #815 `listbox` uses CSS to disable access to its children.
In current main, `radio-group` does not keep track of its children's state.

Issues with the current code:
- In a disabled `listbox`, a selected `option` will still react to hover. This should be addressed by the `option` component in a separate PR.
- The version of the FAST `radio-group` we inherit from considers `disabledChange` as private instead of protected in the current codebase (thus the ugly TS workaround).


libs/components/src/lib/listbox/listbox.scss
```diff
@@ -1,17 +1,17 @@
-@use '../focus/partials/variables' as focus-variables;
-@use '../../../../shared/src/lib/sass/mixins/connotation/config' with (
+@use "../focus/partials/variables" as focus-variables;
```
yinonov, line 1: @rinaok revert those double quotes. the change isn't fixed by styelint rule but probably by a local extension

libs/components/src/lib/listbox/listbox.ts
```diff
@@ -22,25 +23,29 @@ export class Listbox extends FoundationListboxElement {
...
-		if (name === 'disabled') {
-			this.#disableSlottedChildren();
+	/**
+	 * Saves/restore children state when the `disabled` property changes.
+	 *
+	 * @param _ - The previous disabled value
+	 * @param becomesDisabled - The next disabled value
+	 * @internal
+	 */
+	disabledChanged(_: boolean, becomesDisabled: boolean): void {
```
yinonov, line 40: align to lifecycles common convention? (`prev` / `next`)
```suggestion
	disabledChanged(_: boolean, next: boolean): void {
```

yinonov, line 40: what if new items are added to the stack?

yinonov (changes requested):

yinonov:
here's today's mentioned feature. not enough support and won't be of great help here but useful to know about
https://css-tricks.com/custom-state-pseudo-classes-in-chrome/

yinonov:
> > anyway, as you wrote, this PR tries to go beyond CSS by switching a behavior (disabled) on and off. *the cosmetic aspect of that behavior should be left to the child to decide*.
> 
> 
> 
> talking to myself here, but that's actually not true for a native select. options are not disabled when the select is, but are in practice unreachable.

Just throwing an idea here -
We might be able to extend the appearance mixin disabled selector by it to also apply when a parent is disabled.
That's not a real css feature as host-context hasn't reached maturity.
I did see nice tricks with :has to

olaf-k:
> Just throwing an idea here - We might be able to extend the appearance mixin disabled selector by it to also apply when a parent is disabled. That's not a real css feature as host-context hasn't reached maturity. I did see nice tricks with `:has` to select elements by parents. maybe @rachelbt would have references?

yes! I thought about `:has` as well. my main concern is that css solves the appearance side of the problem, not the functionality. for the radio-group, this means that it could *look* disabled while still being submitted as part of its parent form.

libs/components/src/lib/listbox/listbox.ts
rinaok: @olaf-k That's right... never thought of it. Using CSS to set the disabled state will be the best solution, in my opinion. We can just merge this old PR https://github.com/Vonage/vivid-3/pull/815 if you agree. Managing disabled children can cause problems that we didn't even consider, as @yinonov wrote.

rinaok:
> > Just throwing an idea here - We might be able to extend the appearance mixin disabled selector by it to also apply when a parent is disabled. That's not a real css feature as host-context hasn't reached maturity. I did see nice tricks with `:has` to select elements by parents. maybe @rachelbt would have references?
> 
> yes! I thought about `:has` as well. my main concern is that css solves the appearance side of the problem, not the functionality. for the radio-group, this means that it could _look_ disabled while still being submitted as part of its parent form.
> 
> here's a summary of current possible paths at the moment:
> 
> ### setting the disabled state via CSS
> #### pros
> * consistent with FAST's listbox implementation.
> * parent-only, children don't have to know/do anything.
> * keep children state.
> 
> #### cons
> * limited visually: you can only set a dim veil over children. other solutions require some kind of adherence between the children and the parent (i.e., FAST's options have a dedicated CSS rule acknowledging the parent's disabled state).
> * what about functionality? a disabled component should not react to user inputs _and_ should be ignored in forms.
> 
> ### setting the disabled state via JS (or attribute)
> #### pros
> * consistent with FAST's radio-group implementation, although it doesn't keep state as is.
> * parent-only, children don't have to know/do anything.
> * appearance and functionality are consistent.
> 
> #### cons
> * edge-cases (the children list is modified while disabled) to account for when keeping state.
> * not consistent with native behavior (a disabled select does not set its children options to disabled).
> 
> ### notes
> * the read-only state for radios shares the same pattern with an additional constraint: the state is only valid when used inside a radio-group. this is the native behavior but we could ignore it.
> * "the listbox component has no internals related to form association" from [FAST's documentation](https://www.fast.design/docs/components/listbox).

https://github.com/Vonage/vivid-3/pull/888#discussion_r1045719265

olaf-k:
OK, so the latest commits here go the JS route for you to see what it'd look like. Basically, we keep a record of the children's state when turning the parent (listbox/radio-group) disabled and we restore the recorded state when turning disabled off.

Children updates (add/removal) are monitored and update the records.

It works and doesn't require CSS, but definitely needs more code than the CSS solution. This CSS solution would consist in:
- merging @rinaok code in #815 
- replicating it for radio-group (meaning updating the radios CSS and override the current `disabledChanged` method)
- preventing the radio-group from submitting: this can be achieved by removing its `name` while disabled, but is it acceptable?

olaf-k:
For reference https://github.com/microsoft/fast/issues/6567

acelletti:
Add JIRA ticket [VIV-866 - Investigate disable state bug in Fast](https://jira.vonage.com/browse/VIV-866)

PR #890, opened 2022-12-06
feat(components): override default prefix api

yinonov:
- [x] ~~only `index.js` exists in the badge folder. need to split it into index.js and definition.js~~
- [x] definition now don't export by default. should have rollup export it by default
- [x] rename definition export. instead of `badgeDefinition` to just `badge`
- [x] don't export anything from base class to be exposed outside the library (like the actual Badge class)
- [x] remove https://github.com/Vonage/vivid-3/blob/8799fee84fdf3291180dd840e881f1c287eaa186/libs/components/src/shared/design-system/index.ts#L4-L5
- [x] remove https://github.com/Vonage/vivid-3/blob/8799fee84fdf3291180dd840e881f1c287eaa186/libs/components/src/shared/design-system/index.ts#L12
- [x] document new feature
- [x] allow (and document) the following interface

```js
import { register } from '@vonage/vivid';
import { badge, button, ... } from '@vonage/vivid';

register('vvd', badge, button, ...);
```

<hr>

~~an approach simpler to manage will revert back to using side effect imports. instead of `import.meta.url` we expose a setter for the prefix to override globally, prior to components import.~~

~~in a react app this is ordered by calling the setter from `index.js`~~

image

~~and `import()` of custom element registration from react components~~
~~dynamic import is required for the prefix override call to kick in before registration fires~~

image

~~no significant difference shown in bundled assets~~


YonatanKra (changes requested):
Tests? ;)

libs/components/src/shared/design-system/index.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { icon } from '../../lib/icon/definition';
+import { badge } from '../../lib/badge/definition';
+import { registerFactorial } from '.';
+
+describe('design system', () => {
```
YonatanKra, line 5: image

YonatanKra (changes requested):
The component generation template should be updated. Let me make my attempt at it.

libs/components/src/shared/design-system/index.ts
```diff
@@ -1,12 +1,13 @@
...
 // common components dependency
 /**
  * @param {HTMLElement} element - The element to get or create a design system for.
  * @param prefix
  * @returns {DesignSystem} - Represents a configurable design system. An API gateway to design system features.
  */
-export const designSystem = DesignSystem.getOrCreate();
+const designSystem = DesignSystem.getOrCreate();
+
+export const registerFactorial = (...args: any[]) => (prefix = defaultPrefix) => designSystem.withPrefix(prefix).register(...args);
```
olaf-k, line 13: Unless I'm missing something, there is no need to destructure your arrays (creating copies) every time?
```suggestion
export const registerFactorial = (args: any[]) => (prefix = defaultPrefix) => designSystem.withPrefix(prefix).register(args);
```
and in every definition file, just call `export const registerSomething = registerFactorial(somethingElements);`

libs/components/README.md
```diff
@@ -77,27 +77,32 @@ Note: scss users can simply [forward](https://sass-lang.com/documentation/at-rul
...
 
 ```html
-<dashboard-badge text="I'm a custom prefixed badge"></dashboard-badge>
+<dashboard-badge text="dashboard scoped badge"></dashboard-badge>
 ```
 
-Even though the custom elements are registered under different namespaces, [npm packages version range handling](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies) can still be used for multiple versions saving.
+Remember to not include the default side-effect import (`import '@vonage/vivid/button';`) when using scoped elements as it will register the default namespace.
+
+one of the pros of this approach is that, although custom elements are registered under different namespaces, [npm packages version range handling](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies) can still be used for multiple versions saving.
```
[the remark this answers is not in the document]
YonatanKra: Yea. I don't understand what `versions saving` means.

libs/components/src/shared/design-system/index.ts
yinonov: It accepts parameters not array. Did u try?

libs/nx-vivid/src/generators/component/files/definition.ts__tmpl__
```diff
@@ -0,0 +1,21 @@
+import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
+import { registerFactorial } from '../../shared/design-system';
+import styles from './<%= fileName %>.scss';
+
+import { <%= className %> } from './<%= fileName %>';
+import { <%= className %>Template as template } from './<%= fileName %>.template';
+
+export const vivid<%= className %> = <%= className %>.compose<FoundationElementDefinition>({
```
yinonov, line 8: class name here should be camel cased
```suggestion
export const <%= className %> = <%= className %>.compose<FoundationElementDefinition>({
```

libs/nx-vivid/src/generators/component/files/definition.ts__tmpl__
```diff
@@ -0,0 +1,21 @@
...
+	baseName: '<%= name %>',
+	template: template as any,
+	styles,
+})();
+
+export const <%= camelCasedName %>Elements = [vivid<%= className %>];
```
yinonov, line 14: here too
```suggestion
export const <%= camelCasedName %>Elements = [<%= className %>];
```

libs/components/README.md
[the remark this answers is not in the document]
yinonov: what's the correct term for when npm resolute to the same version even if 2 dependencies point to different versions. to avoid bloating the bindle size

libs/components/src/shared/design-system/index.ts
olaf-k: oops, you're right, suggestion updated.

libs/components/README.md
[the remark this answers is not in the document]
olaf-k: an error? 😁

(are you thinking of `npm dedupe`?)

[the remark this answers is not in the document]
yinonov: Yes. I'll rephrase

libs/components/src/shared/design-system/index.ts
yinonov: The new suggestion will assign only the 1st argument to `args`

YonatanKra (dismissed):
LGTM

libs/components/src/shared/design-system/index.ts
olaf-k: in definition.ts files:
```js
export const widgetElements = [widget, ...iconElements, ...focusElements];
```
`widgetElements` is an **array**, that you pass to `registerFactorial`:
```js
export const registerWidget = registerFactorial(widgetElements);
```
thus, in the shared util file:
```js
export const registerFactorial = (args: any[]) => (prefix = defaultPrefix) => designSystem.withPrefix(prefix).register(...args);
```
`registerFactorial` is defined as a function accepting an **array**, which is later spread since register accepts 1..n arguments.

your original code defines `registerFactorial` as a function with 1..n arguments (using the spread operator), which forces you to spread each `widgetElements` when calling it instead of simply passing its reference as I suggest.

yinonov: thanks! commited

olaf-k (dismissed):
👍

YonatanKra (dismissed):
LGTM. Exciting :)

PR #893, opened 2022-12-06
refactor(components): prevent events propagation

yinonov:
Closes vonage/vivid#1364

### DIFF
--- a/apps/docs/.eleventy.js
+++ b/apps/docs/.eleventy.js
@@ -26,6 +26,9 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.setBrowserSyncConfig({
     server: {
       baseDir: OUTPUT_DIR
+	},
+    snippetOptions: {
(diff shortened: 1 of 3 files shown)


YonatanKra (changes requested):
Since this change changes the API and we'd like it to be documented, a test is in order

yinonov:
accordion test fail when listening to `opened` events bubbling up from its assigned accordion-items.
This PR aims to prevent events from bubbling up the DOM tree, thus fails accordion from handling ['should only allow one accordion items open at a time'](https://github.com/Vonage/vivid-3/blob/c0475c616c5bf6f118f54803c38bee0fed88de5f/libs/components/src/lib/accordion/accordion.spec.ts#L44)

yinonov:
follow https://vonage.slack.com/archives/G016RPD3ANB/p1671620312992219 to address next steps

PR #895, opened 2022-12-07
fix(components): textfield ghost appearance missing shades

### LINKED ISSUE #889: [appearance]: bleeding css variables
When inserting a text-field inside action-group the text-field gets `faint` background color on hover.
Not sure that action-group needs all its connotation shades

https://codepen.io/rachelbt/pen/jOKXypY

yinonov:
Closes #889


rachelbt:
ghost text-area & text-field need the faint bg on hover?

yinonov:
> ghost text-area & text-field need the faint bg on hover?

yes
image

libs/components/src/lib/text-area/text-area.scss
```diff
@@ -3,7 +3,7 @@
 @use '../../../../../dist/libs/tokens/scss/tokens.constants' as constants;
 @use '../../../../shared/src/lib/sass/mixins/connotation/config' with (
 	$connotations: accent success alert,
-	$shades: primary backdrop intermediate firm,
+	$shades: primary backdrop intermediate firm faint soft,
```
rachelbt, line 6: it has no effect cause there's no ghost appearance to text-area. But I guess its right that it will be here if some day we will have the ghost

rachelbt (approved):
we should really have test for over and pressed... But thats another issue.
LGTM

libs/components/src/lib/text-area/text-area.scss
yinonov: we should

PR #903, opened 2022-12-12
feat(popup): anchor can recieve  HTMLElement

rinaok:


libs/components/src/lib/popup/popup.ts
```diff
@@ -157,7 +157,10 @@ export class Popup extends FoundationElement {
 	/**
 	 * Gets the anchor element by id
 	 */
-	#getAnchorById(): HTMLElement | null {
+	#getAnchor(): HTMLElement | null {
+		if(this.anchor instanceof HTMLElement){
+			return this.anchor;
+		}
 		return document.getElementById(this.anchor);
```
olaf-k, line 164:
```suggestion
		return this.anchor instanceof HTMLElement ? this.anchor : document.getElementById(this.anchor);
```

olaf-k (dismissed):

PR #907, opened 2022-12-13
feat(popup): add absolute strategy

rinaok:
needs to be merged before https://github.com/Vonage/vivid-3/pull/825


libs/components/src/lib/popup/popup.template.ts
```diff
@@ -32,7 +32,7 @@ export const popupTemplate: (
 	return html`
   <${elevationTag}>
     <!--the popup-wrapper is needed for alternating the inside of the popup nd not its shadow-->
```
olaf-k, line 34: not related to this PR but this comment should be extracted from the template so it doesn't pollute the DOM

olaf-k (approved):

PR #909, opened 2022-12-13
feat(components): preload icons

olaf-k:
Affected components:
- accordion-item
- nav-disclosure
- menu-item

(Note that this could also be a `preload?` parameter of the functions returned by `affixIconTemplateFactory`, but this would restrain the scope.)

(Also it's not *strictly speaking* preloading since we're not awaiting the load 😶)


yinonov:
Consider the alternative of using inline svg in markup or even inline svg via css

olaf-k:
> Consider the alternative of using inline svg in markup or even inline svg via css

the current solution favors reusing the existing memoization mechanism

olaf-k:
> Reference

olaf-k:
> bear in mind that in Accordion the icon can be changed by setting the property `icon`. We can in the `nav-discloure` consider using only the up icon and us css to rotate it. We can do it in accordion also but only when icon is not set by consumers

yup, this PR only focuses on the default open/close glyphs. we could think of a more generic preloading mechanism for all icons/images used across widgets but that goes beyond the scope of this particular PR then ;)

yinonov (changes requested):
can we feature that in menu item too?
image

olaf-k:
> can we feature that in menu item too? image

I need to fix that damn UT first 😅
but yes, sure, good point!

olaf-k:
> I need to fix that damn UT first

tests are passing when replacing the default mock by `fetch.mockResponse(() => new Promise(() => {}));` in setupJestTests, which would be valid for all tests. I tried applying that locally in affected tests without success.

(also, `resolveIcon` now exported needs to be tested...)

PR #913, opened 2022-12-14
fix(calendar): test

rinaok:
Screenshot 2022-12-14 at 16 48 10


olaf-k:
> Screenshot 2022-12-14 at 16 48 10

Thanks, I've been meaning to fix that. The problem comes from the weird whitespace though, not the capitalization. Also, this tries to fix the test but not the existence of said weird whitespace.

olaf-k (approved):

PR #915, opened 2022-12-14
refactor(components): align all icons integrations with affix pattern

rachelbt:


libs/components/src/shared/patterns/affix.ts
```diff
@@ -51,6 +52,8 @@ export const affixIconTemplateFactory: (context: ElementDefinitionContext) =>
 (icon?: string, className?: string) => ViewTemplate<AffixIcon> | null = (context: ElementDefinitionContext) => {
 	const iconTag = context.tagFor(Icon);
 	return (icon?: string, className?: string) => icon
-		? html`<span class="${className ? className : 'icon'}"><${iconTag} :name="${() => icon}"></${iconTag}></span>`
+		? html`
+			<style>${messageStyles}</style>
```
olaf-k, line 56: the style should probably be inlined or be made more generic rather than injecting `<styles>`.

libs/components/src/shared/patterns/affix.scss
```diff
@@ -0,0 +1,5 @@
+$icon-line-height: --_icon-line-height;
+
+.icon {
+	line-height: var(#{$icon-line-height}, 1);
```
olaf-k, line 4: the class name is an optional parameter of the function returned by `affixIconTemplateFactory`, it may not always be `icon`

olaf-k (changes requested):

libs/components/src/lib/avatar/avatar.spec.ts
```diff
@@ -24,7 +25,6 @@ describe('vwc-avatar', () => {
 
 	describe('basic', () => {
 		it('should be initialized as a vwc-avatar', async () => {
-			expect(avatarDefinition()).toBeInstanceOf(FoundationElementRegistry);
```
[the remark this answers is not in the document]
rachelbt: It may be removed by me when I merged from main...
@YonatanKra mind looking?

libs/components/src/shared/patterns/affix.scss
rachelbt: oh - I missed the part that its not have to be icon class....
Do you think it should be line an inline style?
```<span style="line-height: var(--_icon-line-height, 1);" class="${className ? className : 'icon'}"><${iconTag} :name="${() => icon}"></${iconTag}></span>```

Comment (olaf-k) [reply libs/components/src/shared/patterns/affix.scss:4]: you're the CSS wizard! 😁
I'm guessing there's a nicer way to do it but it would involve refactoring accordion-item and having the proper CSS injected when you use the affix helper, soooo... that's definitely faster.

File: libs/components/src/lib/avatar/avatar.spec.ts [hunk 4]
```diff
@@ -24,7 +25,6 @@ describe('vwc-avatar', () => {
 
 	describe('basic', () => {
 		it('should be initialized as a vwc-avatar', async () => {
```
yinonov, line 27:
```suggestion
		it('should be initialized as a vwc-avatar', async () => {
			expect(avatarDefinition()).toBeInstanceOf(FoundationElementRegistry);
```

libs/components/src/shared/patterns/affix.ts
rachelbt: on second thought - this is what we do in the form-element pattern...

libs/components/src/shared/patterns/affix.scss
rachelbt: it actually has nothing with css wizarding... its about the class names or options
@yinonov WDYS? should we add the stile inline?

yinonov: inline style won't scale as it is limited when it comes to dynamic permutations

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -72,6 +72,9 @@
 
 			.icon {
 				font-size: #{size.$vvd-size-extra-condensed};
+				> * {
```
YonatanKra, line 75: Oh?

libs/components/src/shared/patterns/affix.scss
rachelbt: that means that anyone who is using the affix icon or the form pattern will have to remember adding the import of the css. right?

YonatanKra (dismissed):
LGTM

libs/components/src/lib/dialog/dialog.scss
rachelbt: by default its `vertical-align: sub; `but here we need top. Cant target `vwc-icon` so * is the option to use :)

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -52,6 +52,10 @@
 			grid-column: 1/2;
 			grid-row: 1/2;
 			margin-block-end: #{size.$vvd-size-super-condensed};
+
+			> * {
+				vertical-align: top;
+			}
```
yinonov, line 58: relate?

libs/components/src/lib/option/option.scss
```diff
@@ -39,7 +39,6 @@
 	hyphens: auto;
 	inline-size: 100%;
 	min-block-size: $min-block-size;
-	padding-block: 10px;
```
yinonov, line 42: relates?

yinonov (changes requested):

libs/components/src/lib/dialog/dialog.scss
rachelbt: yes :) 
without it the icon doesn't sit as it was.
It was aligned because the line-height was wrong. now that it is fixed - thin is needed :)

libs/components/src/lib/option/option.scss
rachelbt: In a way, yes.  I noticed that this made the option be more than 40px when only one line + icon in it.

libs/components/src/shared/patterns/affix.scss
rachelbt: as was decided in a talk with @yinonov:
I've inserted the style as inline - until [this will be tended](https://github.com/Vonage/vivid-3/issues/950)

rachelbt:
In a second thought we decided to stop and think.
Not all places that use icon need the affix. Maybe we need an icon pattern?
More style is related to affix pattern.
Will be decided later.
Line height:1 will be added to the style of each component where is needed.

libs/components/src/shared/patterns/affix.ts
```diff
@@ -51,6 +51,10 @@ export const affixIconTemplateFactory: (context: ElementDefinitionContext) =>
 (icon?: string, className?: string) => ViewTemplate<AffixIcon> | null = (context: ElementDefinitionContext) => {
 	const iconTag = context.tagFor(Icon);
 	return (icon?: string, className?: string) => icon
-		? html`<span class="${className ? className : 'icon'}"><${iconTag} :name="${() => icon}"></${iconTag}></span>`
+		? html`
+			<!-- Using inline style until the affix + form elements will be created as scss mixin (vivid-3/issues/950)-->
```
olaf-k, line 55: this will be outputted in the markup. should be moved as a JS comment before the `return` line.

rachelbt:
was solved differently in here:
https://github.com/Vonage/vivid-3/pull/951

PR #917, opened 2022-12-14
chore(components): style listbox `disabled` inherits into option role

yinonov:
this reverts disabled appearance of options back to be defined by the listbox widget containing parent.
This a comprisable solution for the interim. it is further researched in microsoft/fast#6567


rachelbt (commented):
looks OK code-wisely.
Its a bit too much grey in a pack.
I know its the tokens - but maybe 100 for disable is better?
image

yinonov:
> looks OK code-wisely.
> 
> Its a bit too much grey in a pack.
> 
> I know its the tokens - but maybe 100 for disable is better?
> 
> image
> 
> 
> 
> 

Listbox don't care what tokens are assigned to appearance fieldset

rachelbt (approved):
LGTM

PR #923, opened 2022-12-21
fix(listbox): inline size

rinaok:


libs/components/src/lib/listbox/listbox.scss
```diff
@@ -15,6 +15,7 @@
 
 :host {
 	display: block;
+	inline-size: fit-content;
```
YonatanKra, line 18: Doesn't fit content come with `max size`?
E.g. `fit-content(--vvd-listbox-max-size)`

rachelbt: @YonatanKra when its on the host the consumer can change it :)

yinonov (dismissed):
reduce ui test image size

rinaok:
> reduce ui test image size

@yinonov it's ok now?

rachelbt (approved):
LGTM

PR #924, opened 2022-12-21
fix(action-group): update snapshots

rinaok:


rachelbt (changes requested):
why did the action group border-color changed?
Is it suppose to be like the change?

rinaok:
> why did the action group border-color changed? Is it suppose to be like the change?

IDK... it's failing in another PR so I updated it

rachelbt (changes requested):
I think that the ui-test is written differently that it's not updating - this should fail all the PR and it doesn't...

PR #930, opened 2022-12-28
feat(components)!: fast-based accordion

olaf-k:
### Read me first:

Breaking API changes (due to inheritance):
- Accordion: `multi` (`true` | `false`) changed to `expand-mode` (`multi` | `single`, default to `single` vs `multi` in FAST).
- AccordionItem: `open` renamed to `expanded`.
- AccordionItem: `headingLevel` (`2` to `6`, default `3`) renamed to `headinglevel` (`1` to `6`, default `2`).

Behavior change:
- Accordion: when `expand-mode` is set to `single`, one item must remain opened at all times (i.e., you can't close all items).

Limitations/issues:
- Design: icons and text have the same color (as in Figma), no leading indicator (contrary to Figma).
- Code: [there are FAST issues open and ongoing](https://github.com/Vonage/vivid-3/pull/930#issuecomment-1367374337)


yinonov:
what about `open` property? changes to `expand`?

yinonov:
- heading level gets picked up by the a11y tree even when inside shadow root.
- it's pretty convenient to use an actual heading tag dynamically with the fast-element. so why not use it
- I think in terms of looks - icon, indicator, leading / trailing, accordion item has already been shaped according to requests so let's keep `No-Indicator`, `Meta`, `Icon`, `Icon-Trailing`

olaf-k:
> * heading level gets picked up by the a11y tree even when inside shadow root.
> * it's pretty convenient to use an actual heading tag dynamically with the fast-element. so why not use it

There might be a misunderstanding here: the heading level *still exists*, only it is assumed by the button's `aria-level`. It's just the unnecessary `<hx>` tag that went away. FAST does something similar (they put `aria-level` on a div, their markup is different).

> * I think in terms of looks - icon, indicator, leading / trailing, accordion item has already been shaped according to requests so let's keep `No-Indicator`, `Meta`, `Icon`, `Icon-Trailing`

I can remove `leading`, no worries, but isn't Figma also a result of users requests?

olaf-k:
Note to self:
- `aria-disabled` is not reset when `expand-mode` changes from `single` to `multi` (FAST bug?)

libs/components/src/lib/accordion/accordion.scss
```diff
@@ -2,9 +2,6 @@
 
 .base {
 	display: flex;
+	box-sizing: border-box;
 	flex-direction: column;
 }
-
```
rachelbt, line 7: why did you moved this to be as bottom border on the accordion-item?

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,65 @@ describe('vwc-accordion-item', () => {
...
+			expect(element.expanded).toBeTruthy();
+			element.shadowRoot?.querySelector('button')?.click();
+			expect(element.expanded).toBeFalsy();
 		});
 	});
 
 	describe('icon', () => {
-		it('should set icon class', async () => {
-			expect(getBaseElement(element).classList.contains('icon')).toBeFalsy();
+		it('should set icon', async () => {
```
YonatanKra, line 47:
```suggestion
		it('should render icon when icon is set', async () => {
```

libs/components/src/lib/accordion-item/accordion-item.template.ts
```diff
@@ -1,66 +1,54 @@
...
-const getClasses = ({
-	open, iconTrailing, icon, noIndicator
-}: AccordionItem) => classNames(
-	'base',
-	['open', open],
-	['icon', Boolean(icon)],
-	['icon-trailing', iconTrailing],
-	['no-indicator', noIndicator],
-);
+	/* eslint-disable @typescript-eslint/indent */
```
YonatanKra, line 11: What? Why the eslint-disable?

YonatanKra (changes requested):
Finished accordion-item.
More after lunch :)

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +36,62 @@ describe('vwc-accordion', () => {
...
+			accordionItem1.expanded = true;
 			await elementUpdated(element);
-			accordionItem2.open = true;
+			accordionItem2.expanded = true;
 			await elementUpdated(element);
 
-			expect(accordionItem1.open).toBeFalsy();
-			expect(accordionItem2.open).toBeTruthy();
+			// FAST issue here
+			// expect(accordionItem1.expanded).toBeFalsy();
```
YonatanKra, line 54: What is the FAST issue?

olaf-k:
> Finished accordion-item. More after lunch :)

haven't read your comments yet but please note that I did not write the tests. I merely updated their syntax to deal with the new prop names (e.g., `open` > `expanded`) and assumed that they were already reviewed for the existing accordion.

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +36,62 @@ describe('vwc-accordion', () => {
...
+			accordionItem1.expanded = true;
 			await elementUpdated(element);
-			accordionItem2.open = true;
+			accordionItem2.expanded = true;
 			await elementUpdated(element);
 
-			expect(accordionItem1.open).toBeTruthy();
-			expect(accordionItem2.open).toBeTruthy();
+			expect(accordionItem1.expanded).toBeTruthy();
+			expect(accordionItem2.expanded).toBeTruthy();
```
YonatanKra, line 73: Here I must insist on the AAA pattern. It is pretty hard to follow what's happening at a glance.

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +36,62 @@ describe('vwc-accordion', () => {
...
 
 			element.closeAll();
 			await elementUpdated(element);
 
-			expect(accordionItem1.open).toBeFalsy();
-			expect(accordionItem2.open).toBeFalsy();
+			expect(accordionItem1.expanded).toBeFalsy();
+			expect(accordionItem2.expanded).toBeFalsy();
```
YonatanKra, line 94: Same AAA comment here as well.

libs/components/src/shared/patterns/affix.ts
```diff
@@ -50,7 +50,7 @@ export class AffixIconWithTrailing extends AffixIcon {
 export const affixIconTemplateFactory: (context: ElementDefinitionContext) =>
 (icon?: string, className?: string) => ViewTemplate<AffixIcon> | null = (context: ElementDefinitionContext) => {
 	const iconTag = context.tagFor(Icon);
-	return (icon?: string, className?: string) => icon
-		? html`<span class="${className ? className : 'icon'}"><${iconTag} :name="${() => icon}"></${iconTag}></span>`
+	return (icon?: string) => icon
+		? html`<span class="${'icon'}"><${iconTag} :name="${() => icon}"></${iconTag}></span>`
```
YonatanKra, line 54: image

YonatanKra (changes requested):

YonatanKra:
> > Finished accordion-item. More after lunch :)
> 
> haven't read your comments yet but please note that I did not write the tests. I merely updated their syntax to deal with the new prop names (e.g., `open` > `expanded`) and assumed that they were already reviewed for the existing accordion.

IMO, that's the problem with code that's not following standards.
Ppl copy-paste or learn not-so-good habits ;)

libs/components/src/lib/accordion-item/README.md
```diff
@@ -23,38 +23,26 @@ Add the `heading` attribute to set the heading text.
 
 ### Heading-Level
 
-Use the `heading-level` attribute (or `headingLevel` property) to set the accordion heading level so it fits correctly within the outline of the page. Read more on [heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+Use the `heading-level` attribute (or `headinglevel` property) to make the accordion heading level fit the page outline. Read more on [heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
 
-- Type: `2` | `3` | `4` | `5` | `6`
-- Default: `3`
+- Type: `1` | `2` | `3` | `4` | `5` | `6`
+- Default: `2`
```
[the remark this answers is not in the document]
olaf-k: yes, these values come directly from the FAST component and I thought the differences were small enough to warrant extra code to override them.
if you think it is important I can add that code though.

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -14,94 +14,70 @@
...
 	display: flex;
+	box-sizing: border-box;
 	flex-direction: column;
-
-	&:not(.open) {
-		.body {
-			display: none;
-		}
-	}
+	border-bottom: 1px solid var(#{constants.$vvd-color-neutral-200});
```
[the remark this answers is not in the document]
olaf-k: reasons to move the border to the item instead of the parent:
1. simpler parent css
2. same as FAST
3. same as [our Figma](https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19073%3A77179&t=0wbeVDNWnTtxFTHy-0)
@AyalaBu opinion?

libs/components/src/lib/accordion/accordion.scss
olaf-k: [see my other comment](https://github.com/Vonage/vivid-3/pull/930#discussion_r1083876783)

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -14,94 +14,70 @@
...
 }
 
-.header {
-	padding: 0;
+h1, h2, h3, h4, h5, h6 {
```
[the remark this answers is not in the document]
olaf-k: 👍 will do

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -14,94 +14,70 @@
...
 	margin: 0;
 }
 
-.button {
+button {
```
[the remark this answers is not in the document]
olaf-k: but that's just one element here though?

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -14,94 +14,70 @@
...
-	.indicator {
-		margin-inline-start: 16px;
+	:host([expanded]) & {
+		display: block;
 	}
 
-	.icon {
-		color: var(#{constants.$vvd-color-neutral-600});
-		margin-inline-end: 16px;
+	:where(h1, h2, h3, h4, h5, h6):has(> button > .icon:nth-child(2)) + & {
```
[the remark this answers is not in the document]
olaf-k: it does work (see the doc's icon example part).
but you're right about support (even FF, such a pity). I'll find another way!

libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -14,94 +14,70 @@
...
+	padding: 8px 32px 24px 16px;
 
-	.indicator {
-		margin-inline-start: 16px;
+	:host([expanded]) & {
+		display: block;
 	}
 
-	.icon {
-		color: var(#{constants.$vvd-color-neutral-600});
```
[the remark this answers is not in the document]
olaf-k: our Figma disagrees.
@AyalaBu should the accordion-item icons have a different color?

libs/components/src/lib/accordion-item/accordion-item.template.ts
olaf-k: eslint struggles with the markup/code mix inside our template strings.
how it looks now:
image
how it looks when eslint is satisfied:
image

I believe linters should not impair readability, but if feel strongly in favor of eslint, I can reformat and remove the directive.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,65 @@ describe('vwc-accordion-item', () => {
...
-			expect(getBaseElement(element).classList.contains('open')).toBeTruthy();
-
-			button.click();
-			await elementUpdated(element);
-			expect(element.open).toBeFalsy();
-			expect(getBaseElement(element).classList.contains('open')).toBeFalsy();
+	describe('click', () => {
+		it('should open/close on click', async () => {
+			expect(element.expanded).toBeFalsy();
+			element.shadowRoot?.querySelector('button')?.click();
```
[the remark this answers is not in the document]
olaf-k: made the code clearer

libs/components/src/lib/accordion/accordion.spec.ts
olaf-k: clarified

libs/components/src/lib/accordion/accordion.spec.ts
olaf-k: just to reiterate: those tests were already approved, I just updated the API calls.

I also have no trouble reading them. I added comments to make it easier to parse quickly. I'm happy to commit any suggestion you have if that's not enough.

libs/components/src/lib/accordion/accordion.spec.ts
olaf-k: same reply as above

libs/components/src/shared/patterns/affix.ts
olaf-k: I actually removed a functionality I introduced myself 😁

libs/components/src/lib/accordion-item/accordion-item.scss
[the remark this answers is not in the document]
rachelbt: the icon color change was part of design alignment:
[see this PR](https://github.com/Vonage/vivid-3/pull/802/files)

libs/components/src/lib/accordion-item/accordion-item.scss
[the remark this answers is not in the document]
olaf-k: I trust you!

libs/components/src/lib/accordion-item/accordion-item.scss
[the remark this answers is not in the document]
olaf-k: did I forgot a conversation I was part of? maybe. MAYBE.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,69 @@ describe('vwc-accordion-item', () => {
...
+			expect(element.expanded).toBeFalsy();
 		});
 	});
 
 	describe('icon', () => {
-		it('should set icon class', async () => {
-			expect(getBaseElement(element).classList.contains('icon')).toBeFalsy();
+		it('should render an icon when the icon property is set', async () => {
+			const itemHeaderButton = element.shadowRoot?.querySelector('.heading-button') as HTMLButtonElement;
+			expect(itemHeaderButton.querySelector(':nth-child(2)')?.classList).not.toContain('icon');
```
YonatanKra, line 50: Caching the item would make this code much simpler.
`const headerButtonIconWrapper itemHeaderButton.querySelector(':nth-child(2)')`;
And use it in the 3 times it is used below.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,69 @@ describe('vwc-accordion-item', () => {
...
+			expect(itemHeaderButton.querySelector(':nth-child(2)')?.classList).toContain('icon');
+			expect(itemHeaderButton.querySelector(':nth-child(2) > vwc-icon')?.getAttribute('name')).toBe('chat-solid');
 		});
-		it('should set iconTrailing', async () => {
-			expect(getBaseElement(element).classList.contains('icon-trailing')).toBeFalsy();
+
+		it('should render a trailing icon when the iconTrailing property is set', async () => {
+			const itemHeaderButton = element.shadowRoot?.querySelector('.heading-button') as HTMLButtonElement;
+			await elementUpdated(element);
+			expect(itemHeaderButton.querySelector(':last-child > vwc-icon')?.getAttribute('name')).toBe('chevron-down-solid');
```
YonatanKra, line 60: Same goes with this long query regarding caching it for later use and clarity.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,69 @@ describe('vwc-accordion-item', () => {
...
-			expect(getBaseElement(element).classList.contains('no-indicator')).toBeTruthy();
+			expect(element.shadowRoot?.querySelector('.icon')).toBeNull();
 		});
 	});
 
 	describe('heading level', () => {
 		it('should update heading level', async () => {
-			expect(element.shadowRoot?.querySelector('.header')?.tagName).toEqual('H3');
-			element.headingLevel = 4;
+			expect(element.shadowRoot?.querySelector(':first-child')?.tagName).toEqual('H2');
```
YonatanKra, line 79: Long query that repeats itself. Should have its own variable

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +36,65 @@ describe('vwc-accordion', () => {
...
+			expect(element.expandmode).toBe('single');
 		});
 	});
 
 	describe('non multi', () => {
-		it('should only allow one accordion items open at a time', async () => {
-			expect(accordionItem1.open).toBeFalsy();
-			expect(accordionItem2.open).toBeFalsy();
+		// test skipped pending FAST issue #6585 https://github.com/microsoft/fast/issues/6585
+		it.skip('should only allow one accordion items open at a time', async () => {
```
YonatanKra, line 45: Thing is, right now it has an API we expose with the bug.
So... we either document what we want and fix it, or we document how it works right now.
This is the current API (with the bug):
```
it('should allow only one accordion item expanded when set to "single"', async () => {
			element.expandmode = 'single';
			accordionItem1.expanded = true;

			accordionItem1.expanded = true;
			accordionItem2.dispatchEvent(new Event('change'));

			expect(accordionItem1.expanded).toBeFalsy();
			expect(accordionItem2.expanded).toBeTruthy();
		});
```
And, of course, add another test to show that just changing the `expand` prop works differently right now.

The other option would be to fix the bug on our end and when they fix it on their end, remove our fix.

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +36,65 @@ describe('vwc-accordion', () => {
...
-			accordionItem1.open = true;
-			await elementUpdated(element);
-			accordionItem2.open = true;
+			
+			// by default, only the first item is expanded
+			expect(accordionItem1.expanded).toBeTruthy();
+			expect(accordionItem2.expanded).toBeFalsy();
+			
+			// if we expand the second one...
+			accordionItem2.expanded = true;
```
YonatanKra, line 69: This can be simplified IMO:
```
		it('should allow expansions of multiple accordion items when set to "multi"', async () => {
			element.expandmode = 'multi';
			accordionItem1.expanded = true;

			accordionItem2.expanded = true;
			accordionItem2.dispatchEvent(new Event('change'));

			expect(accordionItem1.expanded).toBeTruthy();
			expect(accordionItem2.expanded).toBeTruthy();
		});
		```

**Comment by YonatanKra** [comment]
Another general remark - this is not a refactor; it is a breaking change :)
Refactor is when you change the code without any change to the API.

**Comment by yinonov** [comment]
> Another general remark - this is not a refactor; it is a breaking change :)
> 
> Refactor is when you change the code without any change to the API.

By convention any type can be a breaking change

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

File: libs/components/src/lib/accordion-item/accordion-item.spec.ts (hunk 15 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion-item/accordion-item.spec.ts]: you can't do that since what we're testing here is that the icon is properly inserted in the DOM, making `:nth-child(2)` a different element after you set the prop.
I've rewritten the test to make that clearer though.

File: libs/components/src/lib/accordion-item/accordion-item.spec.ts (hunk 16 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion-item/accordion-item.spec.ts]: same as above: the element changes. made the code more explicit.

File: libs/components/src/lib/accordion-item/accordion-item.spec.ts (hunk 17 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion-item/accordion-item.spec.ts]: different DOM elements again

File: libs/components/src/lib/accordion/accordion.spec.ts (hunk 18 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion/accordion.spec.ts:45]: this is unnecessary throwaway work. the bug is known, documented here and awaiting fix (PR is on its way).

File: libs/components/src/lib/accordion/accordion.spec.ts (hunk 19 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion/accordion.spec.ts:69]: soooo, your simplified version is just going back to the original test that [you deemed not readable enough](https://github.com/Vonage/vivid-3/pull/930#discussion_r1067997389) (albeit without the initial `expect`s)
I've left it as is.

**Comment by olaf-k** [comment]
> Another general remark - this is not a refactor; it is a breaking change :) Refactor is when you change the code without any change to the API.

you'll need to take it up to @yinonov 😁
image

**Comment by YonatanKra** [comment]
> > Another general remark - this is not a refactor; it is a breaking change :)
> > Refactor is when you change the code without any change to the API.
> 
> By convention (if we were to align with conventional commits), any type can be a breaking change
> 
> https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

@yinonov 

Let's clear out definitions: 
Refactoring is defined by a change in code that does not change its functionality. This change, by changing how the component works is by definition not a refactor. 

A breaking change means a change in the API. By changing the API (open => expanded) we exhibit a breaking change.

File: libs/components/src/lib/accordion/accordion.spec.ts (hunk 19 above)
Comment (olaf-k) [reply libs/components/src/lib/accordion/accordion.spec.ts:69]: > Note that changing element.expandmode = 'multi';  to element.expandmode = 'single'; the test is still passing, hence the test itself is wrong.

the test is fine. the fact that the accordion allows multiple children to be open programmatically while in single mode is still the result of [the bug mentioned earlier](https://github.com/microsoft/fast/issues/6585). 

> The test I suggest fails on single and passes on multi :)

it shouldn't.

**Review by YonatanKra — DISMISSED** [review]

File: libs/components/src/lib/accordion-item/accordion-item.spec.ts [hunk 20]
```diff
@@ -23,94 +23,89 @@ describe('vwc-accordion-item', () => {
...
 	describe('no-indicator', () => {
-		it('should remove indicator class', async () => {
-			expect(getBaseElement(element).classList.contains('no-indicator')).toBeFalsy();
+		it('should remove indicator', async () => {
+			const indicatorExistsOnInit = !!element.shadowRoot?.querySelector('.icon');
 			element.noIndicator = true;
 			await elementUpdated(element);
-			expect(getBaseElement(element).classList.contains('no-indicator')).toBeTruthy();
+
+			expect(indicatorExistsOnInit).toBeDefined();
```
olaf-k, line 88: `indicatorExistsOnInit` is a boolean, it is *always* defined.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -23,94 +23,89 @@ describe('vwc-accordion-item', () => {
...
+			itemHeaderButton.click();
+			const isClosedAfterSecondClick = element.expanded;
 
-			element.open = false;
-			await elementUpdated(element);
-			expect(element.open).toBeFalsy();
-			expect(getBaseElement(element).classList.contains('open')).toBeFalsy();
+			expect(initExpandState).toBeFalsy();
+			expect(isOpenAfterFirstClick).toBeTruthy();
+			expect(isClosedAfterSecondClick).toBeFalsy();
```
olaf-k, line 48: how this is supposed to be improve readability over
```
	itemHeaderButton.click();
	expect(element.expanded).toBeTruthy();
	itemHeaderButton.click();
	expect(element.expanded).toBeFalsy();
```
is beyond me.

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -9,6 +9,14 @@ import '.';
 const COMPONENT_TAG = 'vwc-accordion';
 
 describe('vwc-accordion', () => {
+	function triggerAccordionUpdate() {
+		const newItem = document.createElement('vwc-accordion-item') as AccordionItem;
+		element.appendChild(newItem);
+	}
+	function toggleAccordionItem(accordionItem: AccordionItem) {
+		accordionItem.expandbutton.click();
+	}
```
olaf-k, line 18: how is
```
toggleAccordionItem(someItem);
```
preferable to
```
someItem.expandbutton.click();
```
?

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +44,146 @@ describe('vwc-accordion', () => {
...
+
+			expect(item1ExpandedBeforeFocus).toBeFalsy();
+			expect(item2ExpandedBeforeFocus).toBeTruthy();
+
+			expect(accordionItem1.expanded).toBeTruthy();
+			expect(accordionItem2.expanded).toBeFalsy();
 		});
 	});
+
+	it('should set aria-disabled on active item', async function () {
```
olaf-k, line 184:
```suggestion
	it('should set aria-disabled on active item in single mode', async function () {
		element.expandmode = 'single';
```

libs/components/src/lib/accordion/accordion.spec.ts
```diff
@@ -36,62 +44,146 @@ describe('vwc-accordion', () => {
...
+	});
+
+	describe('accordion-item keydown', () => {
+
+		function addTwoMoreItems() {
+			const anotherAccordionItem1 = document.createElement('vwc-accordion-item') as AccordionItem;
+			const anotherAccordionItem2 = document.createElement('vwc-accordion-item') as AccordionItem;
+			element.insertBefore(anotherAccordionItem1, accordionItem2);
+			element.insertBefore(anotherAccordionItem2, accordionItem2);
+			return [anotherAccordionItem1, anotherAccordionItem2];
```
olaf-k, line 122: no need to return an array if you're only using the second item

change/@vonage-vivid-6061f3db-b114-4a23-a457-642b1fec705e.json
```diff
@@ -0,0 +1,7 @@
+{
```
yinonov, line 1: no longer required

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.spec.ts
YonatanKra: Nice catch. I left the old test :)
Thanks!

libs/components/src/lib/accordion/accordion.spec.ts
YonatanKra: It is an implementation detail that can change if we change `accordion-item`'s API at any point in time.
As a reader, you are not interested in the implementation detail, but in the flow itself. If you really want, see what the function is doing.

libs/components/src/lib/accordion-item/accordion-item.spec.ts
YonatanKra: A matter of standardization. 

`Arrange, Act, Assert`

But a very good point. I pointed it out in my article - this `it` was testing two things (expand + close). I split them into two `it`s.

libs/components/src/lib/accordion-item/accordion-item.scss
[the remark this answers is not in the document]
yinonov: implementation struggles shouldn't not affect style preference, of which `only-of-type` shouldn't set a border to itself.
How complicated is it to set such style from the accordion container?

yinonov (changes requested):

libs/components/src/lib/accordion-item/accordion-item.scss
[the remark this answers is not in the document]
olaf-k: > implementation struggles shouldn't not affect style preference, of which `only-of-type` shouldn't set a border to itself. How complicated is it to set such style from the accordion container?

it is not a struggle nor is it complicated. it is a matter of consistency: with our Figma file, with FAST, and with the code (as a pure container, the parent does not style anything).

olaf-k (commented):
> implementation struggles shouldn't not affect style preference, of which only-of-type shouldn't set a border to itself.
How complicated is it to set such style from the accordion container?

see [my points here](https://github.com/Vonage/vivid-3/pull/930#discussion_r1088833178).

what is the added value of putting back the border in the parent?

libs/components/src/lib/accordion/accordion.spec.ts
olaf-k: this does not improve readability.
details are important in tests and "toggle the item", because there are several ways of toggle said item, gives me less information than "click on the item's expand button".

yinonov (approved):

PR #934, opened 2023-01-05
fix(dialog): adding subtitle

rachelbt:


libs/components/src/lib/dialog/README.md
```diff
@@ -201,7 +202,7 @@ Use the footer `slot` in order to add action buttons to the bottom of the dialog
 </style>
 <vwc-dialog open
   headline="Dialog with footer"
-  text="To quote Michael Carini, although we should never apologize for being ourselves, we should apologize for the times that we are not.">
+  subtitle="this is an example of the dialog with stooted buttons inside footer">
```
olaf-k, line 205: I'm guessing `slotted`?

rachelbt: you caught me :) :)

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -102,17 +103,18 @@ export const DialogTemplate: (
 								<slot name="graphic">
 									${when(x => x.icon, icon(iconTag))}
 								</slot>
-								${when(x => x.headline, headline())}
+								<div class="headline-text">
```
olaf-k, line 106: maybe wrap the `div` in a `when(x.headline || x.subtitle)` to skip its rendering when none of them are set

libs/components/src/lib/dialog/ui.test.ts
```diff
@@ -8,7 +8,7 @@ import type {Dialog} from './dialog';
 
 const components = ['dialog'];
 
-test('should show the component', async ({ page }: { page: Page }) => {
+test.only('should show the component', async ({ page }: { page: Page }) => {
```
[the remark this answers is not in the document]
rachelbt: :scream: shame on me!!

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -62,23 +62,32 @@
...
 
+
 			.icon {
 				font-size: #{size.$vvd-size-condensed};
 			}
 		}
 
 		.base.icon-placement-side & {
-			column-gap: calc(#{size.$vvd-size-condensed} - #{size.$vvd-size-super-condensed});
+			column-gap: calc(#{size.$vvd-size-condensed} / 2);
```
yinonov, line 72: doesn't scale
in default density it changes the value from 8 to 16
in `-1` density it changes 4 to 14

that's a wild card

yinonov (changes requested):

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -62,23 +62,32 @@
...
 				font-size: #{size.$vvd-size-extra-condensed};
 			}
 		}
 	}
+
+	&-text {
+		display: flex;
+		flex-direction: column;
+		column-gap: 4px;
+	}
```
yinonov, line 84: what's that?

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -34,7 +34,7 @@
 	display: flex;
 	flex-direction: column;
 	padding: 24px;
-	gap: 8px;
+	gap: 12px;
```
yinonov, line 37: this raises a question whether it was authored by design. doesn't relate to this scope. can we extract?

libs/components/src/lib/dialog/dialog.scss
rachelbt: so no dividing but subtraction yes?

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -102,17 +115,15 @@ export const DialogTemplate: (
 								<slot name="graphic">
 									${when(x => x.icon, icon(iconTag))}
 								</slot>
-								${when(x => x.headline, headline())}
+								${when(x => x.headline || x.subtitle, headlineText())}
```
yinonov, line 118: why not make them independent from 1 another? can't this be resolved with easily via css?

libs/components/src/lib/dialog/dialog.scss
rachelbt: style for the headline text wrapper that contains the headline + subtitle

yinonov (changes requested):

libs/components/src/lib/dialog/dialog.template.ts
rachelbt: they are independent - but if none of them is inserted by the consumer - no wrapper div is generated

libs/components/src/lib/dialog/dialog.scss
rachelbt: this is a fix - originally didn't give the right gap....

libs/components/src/lib/dialog/dialog.template.ts
yinonov: how complicated would it be to concede the wrapper in favor of a grid?

rachelbt: I considered it but there are too many options that flex is a much better and fluid one, even though we are "paying" with another wrapper

libs/components/src/lib/dialog/dialog.scss
yinonov: subtraction scales with design unit. gap should still be researched later on for whether it should be fluid. but this is yet to be standardized

rachelbt: so should I use definite unit (8px) or leave it as it is?

libs/components/src/lib/dialog/dialog.template.ts
rachelbt: @yinonov  - I changed the html structure of header, remover 2 wrappers, used `grid`

yinonov (approved):

PR #948, opened 2023-01-10
docs(icon): add px to icon docs

rinaok:
Help users to understand icon sizes


olaf-k (approved):
👍

libs/components/src/lib/icon/README.md
```diff
@@ -34,18 +34,18 @@ Use the `size` attribute/property to set the icon's size.
...
-<vwc-icon name="close-line" size="-3"></vwc-icon>
-<vwc-icon name="close-line" size="-2"></vwc-icon>
-<vwc-icon name="close-line" size="-1"></vwc-icon>
-<vwc-icon name="close-line" size="0"></vwc-icon>
-<vwc-icon name="close-line" size="1"></vwc-icon>
-<vwc-icon name="close-line" size="2"></vwc-icon>
-<vwc-icon name="close-line" size="3"></vwc-icon>
-<vwc-icon name="close-line" size="4"></vwc-icon>
-<vwc-icon name="close-line" size="5"></vwc-icon>
+<vwc-icon name="close-line" size="-6"></vwc-icon>16px
```
olaf-k, line 37: shouldn't the measurements be in `em`?

olaf-k, line 37: shouldn't the measurements be in `em`?

olaf-k (changes requested):

libs/components/src/lib/icon/README.md
rinaok: yes, it's just to help users to compare it with figma

olaf-k: answering to myself: they are `px` indeed (so the README should be updated) but their actual size depend on the `--vvd-size-density` value.
image

rachelbt: I think that there are 2 issues here:
1. under size we have `Default: '1em' (inherits font-size)` but I'm not sure that this is clear enough that this is set if theres not size member that is being added.
2.  our consumers are looking for an icon that is 24px width and do not know which `size` is it. @rinaok tried to solve that. but, @olaf-k you are right - if one changer the `vvd-size-density` than the px units are changing....

olaf-k: > 1. under size we have `Default: '1em' (inherits font-size)` but I'm not sure that this is clear enough that this is set if theres not size member that is being added.
indeed. I first thought it was a mistake since `$vvd-size-normal` was not set at 1em, but had to look at the code to understand this was not the default.

rinaok: @olaf-k @rachelbt Do you have any suggestions on how to help users understand the right size for their project? I find it very confusing.

olaf-k: I'd change the README to read:

Use the `size` attribute/property to set the icon's size. Note that changing `--vvd-size-density` will affect the chosen size (the example below shows the size in `px` for the default `--vvd-size-density` value).

- Type: `-6` | `-5` | `-4` | `-3` | `-2` | `-1` | `0` | `1` | `2` | `3` | `4` | `5`
- Default: `undefined` (the icon dimensions will default to `1em`×`1em`)

rinaok: updated

libs/components/src/lib/icon/README.md
```diff
@@ -29,23 +29,24 @@ View list of available icon at the [vivid icons gallery](https://icons.vivid.von
 ### Size
 
 Use the `size` attribute/property to set the icon's size.
+Note that changing `--vvd-size-density` will affect the chosen size (the example below shows the size in px for the its default value).
```
olaf-k, line 32:
```suggestion
Note that changing `--vvd-size-density` will affect the chosen size (the example below shows the size in px for its default value).
```
(the joys of copy/pasting ;)

libs/components/src/lib/icon/README.md
```diff
@@ -29,23 +29,24 @@ View list of available icon at the [vivid icons gallery](https://icons.vivid.von
...
+<vwc-icon name="close-line" size="-4"></vwc-icon>24px
+<vwc-icon name="close-line" size="-3"></vwc-icon>28px
+<vwc-icon name="close-line" size="-2"></vwc-icon>32px
+<vwc-icon name="close-line" size="-1"></vwc-icon>36px
+<vwc-icon name="close-line" size="0"></vwc-icon>40px
+<vwc-icon name="close-line" size="1"></vwc-icon>44px
+<vwc-icon name="close-line" size="2"></vwc-icon>48px
+<vwc-icon name="close-line" size="3"></vwc-icon>52px
+<vwc-icon name="close-line" size="4"></vwc-icon>56px
+<vwc-icon name="close-line" size="5"></vwc-icon>60px
```
yinonov, line 49: what if we scale up / down? size units won't be correct

https://github.com/Vonage/vivid-3/blob/36329cba95612d974f0c03055a7cee0031fe002e/docs/design/design-tokens.md#L133-L140

yinonov (changes requested):

libs/components/src/lib/icon/README.md
rinaok: @yinonov See Olaf's comment

olaf-k (approved):

PR #952, opened 2023-01-11
chore(workspace): remove beachball

olaf-k:
first step.
next (in another PR) is to introduce release-please.

this PR removes references to beachball in actions and documentation. `_publish_package.yml` still works but version is not bumped and changelogs are not recorded anymore (these should be taken care of by release-please).


yinonov:
already addressed in #983

PR #954, opened 2023-01-13
feat(workspace): use release-please

olaf-k:
(forget #952 this PR also takes care of removing beachball.)

# READ ME PLEASE

This PR aims at changing the current Vivid release process (i.e., creating releases and corresponding npm packages) handled with beachball.

It does so by using the [release-please action](https://github.com/google-github-actions/release-please-action).

Every time PR is pushed on main, this action will be triggered. It will collect all the commits made on main related to `/libs/components` and update its own release PR (or create it if there is none) with this list and the next version number.

The next version number is determined by the format of commit titles. They must follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) rules.

To actually create the release, this release PR must be merged. It will then create the GitHub release, publish the npm package and deploy to Deno.

# TODO

- [x] The action takes care of components only. A new workflow is necessary to deploy on Deno for doc-only updates. Done in #982
- [ ] The current flow is not optimized. It favors reusing `_deploy-deno.yml` for readability/maintenance purposes but technically, the Deno deployment could be done in the same step as the npm publication, reusing the same setup.

# FIRST "NEW" RELEASE

We'll need to update `bootstrap-sha` (currently set to an arbitrary value) before the first release. It will determine the commit from which `release-please` will start to scan for stuff to include in the release.

After the first new release, **we'll have to remove `release-as`**. It is currently set to `3.0.0` to avoid starting at `1.0.0`.


yinonov:
Thanks for a thorough description. it's entirely clear what to review

yinonov:
extracted the lint-pr-title action replacement to #956

olaf-k:
> 1. We're probably ok with the change

Great! We might want to fine-tune it if we want to replicate the exact checks we have today (e.g., kebab-case?).
Thanks for #956.

> 2. We never used `CHANGELOG.json` but it was probably generated by beachball

Perfect. I'll take care of the two remaining TODOs and we're good to go!

yinonov:
closes #952

yinonov (dismissed):
responsible pull request 🤝

yinonov (approved):

PR #966, opened 2023-01-19
fix(appearance): ghost and outline states

rachelbt:


rachelbt:
as discussed - the subject was not fully investigated by design.
Some states are missing (like - selected and hover).
Closing the issue until its fully covered by design.

PR #971, opened 2023-01-22
fix(dialog): min inline size variable

rachelbt:
also added the backdrop in this PR (https://github.com/Vonage/vivid-3/issues/972#issuecomment-1399481435)


change/@vonage-vivid-818f9abb-f12b-459f-8b72-8f5abd2be869.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "fix(dialog): min-inline-size variable",
```
yinonov, line 3:
```suggestion
  "comment": "fix(dialog): max-inline-size variable",
```

yinonov (changes requested):

change/@vonage-vivid-818f9abb-f12b-459f-8b72-8f5abd2be869.json
rachelbt, line 3:
```suggestion
  "comment": "fix(dialog): max-inline-size variable and dialog scream color",
```

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -11,23 +11,33 @@
...
 		z-index: var(#{variables.$dialog-z-index}, 1);
 	}
+
+	&.modal {
+		filter: var(#{constants.$vvd-shadow-surface-24dp});
+
+		&::backdrop {
+			background-color: currentColor; // using as current color cause ::backdrop doesn't inherit root variables
+			opacity: 0.5;
+		}
```
yinonov, line 38: use https://github.com/Vonage/vivid-3/blob/main/libs/shared/src/lib/sass/mixins/scrim.scss#L9

yinonov (changes requested):

libs/components/src/lib/dialog/dialog.scss
rachelbt: it doesn't work (tried...)
_It does not inherit from any element and is not inherited from. No restrictions are made on what properties apply to this pseudo-element either._

https://fullscreen.spec.whatwg.org/#::backdrop-pseudo-element

also in SO question:
https://stackoverflow.com/questions/58818299/css-variables-not-working-in-dialogbackdrop/63322762#63322762

I think the use of currentColor is working well here.
I know its kind of redundant code but for now there's no other option for setting the `::backdrop` color

yinonov: I see. then change the mixin

yinonov (approved):

Issue #678, opened 2022-09-12
[light to dark theme]: algorithm research


AyalaBu:
When light theme transforms into dark theme 950 color turns into 0 (always #ffffff).
This presents a problem, because then basically there is no 950 color in dark theme.

![Image](https://user-images.githubusercontent.com/106529909/189627383-b41d3122-9c06-4731-bdd3-d796ec01a431.png)

After some research I came up with a possible solution:

![Image](https://user-images.githubusercontent.com/106529909/189684151-fe893c7f-d205-4f20-be55-5f8f61d27566.png)

You can also see it here:

https://www.figma.com/file/QG5hvuw5qsfS0XYMVJ6V3L/Vivid-UI-Kit---Exploration?node-id=3965%3A50437

@yinonov 
What do you think? 
How problematic is it from development point of view?

yinonov:
semantic color tokens are the layer where you define those references. if the change requires a modification outside the tokens layer and have development conditionally refer tokens base on the theming context, it's a problem.

does this affect the palette? or just semantics tokens?

AyalaBu:
The semantic tokens remain the same.

The thing that changes is "switching mechanism" between light/dark theme.
If we add "1000 neutral" token to the algorithm then 500 token stays the same while switching themes,
white switches to black,
and other colors switch as shown.

The only issue it presents us with is that white on black/Black on white is too much contrast,
and can make users eyes feel uncomfortable.
So take a look at the images I added, these are possible solutions that I thought can be considered for this situation.

And by this method we will also have a 950 in dark theme without any problems.

AyalaBu:
After a conversation with @yinonov  was decided to map the color tokens usage as shown in the add image,
in order to have a better picture how to progress with this issue, 

![Image](https://user-images.githubusercontent.com/106529909/190891194-35959057-f7b4-47af-8ebe-82b9ea350e3f.jpg)

yinonov:
The only manual work and switching should be handled in the semantic token scope (circled)

PR #976, opened 2023-01-23
fix(docs): add 950 color example

rinaok:
waiting for https://github.com/Vonage/vivid-3/issues/678


yinonov:
There are no such tokens

rinaok:
> There are no such tokens 

I remember that @AyalaBu said we should add them

PR #987, opened 2023-01-29
feat(tree): add component

rinaok:
Tree View:

A tree view widget presents a hierarchical list. Any item in the hierarchy may have child items, and items that have children may be expanded or collapsed to show or hide the children. For example, in a file system navigator that uses a tree view to display folders and files, an item representing a folder can be expanded to reveal the contents of the folder, which may be files, folders, or both.

Structure:

1. The vwc-tree-view should have a role=”tree”.
2. The vwc-tree-view-item should have a role=”treeitem”.
3. role="group" is applied to child tree node containers

Behaviour:

1. Clicking on a tree item creates a selection
2. Tree items can only contain text values, no actionable elements, apart from our toggle button, can be placed inside a Tree item.


rinaok:
@AyalaBu Should the `selected` state look like in option:

Screenshot 2023-02-07 at 15 59 05

Or like in tabs:
Screenshot 2023-02-07 at 15 59 39

In Fast it looks like this:
Screenshot 2023-02-07 at 16 00 14

rinaok:
@AyalaBu How should the chevron look like?
I used `chevron-right-solid` and rotated it in 45 degrees.

AyalaBu:
@rinaok 
I looked into the issue you presented here,
and I saw a bit different behavior.

Based on these examples you can use the chevrons we have in store:
https://mui.com/material-ui/react-tree-view/

https://w3c.github.io/aria-practices/examples/treeview/treeview-navigation.html

But also there is this example:
https://w3c.github.io/aria-practices/examples/treeview/treeview-1/treeview-1a.html

I think in a situation like this you can use a "+" and a "-" icons.

libs/components/src/lib/tree-item/tree-item.scss
```diff
@@ -0,0 +1,104 @@
...
+	align-items: center;
+	font-size: 20px;
+}
+
+.items {
+	display: flex;
+	flex-direction: column;
+	gap: 4px;
+	margin-block: 4px;
+	padding-inline: 60px;
```
olaf-k, line 88: I don't know if it's the wanted effect but this adds padding on both side, making the subitems shrink fast and not right-aligned.
image

maybe `padding-inline-start` would be preferable?

rinaok: Yeah actually you are right, only the text should be with padding like this:

Screenshot 2023-02-13 at 16 52 37

rinaok: @olaf-k Can you suggest  it as a change?

libs/components/src/lib/tree-view/tree-view.scss
```diff
@@ -0,0 +1,7 @@
+
+.control {
+	position: relative;
```
olaf-k, line 3: is this needed?

libs/components/src/lib/tree-item/tree-item.scss
```diff
@@ -0,0 +1,104 @@
...
+		pointer-events: none;
+	}
+
+	.text {
+		font: var(#{constants.$vvd-typography-base});
+	}
+}
+
+.expandCollapseButton {
+	position: relative;
```
olaf-k, line 77: is this needed?

libs/components/src/lib/tree-item/tree-item.scss
olaf-k, line 88:
```suggestion
	padding-inline-start: 48px;
```
(I also updated the value to align the start of the child at the end of the parent's expand/collapse indicator)

libs/components/src/lib/tree-item/tree-item.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+	describe('tree-item click', () => {
+		it('should expand when click', async () => {
+			const expandButton = getControlElement(treeItem1).querySelector(ICON_SELECTOR) as Icon;
+			expandButton.click();
+
+			await elementUpdated(treeItem1);
+			expect(treeItem1.getAttribute('aria-expanded')).toEqual('true');
+		});
+
+		it('should not expand when keydown arrow left', async () => {
```
olaf-k, line 98: there's a bunch of keys for which it shouldn't expand 😁
it might be more interesting that it expands/collapses when the proper keys are used?

olaf-k (changes requested):

libs/components/src/lib/tree-item/tree-item.spec.ts
olaf-k: speaking of key navigation: using right/left apparently also navigates inside a tree item. is it the wanted behavior? I'm asking because the FAST component doesn't do that.

libs/components/src/lib/tree-item/tree-item.scss
rinaok: probably not

libs/components/src/lib/tree-item/tree-item.spec.ts
rinaok: @olaf-k You can expand and close the tree view by using right/left. but not navigate up and down. It's the same as in FAST

rinaok: @olaf-k removed the test. There is a test which tests it: "should expand when click"

olaf-k: > @olaf-k You can expand and close the tree view by using right/left. but not navigate up and down. It's the same as in FAST

it's not the same: you *can* navigate with the right/left keys. go to the [tree item slot example](https://vivid-n0yyyx55rys0.deno.dev/components/tree-item#item) and focus on Tree item 1. you can use the right key to open the first item then use it again to move down (and the left key to move up).

> @olaf-k removed the test. There is a test which tests it: "should expand when click". I tried to check if it expands with the keys but I failed to do it :(

Jest can be a pain sometimes. I'll see if I can help tomorrow.

libs/components/src/lib/tree-view/tree-view.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+	let element: TreeView;
+	let treeItem1: TreeItem;
+	let treeItem2: TreeItem;
+
+	beforeEach(async () => {
+		element = (await fixture(
+			`<vwc-tree-view>
+				<${COMPONENT_TAG} id="item1"></${COMPONENT_TAG}>
+				<${COMPONENT_TAG} id="item2"></${COMPONENT_TAG}>
+			</vwc-tree-view>`
```
olaf-k, line 20:
```suggestion
			`<${COMPONENT_TAG}>
				<vwc-tree-item id="item1"></vwc-tree-item>
				<vwc-tree-item id="item2"></vwc-tree-item>
			</${COMPONENT_TAG}>`
```
maybe your test issues come from here? you were testing treeItem1/2 as `tree-item`s when they were `tree-view`s.

olaf-k: (you will also need to update the dependencies)

olaf-k:
@rinaok I took the liberty to fix your tests import and added an example of key testing in `tree-item.spec.ts`

yinonov (dismissed):

yinonov (approved):

yinonov (approved):

PR #1005, opened 2023-02-03
feat(documentation): add a "edit in CodePen" button to samples (VIV-1031)

olaf-k:
initially, the button was only displayed for the Deno doc (as opposed to locally served), but whynotboth.jpg (I still left a fix in `eleventy.js`)

there are other live code providers like [repl.it](https://replit.com/), [CodeSandbox](https://codesandbox.io/) or [StackBlitz](https://developer.stackblitz.com/guides/integration/create-with-sdk), I didn't check them all.

I changed the local edit button and made all of them stand out as `cta` in a move that should be controversial enough to distract future review comments from the actual code.


AyalaBu:
I would suggest to: 
1. change the btns and code background to neutral 50
2. place a separator between the btns & the code

YonatanKra:
> I would suggest to:
> 
> 1. change the btns and code background to neutral 50
> 2. place a separator between the btns & the code
> 
> Screenshot 2023-04-24 at 10 06 35
> 
> Screenshot 2023-04-24 at 10 06 48

@olaf-k - can we do that and merge?

olaf-k:
> @olaf-k - can we do that and merge?

changing the code's background means fiddling with the editor's theme, I can have a look at that and check if that's 1) easy to do and 2) plays well with selection highlighting

also, what about dark theme?

AyalaBu:
It's an algorithm, isn't it?

apps/docs/transformers/code-block-demo/index.js
```diff
@@ -21,25 +21,35 @@ module.exports = function (content, outputPath) {
...
+const renderiFrame = (index, src, content, componentData) => {
+	const deps = componentData.modules
+		.map(m => m.split('/')[4])
+		.join(',');
+
+	return JSDOM.fragment(`
+	<div class="${CBD_CONTAINER}" style="--tooltip-inline-size: auto;">
 		<vwc-card elevation="0">
 			<iframe id="iframe-sample-${index}" src="${src}" class="${CBD_DEMO}" onload=onloadIframe(this) loading="lazy" aria-label="code block preview iframe" slot="main"></iframe>
 			<vwc-action-group appearance="ghost" style="direction: rtl;" slot="main">
```
rachelbt, line 49:
```suggestion
			<vwc-action-group appearance="ghost" style="direction: rtl; background-color: var(--vvd-color-neutral-300)" slot="main">
```

rachelbt (commented):
@olaf-k if we add this comment + add in the main.style to the class `.cbd-live-sample` background-color: var(--vvd-color-neutral-300) + border top. of whatever - wouldn't that work?

apps/docs/transformers/code-block-demo/index.js
olaf-k: it's unfortunately not as simple as that. as I said, changing the code background requires changing colors in codemirror.
I tried brute forcing `var(--vvd-color-neutral-50);`, which appears to work but actually prevents selection from working, because somehow the bg color [is expected to have transparency](https://codemirror.net/examples/styling/):
image

YonatanKra (approved):
I think it is a cool feature and should go in even as is.
@rachelbt , would you like to open a ticket for yourself to try some CSS trick?

PR #1007, opened 2023-02-06
fix(documentation): getting started

rachelbt:
as we talked about:
image

**Still not added:**

- [ ] for mirroring content we can open a separate issue
- [ ] links inside the 2 new docs

**additional PR in the same issue:**
https://github.com/Vonage/vivid-3/pull/940
https://github.com/Vonage/vivid-3/pull/968

PR #1014, opened 2023-02-07
feat(styles): contain css rules within cascade layers

yinonov:
let authors take advantage of the cascade [`@layer statement`](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers#the_layer_statement_at-rule_for_named_layers) and set their preferred layers priority by containing vivid css rules within a specified `@layer`.

- [ ] must clarify conflict with Vivid@2.x tokens - https://github.com/Vonage/vivid-3/discussions/1010#discussioncomment-4887325
  - one feasible solution is to apply `revert-layer` value on conflicting rules, which will revert to a previous layer
  ```css
  .vivid-3-component {
     --vvd-color-neutral-50: revert-layer;
  }
   ```
- [ ] awaits #1007 to add relevant documentation

### DIFF
--- a/libs/styles/src/lib/core/typography.scss
+++ b/libs/styles/src/lib/core/typography.scss
@@ -16,84 +16,87 @@ $selectors: (
 	-moz-osx-font-smoothing: grayscale;  /* apply font anti-aliasing for Firefox on OSX */
 }
 
-.vvd-root {
-	&:root { // <--- When 'vvd-root' is set on the `:root` (html element), typeface sizes are able to descend from the root font-size, thus comply with the [WCAG 1.4.4](https://www.w3.org/WAI/WCAG21/Understanding/resize-text)
-		--vvd-size-font-scale-base: 1rem;
-		font-size: unset; // resolves to the user-agent default font size (generally 16px)
+@layer vivid.typography {
+	.vvd-root {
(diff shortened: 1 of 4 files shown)


libs/styles/src/lib/core/typography.scss
```diff
@@ -16,84 +16,87 @@ $selectors: (
 	-moz-osx-font-smoothing: grayscale;  /* apply font anti-aliasing for Firefox on OSX */
 }
 
-.vvd-root {
-	&:root { // <--- When 'vvd-root' is set on the `:root` (html element), typeface sizes are able to descend from the root font-size, thus comply with the [WCAG 1.4.4](https://www.w3.org/WAI/WCAG21/Understanding/resize-text)
-		--vvd-size-font-scale-base: 1rem;
-		font-size: unset; // resolves to the user-agent default font size (generally 16px)
+@layer vivid.typography {
```
rachelbt, line 19: adding this to a layer means that it will not run over style in the project that vivid-3 is implemented - is that we want? I'm not sure what is the right way to do here... how "intrusive" we should or shouldn't be. 
The thing is - that even with high specificity layered style will not "win" maybe this one need to be left for authors?

yinonov: a library should not enforce its rules over existing app rules. it's up for authors to decide how to prioritize assets. layers simply enable declarative prioritization of style sheet instead of battling through specificity

PR #1020, opened 2023-02-08
docs(components): add radio group example to action group

yinonov:


libs/components/src/lib/action-group/README.md
```diff
@@ -127,6 +127,26 @@ Use `<vwc-divider>` for adding separator between the action elements
 </vwc-action-group>
 ```
 
+### Radio Group
+
+Note: this is a simplified example illustrating a styled radio group and is not intended to be a complete implementation of the [Radio Group Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/).
```
rinaok, line 132:
```suggestion
Note: This is a simplified example illustrating a styled radio group and is not intended to be a complete implementation of the [Radio Group Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/).
```

rinaok (approved):

PR #1022, opened 2023-02-08
fix(documentation): samples iframes theme swapping

olaf-k:
fixes:
- the iframes markup, properly separated between `<head>` and `<body>` (+ BrowserSync config update to not add its script in the iframes)
- a bug that makes theme swapping adding new `<link>` elements indefinitely in the iframe and main page markup
image

edit: as the main page is also affected, some DRY refactoring would be nice at some point.


yinonov (approved):

apps/docs/assets/scripts/theme-handler.js
```diff
@@ -15,9 +15,15 @@ function setTheme(newTheme) {
...
-    document.head?.insertAdjacentHTML("beforeend", themeStyle);
+    const displayMode = isDark() ? 'dark' : 'light';
+    const themeStyle = `<link id="theme-link" rel="stylesheet" href="/assets/styles/tokens/theme-${displayMode}.css" media="all">`;
+  
+    const themeLink = document.head.querySelector('#theme-link');
+    if (themeLink) {
+        themeLink.outerHTML = themeStyle;
+    } else {
+        document.head.insertAdjacentHTML("beforeend", themeStyle);
+    }
```
yinonov, line 26: refactoring this in #1024. can you please revert?

yinonov (changes requested):

apps/docs/assets/scripts/theme-handler.js
olaf-k: done.
keep in mind that it's the same bug / fix as for iframes, so unless you change the mechanism you'll need a similar code.

yinonov (approved):

PR #1024, opened 2023-02-08
refactor(docs): refactor theme selection to use menu

yinonov:


rachelbt (commented):
the code looks OK but visually why do we need the radio buttons?

yinonov:
> the code looks OK but visually why do we need the radio buttons?

semantically (and behaviorally), this should be a radio item as only 1 item is allowed to be checked.

this PR is using the Vivid menu existing API. if it lacks reasonable options, it lacks the same for consumers of the library

not being able to override the `menu-item-radio`'s leading icon, that's a design issue to be addressed by @AyalaBu

AyalaBu:
After some research I realized that for some reason this type of behavior is not very common.
I found only two examples:

**Atlassian:**
Link:
https://atlassian.design/components/dropdown-menu/examples#open

Screenshot:
Screenshot 2023-02-12 at 11 02 11

**Github:**
Link:
https://github.com/microsoft/microsoft-ui-xaml/issues/173

Screenshot:
Screenshot 2023-02-12 at 11 04 00

Personally I would prefer the second solution, 
because as with the checked menu item,
when there in just an empty space in a component that doesn't have too much of a free space anyway,
it look way nicer and much more neat and esthetic.

So what do you guys think?
@yinonov @olaf-k @rachelbt @rinaok

rinaok:
> After some research I realized that for some reason this type of behavior is not very common. I found only two examples:
> 
> **Atlassian:** Link: [atlassian.design/components/dropdown-menu/examples#open](https://atlassian.design/components/dropdown-menu/examples#open)
> 
> Screenshot: Screenshot 2023-02-12 at 11 02 11
> 
> **Github:** Link: [microsoft/microsoft-ui-xaml#173](https://github.com/microsoft/microsoft-ui-xaml/issues/173)
> 
> Screenshot: Screenshot 2023-02-12 at 11 04 00
> 
> Personally I would prefer the second solution, because as with the checked menu item, when there in just an empty space in a component that doesn't have too much of a free space anyway, it look way nicer and much more neat and esthetic.
> 
> So what do you guys think? @yinonov @olaf-k @rachelbt @rinaok

@AyalaBu What do you mean in "the second option"?

yinonov:
> zed that for some reason this type of behavior is not very common.
> I found only two examples:

unrelated to this PR but sure, when it'll be added to figma, we will align
https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?node-id=19119%3A84374&t=IIHevjANUOEBfxpw-1

AyalaBu:
@rinaok 
By second option I referred to the GitHub reference.
I think it looks nicer and doesn't crown any more a already crowded space 👇

Screenshot 2023-02-12 at 11 04 00

olaf-k:
> After some research I realized that for some reason this type of behavior is not very common. I found only two examples

as a Windows user I find the behavior quite common. these days, lots of MS apps tend to rely on a "ribbon" display so these menus are less prevalent in the MS apps or the OS itself, but many others still use checkbox and radio-like choices in menus.

the look'n'feel however is inconsistent.
some, such as 7-zip,  clearly state the different between radio vs checkbox:
image
while others, like Acrobat, don't bother and use a checkmark indicator in both cases:
image
image

yinonov:
> > After some research I realized that for some reason this type of behavior is not very common. I found only two examples
> 
> as a Windows user I find the behavior quite common. these days, lots of MS apps tend to rely on a "ribbon" display so these menus are less prevalent in the MS apps or the OS itself, but many others still use checkbox and radio-like choices in menus.
> 
> the look'n'feel however is inconsistent. some, such as 7-zip, clearly state the different between radio vs checkbox: image while others, like Acrobat, don't bother and use a checkmark indicator in both cases: image image

good input for design to consider @AyalaBu

AyalaBu:
After having multiple discussions with Jay and designers this is the conclusion that was reached:
In some cases (primarily menu item) there is need for a regular check mark,
not a checkbox or a radio button.

Also this behavior appears in multiple design systems.

The design will be provided asap.

PR #1027, opened 2023-02-13
fix(documentation): repeated icon note

olaf-k:
there is an a11y note warning authors for using only an icon repeated in several components. ~~this PR replaces it with a small include.~~
This PR:
+ adds the note to `accordion-item` and the `option`.
+ corrects the `badge` note (which was truncated) and the card one (which was wrong).
+ removes the "Icon only" paragraph of `nav-disclosure` for consistency (we don't do it for other widgets).


apps/docs/_includes/icon-note.md
```diff
@@ -0,0 +1 @@
+Note: An icon on its own doesn't make a discernible text. An `aria-label`, `aria-labelledby` or `title` must be provided to ensure that the user can understand the component's purpose.
```
rachelbt, line 1: When the icon is followed by with text and has no real need for screen readers than authors should add `aria-hidden="true"`

rachelbt (commented):
nice docs maintaining touch-up!
How come there are new snapshots for the nav-discloser? seems not relevant to this PR...

apps/docs/_includes/icon-note.md
olaf-k: true, but that's valid for every bit of text in an app, not just when using an icon.

olaf-k:
> nice docs maintaining touch-up! How come there are new snapshots for the nav-discloser? seems not relevant to this PR...

you're right, this should have been a doc only PR. but since I updated the `nav-disclosure` doc to *remove* a paragraph, it shows in the ui tests!

apps/docs/_includes/icon-note.md
rachelbt: not sure I understand..

rachelbt:
> > nice docs maintaining touch-up! How come there are new snapshots for the nav-discloser? seems not relevant to this PR...
> 
> you're right, this should have been a doc only PR. but since I updated the `nav-disclosure` doc to _remove_ a paragraph, it shows in the ui tests!

why did you removed it? don't we have an icon-only option?

olaf-k:
> why did you removed it? don't we have an icon-only option?

this was the `nav-disclosure` comment for the paragraph I removed:
image
as I wrote, I did it for consistency purposes: there are other widgets that allow authors to inject an icon while not providing text/label (accordion-item, badge, button, etc.) and we don't provide a specific example for them.

apps/docs/_includes/icon-note.md
yinonov: For SVG, If no explicit role/aria set, it's hidden by default. Is it not?

yinonov:
> Is this supported in GH markdowns?

No
image

apps/docs/_includes/icon-note.md
rachelbt: no.....

yinonov:
> > > Is this supported in GH markdowns?
> 
> > 
> 
> > No 
> 
> 
> 
> of course, that's nunjucks syntax, markdown is not a templating language. is that a dealbreaker? do you expect users to read the doc on GH?
> 
> 

Yes. Otherwise we would have wrote directly in 11ty templates

olaf-k:
> Yes. Otherwise we would have wrote directly in 11ty templates

that's (partly) the point of this PR: we have repeating information and a templating engine, why not take advantage of it?

apps/docs/_includes/icon-note.md
yinonov: > When the icon is followed by with text and has no real need for screen readers than authors should add `aria-hidden="true"`

the sentence starts with _"An icon on its own..."_

yinonov:
> > Yes. Otherwise we would have wrote directly in 11ty templates
> 
> that's (partly) the point of this PR: we have repeating information and a templating engine, why not take advantage of it?

because you lose documentation in GitHub. the closest you can get to mirroring content in GH is via permalinks. where in 11ty this type of url should be intercepted and replaced with the actual mirrored content. but that windowed content can still be somewhat confusing

https://github.com/Vonage/vivid-3/blob/ec2e81920b18d39d982492f76856390bf4fb0be0/docs/design/design-tokens.md?plain=1#L6-L12

olaf-k:
tigger-sad

olaf-k:
include reverted and replaced by copy/paste, first post updated.

yinonov (changes requested):
this caveat would make more sense when the documentation flow in this order

1. text / label section
2. icon section (showing icon only without text / label). in the icon section it makes sense to explain about the caveat sense it's a real case
3. icon with text / label

also bear in mind to apply those same rules in the code snippet

PR #1030, opened 2023-02-15
feat(documentation): live samples

olaf-k:


yinonov (changes requested):

apps/docs/transformers/code-block-demo/index.js
```diff
@@ -15,111 +12,94 @@ const CBD_DEMO = 'cbd-demo';
...
       <details class="${CBD_DETAILS}" slot="main">
         <summary></summary>
-        <div class="${CBD_CODE_BLOCK}" role="region" id="${codeBlockId}">
-          ${demoData.codeStr}
-        </div>
+		<div class="cbd-live-sample" data-index="${index}" role="region">
+			${pre.outerHTML}
+		</div>
+		<div style="display:flex; align-items:center; justify-content:flex-end; padding:5px">
+			Ctrl-Enter or
```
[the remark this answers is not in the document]
olaf-k: for two reasons:
- personal: I don't really like editors constantly reloading when I'm in the middle of coding, like CodePen does. I find it annoying (yes there's an error OF COURSE I AM NOT DONE YET) and distracting. but that's personal, of course.
- technical: it's just more complicated. this needs as many `setInterval` as there are samples and needs debounce as you said. the manual reload is just a function call 😁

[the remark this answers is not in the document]
yinonov: dev-tools takes a different approach of submission on blur. how about that to reduce complexity?

yinonov:
> * Some samples will not update properly because they use `const`: refreshing the code is done in the same context (needs samples editing).

how complex would it be to fully reload? that's a common switch in sandboxes to disable HMR

apps/docs/transformers/code-block-demo/index.js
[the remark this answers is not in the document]
olaf-k: I can have a look! the codemirror doc is not as user-friendly as I'd hoped so that's where the main complexity resides at the moment unfortunately.

olaf-k:
> how complex would it be to fully reload? that's a common switch in sandboxes to disable HMR

I don't really know at this point. I could *replace* the entire document like I did in the [load-samples-runtime branch](https://github.com/Vonage/vivid-3/compare/main...load-samples-runtime), but only after the samples are loaded server-side first. I don't know if that would clean the JS environment though? to "reload" fully, the only other option I can think of is `srcdoc`.

apps/docs/assets/scripts/live-sample.js
```diff
@@ -0,0 +1,99 @@
...
+	return true;
+}
+
+function addSamplesEditors() {
+	const codeBlocks = document.querySelectorAll('.cbd-live-sample');
+
+	codeBlocks.forEach(cbd => {
+		const code = cbd.textContent.trim();
+		cbd.innerHTML = '';
+		const idx = +cbd.dataset.index;
```
YonatanKra, line 45: Is that instead of writing `Number(cbd.dataset.index);`

apps/docs/assets/scripts/live-sample.js
```diff
@@ -0,0 +1,99 @@
...
+			parent: cbd,
+			root: window.document
+		});
+
+		samplesEditors.set(idx, {
+			view,
+			iframe
+		});
+
+		setEditorsTheme();
```
YonatanKra, line 67: Should this run on every iteration, or is it enough for it to run only once everything is up to date?

apps/docs/assets/scripts/live-sample.js
olaf-k: I'm not big on shorthand in general but this one is so common I didn't really pay attention. (eslint doesn't yell if you do that in a component.)

apps/docs/assets/scripts/live-sample.js
olaf-k: 🤦‍♂️good catch!

rachelbt:
didn't look at the code - love the end result.
I'm wondering though if its obvious that devs can change. I wouldn't think of it, but I'm not the user :)

olaf-k:
> didn't look at the code - love the end result. I'm wondering though if its obvious that devs can change. I wouldn't think of it, but I'm not the user :)

that's a good remark. do you have suggestions as to how to advertise this feature to a doc user?

rinaok:
> > didn't look at the code - love the end result. I'm wondering though if its obvious that devs can change. I wouldn't think of it, but I'm not the user :)
> 
> that's a good remark. do you have suggestions as to how to advertise this feature to a doc user?

@olaf-k Maybe we can add `edit`/`copy` tooltips to the buttons?

olaf-k:
tooltips are a good idea yes, I never noticed we didn't have any.
to make it perfectly obvious, what about changing the code display icon with a label (this one or another)?
image
wdyt @rachelbt @rinaok ?

rinaok:
@olaf-k What happened to the code blocks? They are stretched now on the whole page

apps/docs/transformers/code-block-demo/index.js
```diff
@@ -19,111 +16,90 @@ const CBD_DEMO = 'cbd-demo';
...
+			</vwc-action-group>
+			<details class="${CBD_DETAILS}" slot="main">
+				<summary></summary>
+				<div class="cbd-live-sample" data-index="${index}" role="region">
+					${content}
+				</div>
+			</details>
+		</vwc-card>
+		<vwc-tooltip anchor="buttonEdit${index}" text="Edit source code" placement="top-start"></vwc-tooltip>
+		<vwc-tooltip anchor="buttonCopy${index}" text="Copy source code" placement="top-start"></vwc-tooltip>
```
rinaok, line 52:
```suggestion
		<vwc-tooltip anchor="buttonEdit${index}" text="Edit" placement="top-start"></vwc-tooltip>
		<vwc-tooltip anchor="buttonCopy${index}" text="Copy" placement="top-start"></vwc-tooltip>
```

@olaf-k Maybe just copy and edit to make it smaller? what do you think?

olaf-k:
> @olaf-k What happened to the code blocks? They are stretched now on the whole page

oops didn't notice because of the devtools, thanks.

apps/docs/transformers/code-block-demo/index.js
olaf-k: let's settle on "Edit code" and "Copy code" as in your example. short and clear.

rinaok:
Screenshot 2023-03-01 at 14 42 28

The edit text is not in the middle. @olaf-k Can you see it?

olaf-k:
> The edit text is not in the middle. @olaf-k Can you see it?
all tooltips are left-aligned, there is actually no option to change that. (well, there would be but that's another PR.)

rinaok (dismissed):

olaf-k:
note (to self mainly 😬): had to slightly alter the eleventy process to make sure the bundled script is not overwritten. the docs build should be working fine now (including during `npm start`).

rinaok (approved):

PR #1036, opened 2023-02-20
feat(data-grid): focus on cell remove header rows from tabfocus queue

### LINKED ISSUE #994: Focus on cell: Remove header rows from tab/focus queue

YonatanKra:
Closes #994

### DIFF
--- a/libs/components/src/lib/data-grid/data-grid-cell.spec.ts
+++ b/libs/components/src/lib/data-grid/data-grid-cell.spec.ts
@@ -83,6 +83,21 @@ describe('vwc-data-grid-cell', () => {
 			expect(element.querySelector('#cell')).toBeNull();
 			expect(element.innerText).toBeUndefined();
 		});
+
+		it('should render cell focus template when not columnheader', async function () {
+			setCellDataAndConfig();
+
+			element.cellType = 'rowheader';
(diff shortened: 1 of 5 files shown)


olaf-k:
I'm fine with the code changes, but what about a11y? Headers are not read by ATs anymore it seems (quickly tested with VoiceOver), that seems like a regression.

YonatanKra:
> VoiceOver

@olaf-k I think it is the same as the spec here: https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/
Do you see the same effect?

YonatanKra:
@olaf-k thanks to you, the issue was is going to be discussed in the next W3C meeting: https://github.com/w3c/aria-practices/issues/2626#issuecomment-1445612824

Closing this now as they mention they are going for focusable headers as the preferred way to do it.

@yinonov - any final words?

olaf-k:
> Closing this now as they mentioned they are going for focusable headers as the preferred way to do it.

your aria markup looks fine so I did a bit more testing and realized that the column headers are properly read when you navigate from cell to cell *using the AT controls* but not when using the arrow keys.
so, focusable headers or not, I'm fine as long as there is a way to know where you are in the table.

PR #1044, opened 2023-02-27
fix(components): proper role for menu anchor

### LINKED ISSUE #1021: [menu button]: buttons triggering menu miss partial implementation
### Description

> Menu button: as described in the [menu button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/), a button is revealed to assistive technologies as a menu button if it has the property [aria-haspopup](https://w3c.github.io/aria/#aria-haspopup) set to either menu or true.

Our menu docs misinform assistive technologies to button triggering menu intent.
follow [menu button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) to complete pattern requirements

### Component(s) impacted

menu, menu-item

### Browser

_No response_

### Vivid Web Components version

v3.1.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

none

### CodeSandbox example

no need

### Steps to reproduce the issue (if applicable)

_No response_

### Release date (if applicable)

_No response_

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/vonage/vivid-3/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/vonage/vivid-3/issues) for duplicate issues

olaf-k:
fixes #1021

### DIFF
--- a/libs/components/src/lib/menu/menu.ts
+++ b/libs/components/src/lib/menu/menu.ts
@@ -38,6 +38,13 @@ export class Menu extends FastMenu {
 	 */
 	@attr anchor?: string;
 
+	anchorChanged(prevAnchor: string, newAnchor: string ) {
+		const prevAnchorEl = document.getElementById(prevAnchor);
+		const newAnchorEl = document.getElementById(newAnchor);
+		prevAnchorEl?.removeAttribute('aria-haspopup');
+		newAnchorEl?.setAttribute('aria-haspopup', 'menu');
+	}
+
 	popupOpenChanged = () => {
 		this.open = (this._popup as Popup).open;
 	};


libs/components/src/lib/menu/menu.ts
```diff
@@ -38,6 +38,13 @@ export class Menu extends FastMenu {
 	 */
 	@attr anchor?: string;
 
+	anchorChanged(prevAnchor: string, newAnchor: string ) {
+		const prevAnchorEl = document.getElementById(prevAnchor);
+		const newAnchorEl = document.getElementById(newAnchor);
+		prevAnchorEl?.removeAttribute('aria-haspopup');
+		newAnchorEl?.setAttribute('aria-haspopup', 'menu');
```
yinonov, line 45: It's nice but I wonder what pitfalls this might introduce

olaf-k: technically, it's a naive implementation that doesn't manipulate the popup's anchor directly, so it is pretty safe.
functionally, the only side-effect I can foresee are:
- `aria-haspopup` shouldn't be removed because it's been set another way, or
- `aria-haspopup` shouldn't be set because it already has a different value
and both look like code mixup.
do you anything else in mind?

rinaok, line 45: @olaf-k I don't think that this is the solution.
vwc-menu doesn’t have id by default, so how can we setAttribute('aria-haspopup', 'menu') if menu doesn't have an id / have a different id.
The solution has to be adding aria-haspopup in the the examples and documented in the docs.
What do you think?

rinaok (changes requested):

rinaok (approved):

PR #1045, opened 2023-02-28
chore(option): add value example

rinaok:


olaf-k:
ok, documenting `value` is a good initiative indeed, but for the sake of consistency I'd have 2 comments:
1. only add one example with `value`, as we do in other pages (1 attribute <-> 1 example), not in all of them.
2. extend the PR to make sure all form elements are covered (I'm thinking `checkbox` and `radio`; `switch` already has an entry, maybe I'm missing others?)

rinaok:
> ok, documenting `value` is a good initiative indeed, but for the sake of consistency I'd have 2 comments:
> 
> 1. only add one example with `value`, as we do in other pages (1 attribute <-> 1 example), not in all of them.
> 2. extend the PR to make sure all form elements are covered (I'm thinking `checkbox` and `radio`; `switch` already has an entry, maybe I'm missing others?)

@olaf-k done

olaf-k:
> @olaf-k done

you missed the `slider`, it was a test 
SasEvilGIF

olaf-k (commented):
👍

olaf-k (approved):

libs/components/src/lib/slider/README.md
```diff
@@ -100,3 +100,14 @@ Toggle the `disabled` member to disable/enable the slider.
...
+### Value
+
+Use `value` in order to set the value of the switch.
+
+- Type: `string`
+- Default: `on`
+
+```html
+<vwc-switch value="my-value"></vwc-switch>
+```
```
olaf-k, line 113:
```suggestion
Use `value` in order to set the value of the slider.

- Type: `string`
- Default: `on`

```html
<vwc-slider value="5"></vwc-slider>
```
```

**Review by olaf-k — CHANGES_REQUESTED** [review]

File: libs/components/src/lib/slider/README.md [hunk 2]
```diff
@@ -100,3 +100,13 @@ Toggle the `disabled` member to disable/enable the slider.
 ```html preview blocks
 <vwc-slider disabled></vwc-slider>
 ```
+
+### Value
+
+Use `value` in order to set the value of the slider.
+
+- Type: `string`
+- Default: `on`
```
olaf-k, line 109:
```suggestion
- Default: `5`
```
(missed that earlier)

olaf-k (changes requested):

olaf-k (approved):

PR #1049, opened 2023-03-01
fix(select): options width and selected-value ellipsis

rachelbt:


libs/components/src/lib/select/README.md
```diff
@@ -21,9 +21,9 @@ Read more about [vwc-option](../../components/option).
   }
 </style>
 <vwc-select label="choose one option">
-  <vwc-option value="1" text="Option 1" selected></vwc-option>
-  <vwc-option value="2" text="Option 2"></vwc-option>
-  <vwc-option value="3" text="Option 3"></vwc-option>
+  <vwc-option value="1" text="Option 1: dogs" selected></vwc-option>
+  <vwc-option value="2" text="Option 2: cats"></vwc-option>
+  <vwc-option value="3" text="Option 3: doges and catas"></vwc-option>
```
rachelbt, line 26:
```suggestion
  <vwc-option value="3" text="Option 3: dogs and cats"></vwc-option>
```

rinaok (approved):

libs/components/src/lib/select/README.md
```diff
@@ -21,9 +21,9 @@ Read more about [vwc-option](../../components/option).
   }
 </style>
 <vwc-select label="choose one option">
-  <vwc-option value="1" text="Option 1" selected></vwc-option>
-  <vwc-option value="2" text="Option 2"></vwc-option>
-  <vwc-option value="3" text="Option 3"></vwc-option>
+  <vwc-option value="1" text="Option 1: dogs" selected></vwc-option>
```
YonatanKra, line 24: Why the change?

YonatanKra (approved):

libs/components/src/lib/select/README.md
rachelbt: for seeing a better diff in text-width when selected

rinaok: @rachelbt I can't see this dogs text in the snapshots

PR #1066, opened 2023-03-10
fix(public-contract): export public contract constant types

k-paxian:
Typescript constant types used/referenced in element properties/events/etc. types **must** be exported by the package to support consistency in the custom elements public contract. 

Those types needed to be referenced in the end user applications business code to make usage of Vivid components/elements fully typesafe and predictable during compile time.

It's always better to prevent potential bugs sooner than later 😃 

Additionally those types could be used to make fully sound typesafe https://github.com/Vonage/vivid-bindings-vue

### DIFF
--- a/libs/components/src/lib/focus/focus.ts
+++ b/libs/components/src/lib/focus/focus.ts
@@ -1,10 +1,10 @@
 import { FoundationElement } from '@microsoft/fast-foundation';
 
 /**
- * Base class for badge
+ * Base class for focus
  *
  * @public
  */
 export class Focus extends FoundationElement {
-	
+
 }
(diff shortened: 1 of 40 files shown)


k-paxian:
regarding failing e2e test, I'm guessing it's a flaky test but I don't have permissions to re-run it

libs/components/src/lib/focus/focus.ts
```diff
@@ -1,10 +1,10 @@
 import { FoundationElement } from '@microsoft/fast-foundation';
 
 /**
- * Base class for badge
+ * Base class for focus
```
olaf-k, line 4: 👍

olaf-k (approved):
image

PR #1076, opened 2023-03-19
feat(select): add icon

rachelbt:


AyalaBu:
@rachelbt 
Looks great!

libs/components/src/lib/select/README.md
```diff
@@ -19,11 +19,14 @@ Read more about [vwc-option](../../components/option).
   html { /* for demo purposes */
     block-size: 230px;
   }
+  vwc-select {
+    width: 150px;
+  }
```
rinaok, line 24: @rachelbt Why not with a custom property?

rachelbt: because there's no need - it is set on the host and effect all the rest.
We use css-variables when we want to infect a dipper element in the shadow-dom that the consumer can't get to.
There's no need for setting a width, but it can be uses to set as a narrow select or set as 100%

rinaok (dismissed):

olaf-k (changes requested):

libs/components/src/lib/select/README.md
```diff
@@ -49,6 +52,27 @@ Add a `label` attribute to add label to the Select.
...
+- Default: `undefined`
+
+```html preview
+<style>
+  html { /* for demo purposes */
+    block-size: 230px;
+  }
+  vwc-select {width: 150px;}
+</style>
+<vwc-select label="phone" icon="search-line">
```
olaf-k, line 69:
```suggestion
<vwc-select label="choose one option" icon="search-line">
```

libs/components/src/lib/select/README.md
```diff
@@ -177,8 +201,54 @@ Default: `fit-content`
 </vwc-select>
 ```
 
+### Select-width
```
olaf-k, line 204:
```suggestion
### Width
```

olaf-k (changes requested):

libs/components/src/lib/select/README.md
```diff
@@ -149,12 +173,12 @@ You can add a `label` attribute to the `vwc-option` to set a custom display valu
   <vwc-option label="Custom Label 3" value="3" text="Option 3"></vwc-option>
 </vwc-select>
 ```
-## CSS Variables
+## Dimensions
 
-### Select-height
-If there are many options in the list-box, set a height with the `--select-height`.
+### CSS Variables - Select-height
```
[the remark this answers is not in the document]
rachelbt: but in our docs we point out if its css-variable...

libs/components/src/lib/select/README.md
rachelbt:
```suggestion
<vwc-select label="country code" icon="search-line">
```

libs/components/src/lib/select/README.md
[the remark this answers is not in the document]
olaf-k: I find
> Use `--select-height`

sufficiently clear, but we can drive the nail farther with

> Use the `--select-height` CSS variable

libs/components/src/lib/select/README.md
```diff
@@ -49,6 +52,27 @@ Add a `label` attribute to add label to the Select.
...
+- Default: `undefined`
+
+```html preview
+<style>
+  html { /* for demo purposes */
+    block-size: 230px;
+  }
+  vwc-select {width: 150px;}
+</style>
+<vwc-select label="country code" icon="search-line">
```
olaf-k, line 69: why not stick with "choose one option"? these options are not country codes + consistency.

olaf-k (changes requested):

libs/components/src/lib/select/README.md
rachelbt: because the label breaks when we set a width on the select :rolling_on_the_floor_laughing:

libs/components/src/lib/select/README.md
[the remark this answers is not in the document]
rachelbt: I want to stick to pointing out that this is a variable - how about  "height - CSS Variables"

olaf-k (approved):
👍

PR #1088, opened 2023-03-28
feat(components): VIV-881 button pending state

olaf-k:
### Description
3 ways to recreate the Vivid 2 behavior:
1. Adding a slot to let the author add a progress ring component (Vivid 2). This is cumbersome and could be used to insert markup that doesn't make sense UX-wise.
2. Taking advantage of the `icon-trailing` existing property. Probably the simplest solution, but we unfortunately don't have an animated icon matching the display :(
**3. Creating a `pending` attribute that disables the button and adds the loading indicator > this PR.**

edit: did a bit of reading and turns out that 2 is not straightforward as support for SMIL is not clear.

### Caveat
- I'm using the smallest progress ring size available, which still looks OK with `size="condensed"` but note that the indicator size remains the same.
- As discussed below, displaying the processing indicator increases the width of the button.
- When used with an icon only button, using pending does not increase the fixed width and icons shrink.
- The button display is broken when `stacked` is used (under investigation).


rachelbt (commented):
if the icon is set as trailing - will the progress ring go on-top of it or after it. if the latest is the case - it can be problematic since it will change the button width. or what if there's no icon at all....
I'm not sure how the pending button is being used :(

acelletti (dismissed):
Looks good

acelletti (approved):

PR #1092, opened 2023-03-31
docs(components): VIV-952 Add `@slot` attribute to Vivid 3 components.

acelletti:


RichardHelm (approved):

libs/components/src/lib/calendar/calendar.ts
```diff
@@ -18,6 +18,13 @@ import { getEventContext } from './helpers/calendar.event-context';
  * Base class for calendar
  *
  * @public
+ * @slot day-0 - Assign elements to corresponding day column using this slot.
+ * @slot day-1 - Assign elements to corresponding day column using this slot.
+ * @slot day-2 - Assign elements to corresponding day column using this slot.
+ * @slot day-3 - Assign elements to corresponding day column using this slot.
+ * @slot day-4 - Assign elements to corresponding day column using this slot.
+ * @slot day-5 - Assign elements to corresponding day column using this slot.
+ * @slot day-6 - Assign elements to corresponding day column using this slot.
```
olaf-k, line 27: note (to self) the calendar template appears to also have a default slot, which is 1) not documented and 2) useless, to be investigated.

olaf-k, line 27: there's also a default slot to embed calendar-event components.

libs/components/src/lib/header/header.ts
```diff
@@ -5,6 +5,7 @@ import { attr } from '@microsoft/fast-element';
  * Base class for header
  *
  * @public
+ * @slot - Default slot.
```
olaf-k, line 8:
```suggestion
 * @slot - Default slot.
 * @slot action-items - Nodes assigned to action-items slot will be set at the end of the header.
 * @slot app-content - Content vertically aligned with header.
```

libs/components/src/lib/tabs/tabs.ts
```diff
@@ -8,6 +8,7 @@ export const ACTIVE_TAB_WIDTH = '--_tabs-active-tab-inline-size';
  * Base class for tabs
  *
  * @public
+ * @slot - Default slot.
```
olaf-k, line 11: not exactly. `tabs` will process children elements as long as they are slotted as `tab` and `tabpanel`.

libs/components/src/lib/text-anchor/text-anchor.ts
```diff
@@ -6,6 +6,7 @@ import { AffixIcon } from '../../shared/patterns/affix';
  * Base class for text-anchor
  *
  * @public
+ * @slot - Default slot.
```
olaf-k, line 9: no such slot + underlying component

olaf-k (changes requested):
missing:

- menu-item > lacks the undocumented `expand-collapse-indicator` (maybe undocumented on purpose? to be checked)
- popup > lacks `default` (underlying component but since elevation is documented...)

libs/components/src/lib/tabs/tabs.ts
acelletti: This is fine, but the default slot still needs declaring so it can be added in the wrappers.

libs/components/src/lib/calendar/calendar.ts
acelletti: Added.

libs/components/src/lib/header/header.ts
acelletti: Done.

libs/components/src/lib/text-anchor/text-anchor.ts
acelletti: I was looking at the underlying component https://www.fast.design/docs/components/anchor/ and it looks like this does provide a default slot.

olaf-k: looking at the Fast components may be misleading as our templates use fewer slots and rely more on attributes.

libs/components/src/lib/tabs/tabs.ts
olaf-k: that's my point: there is no default slot, there are two named slots (`tab` and `tabpanel`).

olaf-k (approved):
The tabs TSDoc change is wrong and describes the wrapper's API rather than the Vivid component's. Approving on @acelletti 's request.

PR #1094, opened 2023-04-02
edit docs

rinaok:


docs/getting-started/advanced.md
```diff
@@ -3,38 +3,34 @@
...
 ```
-
-Scss users can simply use [@forward](https://sass-lang.com/documentation/at-rules/forward).
+**Or**
 ```js
-@forward 'node_modules/@vonage/vivid/styles/[path to file].css';
+// Sets typography related styles
+'node_modules/@vonage/vivid/styles/core/typography.css';
+```
+**Or**
```
rachelbt, line 23:
```suggestion
**Or all if both files are needed**
```

libs/components/README.md
```diff
@@ -35,15 +35,19 @@ For a full list of components and API, explore the [components docs](https://viv
 
 ### Tokens
 
-To **include the tokens**, its css files must be loaded into the project from the `node_modules/@vonage/vivid/styles/tokens` folder.  
+To **include the tokens**, you must load **one** of the following themse:
```
rachelbt, line 38:
```suggestion
To **include the tokens**, you must load **one** of the following themes:
```

rachelbt: not sure we need this line with the _one_ emphasis, when there's a note right after it.
Maybe this is better:
To **include the tokens**, you must load theme css:

rachelbt (approved):
LGTM

PR #1099, opened 2023-04-03
fix(components): focus indicator when textarea is resized (VIV-905)

olaf-k:


rachelbt (changes requested):
I think that we need to keep the focus design also here and no the native one.

olaf-k:
> I think that we need to keep the focus design also here and no the native one.

the difference is subtle but it's here indeed 👍 I'll have a look

libs/components/src/lib/text-area/text-area.scss
```diff
@@ -26,9 +29,9 @@ $low-ink-color: --_low-ink-color;
 	@include connotation.connotation;
 
 	display: inline-grid;
-	width: 100%;
-	gap: 4px;
+	width: inherit;
 	grid-template-columns: 1fr max-content;
```
rachelbt, line 33:
```suggestion
	grid-template-columns: 1fr;
```
not relevant to this. PT but there are no case where there are 2 columns in the grid

libs/components/src/lib/text-area/text-area.scss
```diff
@@ -88,19 +82,8 @@ $low-ink-color: --_low-ink-color;
...
-}
-
-.focus-indicator {
-	#{focus-variables.$focus-stroke-gap-color}: transparent;
-
-	pointer-events: none;
-
-	.fieldset:not(:focus-visible, :focus-within) > & {
-		display: none;
+	&:focus {
```
rachelbt, line 85: should we add TODO: remove once VIV-904 is done?

rachelbt (approved):
LGTM

PR #1107, opened 2023-04-12
feat(css): add Spezia font CSS to library (VIV-989)

acelletti:


libs/styles/project.json
```diff
@@ -23,6 +23,7 @@
 			"configurations": {
 				"assets": {
 					"commands": [
+						"cp -r libs/styles/src/lib/fonts dist/libs/styles"
```
olaf-k, line 26: would probably be more consistent to rename the CSS file with an `.scss` extension and add
```
"sass libs/styles/src/lib/fonts:dist/libs/styles/fonts"
```
to the scss/commands part of the project file.

(not related to this PR but `lintFilePatterns` in this very file should probably target scss files and not css ones)

olaf-k (approved):
maybe @rachelbt can clarify why we have rules for `SpeziaMonoCompleteVariable` and `SpeziaCompleteVariableItalic` which don't seem to be used at first glance?

libs/styles/src/lib/fonts/spezia-variable.scss
```diff
@@ -0,0 +1,22 @@
+@font-face {
+	font-family: SpeziaCompleteVariableUpright;
+	font-stretch: 50% 200%;
+	font-weight: 1 1000;
+	src: url('https://fonts.resources.vonage.com/fonts/v1/Spezia_Web_Complete_Upright.woff2') format('woff2');
+}
+
```
rachelbt, line 7:
```suggestion
@font-face {
 font-family: SpeziaMonoCompleteVariable;
 font-stretch: 50% 200%;
 font-weight: 1 1000;
 src: '');
}
```

rachelbt (changes requested):
why is it better this way?
Its better for performance that the font will be inside the project and not from an outer URL. 
not that important but - if project is using Volta - this causes them to import the font twice.

Also - the mono font face is missing.

libs/styles/src/lib/fonts/spezia-variable.scss
acelletti: > why is it better this way?

Because it provides the boilerplate CSS for import instead of asking people to copy/paste the CSS into their own projects.

acelletti:
> Its better for performance that the font will be inside the project and not from an outer URL.
> not that important but - if project is using Volta - this causes them to import the font twice.

If they already have the font, they can use their own CSS instead of importing the default file.

libs/styles/src/lib/fonts/spezia-variable.scss
acelletti: Added the mono font, nice catch!

acelletti:
> why is it better this way?

Because it provides the boilerplate CSS for import instead of asking people to copy/paste the CSS into their own projects.

docs/getting-started/fonts-and-tokens.md
```diff
@@ -6,67 +6,14 @@ Vivid uses **Montserrat** and **Roboto Mono** Google fonts.
...
- --vvd-typography-base-code: 400 ultra-condensed calc(var(--vvd-size-font-scale-base, 16px) * 0.875)/1.4285714285714286 SpeziaMonoCompleteVariable;
- --vvd-typography-base-condensed: 400 ultra-condensed calc(var(--vvd-size-font-scale-base, 16px) * 0.75)/1.3333333333333333 SpeziaCompleteVariableUpright;
- --vvd-typography-base-condensed-bold: 600 ultra-condensed calc(var(--vvd-size-font-scale-base, 16px) * 0.75)/1.3333333333333333 SpeziaCompleteVariableUpright;
- --vvd-typography-base-extended: 400 ultra-condensed calc(var(--vvd-size-font-scale-base, 16px))/1.5 SpeziaCompleteVariableUpright;
- --vvd-typography-base-extended-bold: 600 ultra-condensed calc(var(--vvd-size-font-scale-base, 16px))/1.5 SpeziaCompleteVariableUpright;
- /* If vivid typography css core style is included in application, setting the '--vvd-size-font-scale-base'
- css variable as derivative will flexibly update font-size by the user preference */
-}
+```js
+'node_modules/@vonage/styles/fonts/spezia-variable.css';
```
rachelbt, line 12:
```suggestion
Add in the project head:
```html
<link rel="stylesheet" href="node_modules/@vonage/styles/fonts/spezia-variable.css" media="all">
```
```

YonatanKra (approved):

PR #1113, opened 2023-04-13
fix(components): text-area resize (VIV-1004)

olaf-k:


libs/components/src/lib/text-area/README.md
```diff
@@ -49,6 +49,20 @@ Set the `value` attribute to set the default value for the text area. Setting th
 <vwc-text-area value="Default Value"></vwc-text-area>
 ```
 
+### Resize
+
+- Type: `none` | `vertical` | `horizontal` | `both`
+- Default: `none`
```
rachelbt, line 55: why none? in native text-area its initially auto which is: both
image

olaf-k: true. I chose to keep Fast's and Vivid 2's default behavior to minimize code impact.

rachelbt: I guess it make sense :)

rachelbt (commented):
sorry for being the one with tedious comments - but why is the border-bottom of FF png's are missing?

olaf-k:
> sorry for being the one with tedious comments - but why is the border-bottom of FF png's are missing?

no worries, it's a fair question. removing the 5px margin for each sample's div display the page properly but samples are then stacked with Chrome/WebKit. increasing it seems to work for all of them.
I have no idea why the default margin we use results in this though :/

rachelbt (approved):

PR #1114, opened 2023-04-13
feat(tag-group): add tag group component (VIV-945)

rinaok:


rachelbt (changes requested):
the ui png's are empty.
You should also might add width & height so the image will not be so big.

libs/components/src/lib/tag-group/tag-group.template.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ *
+ * @param context
+ * @public
+ */
+export const TagGroupTemplate: (context: ElementDefinitionContext) => ViewTemplate<TagGroup> = (context: ElementDefinitionContext) => {
+	return html<TagGroup>`
+		<div class="${getClasses}" role="listbox" aria-orientation="horizontal">
+			<slot ${slotted({ property: 'slottedTags', filter: elements(context.tagFor(Tag)) })}></slot>
```
olaf-k, line 20: what's the purpose of `slottedTags`?

rinaok: removed

PR #1121, opened 2023-04-18
fix(components): slider change event (VIV-985)

olaf-k:


libs/components/src/lib/slider/slider.ts
```diff
@@ -24,22 +25,20 @@ export class Slider extends FastSlider {
 	 */
 	override valueChanged(previous: string, next: string): void {
 		if (this.$fastController.isConnected) {
+			// min/max constraints backported from v3
```
YonatanKra, line 28: I don't understand this comment. 
Can you elaborate please?

YonatanKra (changes requested):

libs/components/src/lib/slider/slider.ts
olaf-k: the PR fix (mix/max constraints) is backported from Foundation v3 (alpha).

YonatanKra (commented):
Mind removing the comments?
They mean nothing to most people.
We have tests that will fail if we upgrade to v3, right?

olaf-k:
> Mind removing the comments? They mean nothing to most people.

I actually do? they are useful to explain why the code is here and why it looks like that. that is a strange request.

> We have tests that will fail if we upgrade to v3, right?

we don't have tests designed to fail when you upgrade Fast's version, no, why would we?

YonatanKra:
> > Mind removing the comments? They mean nothing to most people.
> 
> I actually do? they are useful to explain why the code is here and why it looks like that. that is a strange request.
> 
> > We have tests that will fail if we upgrade to v3, right?
> 
> we don't have tests designed to fail when you upgrade Fast's version, no, why would we?

According to your comments, if we upgrade FAST to v3, the tests are supposed to fail.
Am I correct?

olaf-k:
> According to your comments, if we upgrade FAST to v3, the tests are supposed to fail. Am I correct?

the comments are to remind me/whoever will maintain that:
1) the min/max computation—lacking in the v2 Fast component which we inherit from—is pretty much a copy/paste of the v3. it will not fail upon upgrade but can then be removed as duplicate.
2) the rest of the logic (setThumbPositionForOrientation + change event) is—by opposition to 1—correctly done in the v2 parent class.

YonatanKra:
> > According to your comments, if we upgrade FAST to v3, the tests are supposed to fail. Am I correct?
> 
> the comments are to remind me/whoever will maintain that:
> 
> 1. the min/max computation—lacking in the v2 Fast component which we inherit from—is pretty much a copy/paste of the v3. it will not fail upon upgrade but can then be removed as duplicate.
> 2. the rest of the logic (setThumbPositionForOrientation + change event) is—by opposition to 1—correctly done in the v2 parent class.

@olaf-k 
Anyway - if there's a test covering the `valueChanged` method, I think we can remove these comments. The fact that I (a possible future maintainer) asked what they mean makes the comments rather useless.

olaf-k:
> @olaf-k Anyway - if there's a test covering the `valueChanged` method, I think we can remove these comments. The fact that I (a possible future maintainer) asked what they mean makes the comments rather useless.

you not understanding something doesn't make it useless Yonatan.

YonatanKra:
> > @olaf-k Anyway - if there's a test covering the `valueChanged` method, I think we can remove these comments. The fact that I (a possible future maintainer) asked what they mean makes the comments rather useless.
> 
> you not understanding something doesn't make it useless Yonatan.

I agree. It doesn't make it that way, it hints on it. Need to make some more experiments :)
Can we ditch the comments and merge this?

PR #1122, opened 2023-04-18
fix(components): paused progress with `pacific` connotation (VIV-1003)

olaf-k:


libs/components/src/lib/progress/progress.scss
```diff
@@ -70,6 +70,7 @@ $rounded-border-radius: 3px;
 
 	.paused & {
 		background-color: var(#{constants.$vvd-color-neutral-300});
+		background-image: none;
```
rachelbt, line 73:
```suggestion
		background: var(#{constants.$vvd-color-neutral-300});
```

rachelbt (approved):
👍

PR #1125, opened 2023-04-20
fix(calendar): failing test on Windows (VIV-000)

olaf-k:
meant to fix that for a while...

on Windows/Node 19:
`new Intl.DateTimeFormat('en-US', {hour: 'numeric', hour12: true}).format(new Date('2022-12-16T13:00:00.000'));`
is `1 PM`, but
`new Intl.DateTimeFormat(undefined, {hour: 'numeric', hour12: true}).format(new Date('2022-12-16T13:00:00.000'));`
is `1 pm`

which fails the "should display time in 12h format" test equality (that assumes the wrong default locale value).


rachelbt (approved):
👍

PR #1126, opened 2023-04-20
feat(text-area): add a way to manually set text-area in error state (VIV-990)

olaf-k:
pretty much a copy/paste of #1103


libs/components/src/lib/text-area/text-area.spec.ts
```diff
@@ -38,7 +38,7 @@ describe('vwc-text-area', () => {
 	});
 
 	describe('basic', () => {
-		it('should be initialized as a vwc-text-field', async () => {
+		it('should be initialized as a vwc-text-area', async () => {
```
YonatanKra, line 41: :)

libs/components/src/lib/text-area/text-area.spec.ts
```diff
@@ -448,6 +448,61 @@ describe('vwc-text-area', () => {
...
+		});
+
+		it('should display the given error message', async function () {
+			element.errorText = forcedErrorMessage;
+			await elementUpdated(element);
+			const errorElement = element.shadowRoot?.querySelector('.error-message');
+			expect(errorElement !== null).toBeTruthy();
+		});
+
+		it('should replace/restore the current error state, if any, when set/removed', async function () {
```
YonatanKra, line 477: Please split this to two different `it`s.

YonatanKra (changes requested):

libs/components/src/lib/text-area/text-area.spec.ts
olaf-k: we had the same exchange in https://github.com/Vonage/vivid-3/pull/1103#discussion_r1170161786
how would you split the test?

libs/components/src/lib/text-area/text-area.spec.ts
```diff
@@ -448,6 +448,61 @@ describe('vwc-text-area', () => {
...
+			expect(element.validationMessage).toBe(forcedErrorMessage);
+			expect(element.validity.valid).toBeFalsy();
+
+			element.errorText = '';
+			await elementUpdated(element);
+
+			expect(element.validationMessage).toBe(initialErrorMessage);
+			expect(element.validity.valid).toBeFalsy();
+		});
+	});
```
YonatanKra, line 504: Something in the spirit of:
```suggestion
		it('should replace the current error state when set', function () {
		element.required = true;
		setToBlurred();
		element.errorText = forcedErrorMessage;
		await elementUpdated(element);

		expect(element.validationMessage).toBe(forcedErrorMessage);
	});
	
	it('should restore the current error state when removed', async function () {
		element.required = true;
		setToBlurred();
		await elementUpdated(element);
		const initialErrorMessage = element.validationMessage;
		element.errorText = forcedErrorMessage;
		await elementUpdated(element);

		element.errorText = '';
		expect(element.validationMessage).toBe(initialErrorMessage);
	});
```

YonatanKra (approved):
LGTM.
image

PR #1127, opened 2023-04-20
fix(note): align-items start in slot (VIV-1024)

rachelbt:


libs/components/src/lib/note/note.scss
```diff
@@ -40,15 +41,15 @@
 	align-self: flex-start;
```
rachelbt, line 41:
```suggestion

```

libs/components/src/lib/note/note.scss
```diff
@@ -20,12 +20,13 @@
 	@include appearance.appearance;
 
 	display: flex;
-	align-items: stretch;
+	align-items: flex-start;
```
olaf-k, line 23: is this needed? (not a trick question, it looks like removing it doesn't change the top alignment)

rachelbt: any way - regarding this changed - I added this to override the default and removed the `align-self: flex-start;` from the icon

olaf-k (approved):

PR #1129, opened 2023-04-21
fix(tooltip): hide on Escape

olaf-k:
see https://jira.vonage.com/browse/VIV-1017

making `closeOnEscape` private yields the TypeError `Cannot read private member from an object whose class did not declare it`, maybe an `override` bug? don't want to spend too much time investigating anyway.

### DIFF
--- a/libs/components/src/lib/tooltip/tooltip.ts
+++ b/libs/components/src/lib/tooltip/tooltip.ts
@@ -56,4 +56,17 @@ export class Tooltip extends Popup {
 	#hide = () => {
 		this.open = false;
 	};
+
+	_closeOnEscape = (e:KeyboardEvent) => {
+		if (e.key === 'Escape') this.#hide();
+	};
+	
+	override openChanged(_: boolean, newValue: boolean): void {
+		super.openChanged(_, newValue);
+		if (newValue) {
+			document.addEventListener('keydown', this._closeOnEscape);
+		} else {
+			document.removeEventListener('keydown', this._closeOnEscape);
+		}
+	}	
 }
--- a/libs/components/src/lib/tooltip/tooltip.spec.ts
+++ b/libs/components/src/lib/tooltip/tooltip.spec.ts
@@ -68,6 +68,25 @@ describe('vwc-tooltip', () => {
 		});
 	});
 
+	describe('escape', () => {
+		it('should disappear when Escape is pressed', async () => {
+			const anchor = await setAnchor();
+			element.anchor = 'anchor';
+			await elementUpdated(element);
+
+			fireEvent(anchor, new MouseEvent('mouseover'));
+			await elementUpdated(element);
+			expect(element.open)
+				.toEqual(true);
+
+			fireEvent(document, new KeyboardEvent('keydown', {key: 'Escape'}));
+			await elementUpdated(element);
+
+			expect(element.open)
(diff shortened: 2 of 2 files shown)


rinaok (changes requested):
It was documented because we had this behavior but removed it to add the polyfill :(  Can you please add it to the popup instead of the tooltip? The behavior should be: When the anchor is focused, ECP should close it. Also, a removeEventListener should be added to disconnectedCallback()

rinaok:
> > It was documented because we had this behavior but removed it to add the polyfill :( Can you please add it to the popup instead of the tooltip? The behavior should be: When the anchor is focused, ECP should close it.
> 
> do we want all components based on popup to inherit this behavior? also it looks like combobox and select already implement it but not menu. also note that in this PR, esc dismisses the tooltip when it is opened (i.e., anchor is focused _or hovered_).
> 
> > Also, a removeEventListener should be added to disconnectedCallback()
> 
> indeed 👍

@olaf-k If you think that not all the components should inherit this behavior, just add the removeEventListener and I'll approve

YonatanKra:
I can't think of a component that uses `popup` that would not need the `escape` button. I think It is an a11y issue. Then again, it might be a YAGNI issue, so we should only extract it to a pattern if there's more than one uses for it. Dialog has its own logic as it uses the native dialog, but this pattern can also apply to it as it only changes the `open` state IMO (I didn't heavily research it, it can be a new JIRA issue in itself to research if a pattern is needed and create it).

Regarding the public function, I opened [an issue in Fast](https://github.com/microsoft/fast/issues/6716). Hopefully, it could be resolved.

Your "hack" is more performant than adding a listener in the connected callback because that would mean we'll have a live listener for every tooltip. Then again, we need to weigh in we have an exposed method, and removing it would be a breaking change (in all likelihood, nobody will use it, but as a principle, the moment it is exposed, it might break something that's using it).

YonatanKra (changes requested):

libs/components/src/lib/tooltip/tooltip.ts
```diff
@@ -56,4 +57,18 @@ export class Tooltip extends Popup {
...
 		this.open = false;
 	};
+
+	#closeOnEscape = (e:KeyboardEvent) => {
+		if (e.key === 'Escape') this.#hide();
+	};
+
+	override openChanged(oldValue: boolean, newValue: boolean): void {
+		super.openChanged(oldValue, newValue);
+		if (oldValue === undefined) return;
```
YonatanKra, line 67: @olaf-k 
Thanks to a hint from the Fast team, I managed to make the change event work with a private function.
I pushed it here. Let me know if this works for you.

olaf-k: yeah basically `openChanged` is called in the constructor during instantiation and my understanding is that `override` mistakenly believes it is in the Popup context (which doesn't have a `#hide` method).

any hack that will shortcut this first call will work, but is still a hack, which I didn't like. but I don't have a strong opinion on this, so if you think it's worth it I can commit it.

YonatanKra: Better an implementation detail hack than an exposed method hack ;)

YonatanKra (approved):

PR #1130, opened 2023-04-21
fix(tooltip): allow setting anchor to an HTMLElement (VIV-1032)

RichardHelm:


libs/components/src/lib/tooltip/tooltip.template.ts
```diff
@@ -28,12 +28,14 @@ export const TooltipTemplate: (
 
 	return html`
 <${popupTag} class="${getClasses}" arrow alternate
-placement=${(x) => x.placement} open=${(x) => x.open} anchor=${(x) => x.anchor}
+	:placement=${(x) => x.placement}
+	:open=${(x) => x.open}
+	:anchor=${(x) => x.anchor}
```
RichardHelm, line 33: This is the actual fix, passing anchor as prop

libs/components/src/lib/tooltip/tooltip.ts
```diff
@@ -1,12 +1,13 @@
...
+import type { Placement, Strategy } from '@floating-ui/dom';
+import { FoundationElement } from '@microsoft/fast-foundation';
 
 /**
  * Base class for tooltip
  *
  * @public
  */
-export class Tooltip extends Popup {
+export class Tooltip extends FoundationElement {
```
RichardHelm, line 10: It doesn't make any sense for tooltip to inherit form Popup so I changed it and instead copied over the relevant attributes.
Similar to how menu works

rinaok (changes requested):
Hi @RichardHelm, I reviewed your changes. Unlike the menu component, which uses the popup inside it, the tooltip component is a private case of the popup. This is why it extends the popup. The popup manages anchor changes and open changes. There is no need to perform the same checks twice in two different components (You can see the UI tests fail now because the open state is not handled). Due to the fact that we didn't know if this was needed by users, we kept it undocumented only for our own use. If you wish, you can use the anchor with an HTMLElement in the tooltip, just add an example to the readme file. :)

rinaok (approved):

PR #1132, opened 2023-04-23
feat(popup): set default z-index of 10 (VIV-1023)

rachelbt:
popup is suppose to be a top layer element.
It had no default `z-index.`
I added `z-index` of 10, thinking that this should be higher than any other element.
However - there are other aspects in `z-index` and stacking context  that may change the hierarchy - in this case a `z-index` can be set on the `:host`.


YonatanKra (approved):
LGTM

PR #1137, opened 2023-04-25
feat(components): toggletip (VIV-865)

olaf-k:


libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -0,0 +1,38 @@
...
+		<${popup}
+			class="${getClasses}"
+			arrow
+			:anchor="${x => x.anchor}"
+			:open="${x => x.open}"
+			?alternate="${x => !x.alternate}"
+			placement="${x => x.placement}"
+			exportparts="vvd-theme-alternate"
+		>
+			<span class="content-wrapper">
```
rachelbt, line 31:
```suggestion
			<div class="content-wrapper">
```
span is an inline-element it shouldn't contain block elements

rachelbt (dismissed):
Nice!!!
But I have some comments :)

I don't know what are the needs for the toggle tip but I think that maybe we should set a different structure:

- Like card, dialog alert - that we provide main slot - so consumers can use vivid comps for the functionality but have their own inside structure.

- Regarding our pre-defined template, maybe we can keep the convention of other components:
header that has our headline, body - slot, footer - slot.
- What about dismiss button? 
-- If we are adding it - is it optional?
Are there any a11y issues that we need to point out in our doc?

libs/components/src/lib/toggletip/toggletip.template.ts
olaf-k: thanks for the comments Rach.

1. sorry I don't think I understand this one though. the toggletip uses the default slot for authors to put whatever structure they want in.
2. I can sure use `header` and `footer`, good point.
3. this was discussed in the [old PR](https://github.com/Vonage/vivid-3/pull/843#issuecomment-1322199844). tl;dr: toggletips don't have a close button, nor does [our design](https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?type=design&node-id=20630-129571&t=H3XXGTfacvmHIejX-0).

olaf-k:
> Nice!!! But I have some comments :)

thanks for the comments Rach.

1. sorry I don't think I understand this one though. the toggletip uses the default slot for authors to put whatever structure they want in.
2. I can sure use `header` and `footer`, good point.
3. this was discussed in the [old PR](https://github.com/Vonage/vivid-3/pull/843#issuecomment-1322199844). tl;dr: toggletips don't have a close button, nor does [our design](https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?type=design&node-id=20630-129571&t=H3XXGTfacvmHIejX-0).

libs/components/src/lib/toggletip/toggletip.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.action-items {
+	display: flex;
+	justify-content: flex-end;
+	gap: 10px;
+}
+
+::slotted([slot=action-items]) {
+	margin-block-start: $padding;
+}
```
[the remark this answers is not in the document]
olaf-k: my initial goal was to only allow buttons in the action-items slot.
I didn't implement the filter, but maybe I should have. the idea is that you really should **not** put anything other than buttons (or maybe links? tbd) there ([see the Figma file](https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/Vivid-UI-Kit---3.0-WIP?type=design&node-id=20630-129571&t=dyyOU4coOjCEH6ku-0)). this is why it's called `action-items` and not just `footer`.

note that, if you *really* want to put your own design, you can just leave the `action-items` slot empty and put whatever crazy design you want in the default one.

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			expect(element).toBeInstanceOf(Toggletip);
+
+			expect(element.open).toBe(false);
+			expect(element.alternate).toBe(false);
+			expect(element.headline).toBeUndefined();
+			expect(element.placement).toBe('right');
+		});
+	});
+
+	describe('open/close', () => {
```
YonatanKra, line 44: Describe the API. There's no `close` property or method.
```suggestion
	describe('open', () => {
```

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			element.open = true;
+			await elementUpdated(element);
+
+			element.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+			await elementUpdated(element);
+
+			expect(element.open).toEqual(true);
+		});
+
+		it('should close when clicked outside', async () => {
```
YonatanKra, line 68:
```suggestion
		it('should set open to false when clicked outside', async () => {
```

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+	describe('anchor', () => {
+		it('should accept an HTMLElement as anchor', async () => {
+			element.anchor = anchor;
+			await elementUpdated(element);
+
+			anchor.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+			await elementUpdated(element);
+
+			expect(element.open).toEqual(true);
```
YonatanKra, line 100: How does this verify it accepts an HTMLelement as an anchor?

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			expect(element.open).toEqual(false);
+
+			anchor.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+			await elementUpdated(element);
+			expect(element.open).toEqual(false);
+
+			anchor2.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+			await elementUpdated(element);
+			expect(element.open).toEqual(true);
+		});
```
YonatanKra, line 122: There are two functionalities tested. Splitting is advised.

```suggestion
		it('should set listener to new anchor when anchor is changed', async () => {
			const anchor2 = fixture(
				'<vwc-button id="anchorButton2"></vwc-button>', ADD_TEMPLATE_TO_FIXTURE
			) as Button;
	
			element.anchor = 'anchorButton';
			await elementUpdated(element);

			element.anchor = 'anchorButton2';
			await elementUpdated(element);
			element.open = false;

			anchor2.dispatchEvent(new MouseEvent('click', {bubbles: true}));
			await elementUpdated(element);
			expect(element.open).toEqual(true);
		});
		
		it('should remove listener on old anchor when anchor is changed', async () => {
			const anchor2 = fixture(
				'<vwc-button id="anchorButton2"></vwc-button>', ADD_TEMPLATE_TO_FIXTURE
			) as Button;
	
			element.anchor = 'anchorButton';
			await elementUpdated(element);

			element.anchor = 'anchorButton2';
			await elementUpdated(element);
			element.open = false;

			anchor.dispatchEvent(new MouseEvent('click', {bubbles: true}));
			await elementUpdated(element);
			expect(element.open).toEqual(false);
		});
```

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			element.anchor = anchor;
+			await elementUpdated(element);
+
+			expect(anchor.ariaLabel).toEqual(' ; Show more information');
+
+			element.anchor = '';
+			await elementUpdated(element);
+
+			expect(anchor.ariaLabel).toEqual('');
+		});
```
YonatanKra, line 134: Standard is expectations in the end. Usually, a test like this means there are two tests in one. Like this:

```suggestion
		it('should remove the aria-label from old anchor when changed', async () => {
			element.anchor = anchor;
			await elementUpdated(element);

			expect(anchor.ariaLabel).toEqual(' ; Show more information');

			element.anchor = 'anchor2';
			await elementUpdated(element);

			expect(anchor.ariaLabel).toEqual('');
		});
		
		it('should set the aria-label on anchor', async () => {
			element.anchor = anchor;
			await elementUpdated(element);

			expect(anchor.ariaLabel).toEqual(' ; Show more information');
		});
```

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			element.anchor = anchor;
+			await elementUpdated(element);
+
+			expect(anchor.ariaLabel).toEqual(initialLabel + ' ; Show more information');
+
+			element.anchor = '';
+			await elementUpdated(element);
+
+			expect(anchor.ariaLabel).toEqual(initialLabel);
+		});
```
YonatanKra, line 149: Again, there are two `its` in this case:
1. Concatenate " ; Show more information" to existing aria-label
2. Revert to original aria-label when anchor is not longer an anchor.

libs/components/src/lib/toggletip/toggletip.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+	describe('headline', () => {
+		it('should have an headline element only when set', async () => {
+			expect(element.shadowRoot?.querySelectorAll('header.headline')).toHaveLength(0);
+
+			element.headline = 'A title!';
+			await elementUpdated(element);
+
+			expect(element.shadowRoot?.querySelectorAll('header.headline')).toHaveLength(1);
+		});
+	});
```
YonatanKra, line 161: Defer expectations to the end and make it clear without understanding the code what is actually being tested.
 
```suggestion
		it('should have an headline element only when set', async () => {
		  const headlineExistsWithoutHeadlineSet = element.shadowRoot?.querySelectorAll('header.headline').length > 0;

			element.headline = 'A title!';
			await elementUpdated(element);
			const headlineExistsWithHeadlineSet = element.shadowRoot?.querySelectorAll('header.headline').length === 1;

      expect(headlineExistsWithoutHeadlineSet).toEqual(false);
			expect(headlineExistsWithHeadlineSet).toEqual(true);
		});
	});
```

libs/components/src/lib/toggletip/toggletip.ts
```diff
@@ -0,0 +1,112 @@
...
+type anchorType = string | HTMLElement;
+
+/**
+ * Base class for toggletip
+ *
+ * @public
+ */
+export class Toggletip extends FoundationElement {
+
+	anchorEl: HTMLElement | null = null;
```
YonatanKra, line 14: This is an exposed API. Should be tested or be private (private as in `#anchorEl`).

libs/components/src/lib/toggletip/toggletip.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	@attr headline?: string;
+
+	/**
+	 * toggle color scheme
+	 *
+	 * @public
+	 * HTML Attribute: alternate
+	 */
+	@attr({ mode: 'boolean'	}) alternate = false;
```
YonatanKra, line 33: Exposed API,  please add a describe with a test (at least that it's reflected).

libs/components/src/lib/toggletip/toggletip.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	/**
+	 * placement of the toggletip
+	 *
+	 * @public
+	 * HTML Attribute: placement
+	 */
+	@attr({ mode: 'fromView' }) placement?: Placement = 'right';
```
YonatanKra, line 41: Exposed untested API.

libs/components/src/lib/toggletip/toggletip.ts
```diff
@@ -0,0 +1,112 @@
...
+	
+	/**
+	 * id or direct reference to the toggletip's anchor element
+	 *
+	 * @public
+	 * HTML Attribute: anchor
+	 */
+	@attr({ mode: 'fromView' }) anchor: anchorType = '';
```
YonatanKra, line 49: Exposed untested API. Is it even used?

libs/components/src/lib/toggletip/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+<vwc-toggletip anchor="button1" placement="top">top</vwc-toggletip>
+<vwc-toggletip anchor="button1" placement="right">right</vwc-toggletip>
+<vwc-toggletip anchor="button1" placement="bottom">bottom</vwc-toggletip>
+<vwc-toggletip anchor="button1" placement="left">left</vwc-toggletip>
+```
+
+## Slots
+
+### Action Items
```
YonatanKra, line 95: I cannot see it in the ui tests.
It seems like there's only one toggle-tip open in the UI tests.
Can you add something like this (I might have gotten the query selector wrong):
```
for (const toggleTip of await page.locator('[listitem](vwc-toggle-tip:not([open])').all())
  await toggleTip.evaluate(node => node.open = true);
```
To open the toggle tips in the test?
Another case might be to add a closed example to show that without open you can't view anything.

YonatanKra (changes requested):
Great job. Mostly test related comments.

YonatanKra (changes requested):
Great job. Mostly test related comments.

libs/components/src/lib/toggletip/toggletip.spec.ts
olaf-k: by ensuring that clicking the anchor has the expected behavior on the toggletip.

libs/components/src/lib/toggletip/toggletip.spec.ts
olaf-k: having multiple `expect`s in a test is reasonable as long as multiple is understood as "small number" and the test is not long.
the suggestion is longer and less readable.

libs/components/src/lib/toggletip/toggletip.ts
olaf-k: indeed, good point.

libs/components/src/lib/toggletip/toggletip.ts
olaf-k: it is "tested" in `basic`. that's all there is to do as it is purely forwarded.

libs/components/src/lib/toggletip/toggletip.ts
olaf-k: same as previous comment.

libs/components/src/lib/toggletip/toggletip.ts
olaf-k: you just did review `anchor` related tests Yonata...

libs/components/src/lib/toggletip/toggletip.spec.ts
olaf-k: see previous comment.

libs/components/src/lib/toggletip/toggletip.spec.ts
olaf-k: this is not an improvement.

libs/components/src/lib/toggletip/toggletip.spec.ts
YonatanKra: See previous answer :)

libs/components/src/lib/toggletip/toggletip.spec.ts
YonatanKra: To me it is :)

libs/components/src/lib/toggletip/toggletip.ts
YonatanKra: And if I remove `@attr`? Tests still pass, right?

libs/components/src/lib/toggletip/toggletip.ts
YonatanKra: Sorry, you are right. For some reason I thought the prop was `anchorType`. Maybe should consider set the type in capitals to make it clearer it is a type and not a prop name :)

libs/components/src/lib/toggletip/toggletip.spec.ts
olaf-k: "for me"

libs/components/src/lib/toggletip/toggletip.ts
olaf-k: `@attr` is a Fast construct, I'm not testing Fast.

YonatanKra (changes requested):
Can you address this commit (only changed the tests):
https://github.com/Vonage/vivid-3/commit/fcae1768bf9d8ea9a0b0f50116abd53cce36ebee
Thanks :)

YonatanKra (approved):
LGTM

PR #1138, opened 2023-05-01
fix(banner): centering content (VIV-1049)

rachelbt:
fixing the failing ui-test in Safari


libs/components/src/lib/banner/banner.spec.ts
```diff
@@ -109,7 +109,7 @@ describe('vwc-banner', () => {
 		it('should change role to role text', async function () {
 			element.role = 'alert';
 			await elementUpdated(element);
-			const role = element.shadowRoot?.querySelector('.banner--message')
+			const role = element.shadowRoot?.querySelector('.banner-message')
 				?.getAttribute('role');
```
YonatanKra, line 113: Can you extract it to a utility function like `getText`?

YonatanKra (changes requested):
Great fix!
Super simplifying the code.
Just one small comment :)

YonatanKra (approved):

PR #1139, opened 2023-05-01
fix(tooltip): dynamically changing anchor now works as expected (VIV-1045)

YonatanKra:


libs/components/src/lib/tooltip/ui.test.ts
```diff
@@ -5,8 +5,12 @@ import {
...
-const components = ['tooltip','button'];
+const components = ['tooltip'];
 
+function changeAnchor() {
+	const tooltip = document.querySelector('#tooltip5');
+	(tooltip as any).anchor = 'anchor2';
+}
 test('should show the component', async ({ page }: { page: Page }) => {
 	const template = `
 <style>
```
rachelbt, line 16: I think you'll need to change the height in the style to 260px and give it padding top of 20-30px, and of course  regenerate the PNGs :)

rachelbt (dismissed):
LGTM

olaf-k (changes requested):
as discussed, there's a memleak to patch in the popup's `attributeChangedCallback`.
aside from that, the popup changes look fine.

the tooltip I'm not so sure. you had to inject a lot of JS to handle the anchor change to compensate the removal of `:anchor` propagation in the template. why remove it in the first place?

olaf-k (commented):
the [current deployment](https://vivid-n8k0bfvkc0bg.deno.dev/components/tooltip) doesn't work

olaf-k (approved):

olaf-k (approved):

PR #1143, opened 2023-05-03
feat(layout): customized rows and gaps (VIV-1046)

rachelbt:


libs/components/src/lib/layout/README.md
```diff
@@ -58,6 +58,31 @@ Use the `column-spacing` attribute to set the initial preferred `spacing` of a c
 </vwc-layout>
 ```
 
+### Row Spacing
+
+Use the `row-spacing` attribute to set the initial preferred `spacing` of a row.
```
YonatanKra, line 63:
```suggestion
Use the `row-spacing` attribute to set the initial preferred margin between rows.
```
Can you also change this for column spacing please?

libs/components/src/lib/layout/README.md
```diff
@@ -178,23 +203,48 @@ Use the `gutters` attribute to set the initial preferred `margin`.
...
   <vwc-card elevation="2"></vwc-card>
 </vwc-layout>
-<vwc-layout style="--layout-grid-template-columns: repeat(auto-fill, minmax(100px, 1fr));">
+```
+
+### Grid-template-rows
+
+Control the grid-template-rows of the layout by setting `--layout-grid-template-rows`.
+
+- Type: `string`
```
YonatanKra, line 222: `type` should be removed from CSS Variables methinks. Makes no sense IMO.
@olaf-k , WDYT?

libs/components/src/lib/layout/README.md
```diff
@@ -178,23 +203,48 @@ Use the `gutters` attribute to set the initial preferred `margin`.
...
   <vwc-card elevation="2"></vwc-card>
   <vwc-card elevation="2"></vwc-card>
+    <vwc-card elevation="2"></vwc-card>
   <vwc-card elevation="2"></vwc-card>
 </vwc-layout>
-<vwc-layout style="--layout-grid-template-columns: 280px repeat(auto-fill, minmax(100px, 1fr));">
+```
+
+### Grid-gap
+#### Grid-template-row-gap
```
YonatanKra, line 237: Shouldn't this be the name of the `css variable`? 
```suggestion
#### --layout-column-gap
```
I know it's not the same in other components, but I believe this makes it clearer.
Anyway, in other components, the title is without the `-` between words.
I think just writing the actual property is the best, as it makes things clearer. Or at least write it as a title without the `-` so it will not confuse ppl (like me).

YonatanKra (changes requested):
My 2 cents :)

libs/components/src/lib/layout/README.md
rachelbt: it is not margin - it is spacing, or better yet - its `gap`

libs/components/src/lib/layout/README.md
olaf-k: I don't think it should be `string`, because it's not, nor entirely removed, because it is useful to know what you can use.
a simple link to https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows would do the job.

rachelbt: @olaf-k , its a good idea, I'll add the link, 
but I think @YonatanKra is asking in general about our documentation. 
We have "type" to all the places where we have css variable.
Maybe this is a discussion for a separate task :)

rachelbt: https://jira.vonage.com/browse/VIV-1057

olaf-k (changes requested):

libs/components/src/lib/layout/partials/_functions.scss
```diff
@@ -13,3 +13,8 @@
 	}
 	@return $selector;
 }
+
+@function get-selector-without-default($prefix, $key, $all-keys) {
+	$selector: "&#{get-default-selector($all-keys, $prefix)}";
+	@return $selector;
+}
```
olaf-k, line 20: this seems to be unused? to be removed if that's the case.

olaf-k (changes requested):

libs/components/src/lib/layout/README.md
YonatanKra: I agree with @olaf-k regarding the link. I believe I offered it on Slack for another place where we assume knowledge (I believe it was regarding `auto-size` or something).

If we "imitate" CSS standards this way, we should link from professional jargon to some external documentation. 

I certainly don't know these things out of my head and need to google them :)

YonatanKra (approved):

libs/components/src/lib/layout/layout.template.ts
```diff
@@ -5,12 +5,13 @@ import type { ElementDefinitionContext, FoundationElementDefinition } from '@mic
...
 
 const getClasses = ({
-	columnBasis, gutters, columnSpacing, autoSizing
+	columnBasis, gutters, columnSpacing, autoSizing, rowSpacing
 }: Layout) => classNames(
 	'control',
 	[`column-basis-${columnBasis}`, Boolean(columnBasis)],
 	[`gutters-${gutters}`, Boolean(gutters)],
 	[`column-spacing-${columnSpacing}`, Boolean(columnSpacing)],
+	[`row-spacing-${!rowSpacing ? columnSpacing : rowSpacing}`, !Boolean(rowSpacing) ? Boolean(columnSpacing) : Boolean(rowSpacing)],
```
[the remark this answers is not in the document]
rachelbt: `Boolean(rowSpacing) || Boolean(columnSpacing)],`
@olaf-k  will it make the same? I don't want it to have the value of this or that. I want it to have column if **theres no** row

olaf-k (approved):

PR #1144, opened 2023-05-04
fix(appearance): readonly state (VIV-1048)

rachelbt:


libs/shared/src/lib/sass/mixins/ReadMe.md
```diff
@@ -67,4 +67,26 @@ hover: (
 	outline: transparent,
 	),
 ...
-``
+```
+
+## Using Appearance
+in the component file we need to import the connotation with the shades that is being used in the appearance
```
YonatanKra, line 73:
```suggestion
In the component file, we need to import the connotation with the shades that are being used in the appearance
```

YonatanKra (changes requested):

YonatanKra (approved):

PR #1145, opened 2023-05-04
fix(tabs): make sure `activeid` is always set

olaf-k:
port of https://github.com/microsoft/fast/pull/6606 but on our v2.48 fast-foundation dep.

### DIFF
--- a/libs/components/src/lib/tabs/tabs.ts
+++ b/libs/components/src/lib/tabs/tabs.ts
@@ -20,21 +20,25 @@ export class Tabs extends FoundationTabs {
 		if (this.orientation === TabsOrientation.vertical) {
 			this.activeIndicatorRef.style.removeProperty(ACTIVE_TAB_WIDTH);
 		}
+		this.patchActiveID();
 	}
 
 	override activeidChanged(oldValue: string, newValue: string): void {
 		super.activeidChanged(oldValue, newValue);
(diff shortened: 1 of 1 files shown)


libs/components/src/lib/tabs/tabs.ts
```diff
@@ -43,4 +47,12 @@ export class Tabs extends FoundationTabs {
 		const width = this.activetab.getBoundingClientRect().width;
 		this.activeIndicatorRef.style.setProperty(ACTIVE_TAB_WIDTH, `${width}px`);
 	}
+
+	// adapted FAST fix https://github.com/microsoft/fast/pull/6606
```
YonatanKra, line 51: What does it solve? 
If there was a bug it should be documented in a test

YonatanKra (changes requested):

libs/components/src/lib/tabs/tabs.ts
olaf-k: the particular bug that requires this fix backport is in the PR title. it is a visual glitch and as such not reproduceable in our tests. hopefully it won't be needed [once the fix lands in Foundation 2.x](https://github.com/microsoft/fast/pull/6606#issuecomment-1541346522).

YonatanKra (approved):
LGTM :)

PR #1153, opened 2023-05-09
chore(cd): split the github packages and npm release (VIV-1042)

YonatanKra:
You can see an example action here: https://github.com/Vonage/vivid-3/actions/runs/4926533963


olaf-k (changes requested):
I'm confused. I believe the goal of this ticket was to split the npm and GH publications to their own separate actions (called from the `release-please` one), so that they could be 1) manually and 2) individually triggered when one of them fail.

Also, `--dry-run` would obviously need to be removed from the actual commands.

YonatanKra:
> I'm confused. I believe the goal of this ticket was to split the npm and GH publications to their own separate actions (called from the `release-please` one), so that they could be 1) manually and 2) individually triggered when one of them fail.
> 
> Also, `--dry-run` would obviously need to be removed from the actual commands.

Hi,
I can't see any need to split them into different files as they are not reused anywhere.
The idea is to separate them to different jobs so if one fails, you can just restart the failed one (in case the failure is what we experienced with github packages downtime).

Nice catch on the dry run ;)

olaf-k:
> Hi, I can't see any need to split them into different files as they are not reused anywhere.

the idea was to be able to give them a manual trigger (like the Deno deploy one). but GH's `re-run failed jobs` might indeed be enough.

olaf-k (approved):

PR #1156, opened 2023-05-11
feat(docs): add components filter (VIV-000)

YonatanKra:


apps/docs/_includes/navigation.njk
```diff
@@ -19,6 +19,27 @@
...
+				const term = e.target.value;
+				const components = Array.from(e.target.parentNode.children);
+				components.forEach(component => {
+					isComponentHidden(component, term) ?
+						component.classList.add('hidden') :
+						component.classList.remove('hidden')
+				});
+			};
+  	</script>
+  	<vwc-text-field placeholder="Search Components" oninput="filterComponentsInput(event)"></vwc-text-field>
```
rachelbt, line 42:
```suggestion
  	<vwc-text-field placeholder="Search Components" oninput="filterComponentsInput(event)" icon="search-line"></vwc-text-field>
```
adding a search icon

YonatanKra: It can be at the top of the list, but the filter only filters the components.  Putting it at the top of the list seems unrelated to me.
Waiting for @AyalaBu final input here.

AyalaBu:
@rachelbt 
I think on the top of the list it looks better

olaf-k (changes requested):
traditionally, JS and CSS used by the doc are stored in `apps/docs/assets` (under `scripts/` and `styles/` respectively), not directly in the template files.

apps/docs/_includes/navigation.njk
```diff
@@ -19,6 +19,27 @@
...
+			}
+  	</style>
+  	<script>
+  		function isComponentHidden(component, term) {
+    		return component.tagName === 'VWC-NAV-ITEM' && !component.text.toLowerCase().includes(term.toLowerCase());
+    	}
+
+  		function filterComponentsInput(e) {
+				const term = e.target.value;
+				const components = Array.from(e.target.parentNode.children);
```
olaf-k, line 34: this doesn't need to be done on every keystroke. get the components list on window.onload for instance and keep it in a var.

apps/docs/assets/scripts/components.filter.js
```diff
@@ -0,0 +1,24 @@
...
+
+function filterComponentsInput(components) {
+	return (e) => {
+		const term = e.target.value;
+		components.forEach(component => {
+			isComponentHidden(component, term) ?
+				component.classList.add('hidden') :
+				component.classList.remove('hidden')
+		});
+		document.querySelectorAll('vwc-nav-disclosure:not([label="Components"])').forEach(disclosure => {
```
olaf-k, line 13: what about also caching this in a var instead of running the query at every keystroke?

olaf-k (approved):

PR #1157, opened 2023-05-11
feat(menu-item): text wrap (VIV-1058)

rachelbt:


libs/components/src/lib/menu-item/menu-item.scss
```diff
@@ -107,20 +110,25 @@
...
+
 .text-primary {
+	-webkit-line-clamp: var(#{menu-item-variables.$text-primary-line-clamp}, 1);
 	.base.two-lines & {
 		font: var(#{constants.$vvd-typography-base-bold});
 	}
 }
 
 .text-secondary {
+	-webkit-line-clamp: var(#{menu-item-variables.$text-secondary-line-clamp}, 1);
```
YonatanKra, line 131: Funny `lint` doesn't shout on vendor prefixes here :)

YonatanKra (approved):
LGTM

libs/components/src/lib/menu-item/menu-item.scss
rachelbt: there's currently no support without it - and it applies to all browsers...
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

PR #1158, opened 2023-05-12
fix(fonts): force usage of proprietary fonts (VIV-1090)

gullerya:
Right now the value of `font-display` is `optional`, which means that browser gives up very quickly on the usage of the correct font and stays with the fallback. This causes in most cases the effect that first render of the page stays on a fallback fonts and only refresh brings the brand fonts, which is unacceptable in the branded application cases. `block` suggested here (or you may consider `swap` as well) will make the swap period infinite and eventually will render the branded fonts, which is the desired behavior in the proprietary Vonage applications page.


olaf-k (approved):
looks good. @rachelbt any counter indication?

YonatanKra:
@rachelbt - any objections?

rachelbt:
> @rachelbt - any objections?

no. we can merge

gullerya:
Cool, when do you think we can get a release version that will include this?

PR #1161, opened 2023-05-15
feat(menu-item): add slot (VIV-1078)

rachelbt:


libs/components/src/lib/menu-item/menu-item.spec.ts
```diff
@@ -278,4 +278,33 @@ describe('vwc-menu-item', () => {
...
+	describe('slot', ()=> (
+		it('should render slot if role menuItem', async function () {
+			element.role = MenuItemRole.menuitem;
+			const metaSlotElement = element.shadowRoot?.querySelector('.base slot[name="meta"]');
+			await elementUpdated(element);
+
+			expect(metaSlotElement).toBeTruthy();
+		}),
+
+		it('should not render slot if role is different then menuItem', async function () {
```
YonatanKra, line 291:
```suggestion
		it('should remove slot if role is different then menuItem', async function () {
```

libs/components/src/lib/menu-item/menu-item.spec.ts
```diff
@@ -278,4 +278,33 @@ describe('vwc-menu-item', () => {
...
+			const metaSlotElement = element.shadowRoot?.querySelector('.base slot[name="meta"]');
+			await elementUpdated(element);
+
+			!expect(metaSlotElement);
+		}),
+
+		it('should add class .has-meta if slot is slotted', async function () {
+			const slottedElement = document.createElement('div');
+			slottedElement.slot = 'meta';
+			slottedElement.id = 'meta';
```
YonatanKra, line 302: Would the test pass if this step is removed?

rachelbt: I guess so - I did the same as was in Card :)

libs/components/src/lib/menu-item/menu-item.template.ts
```diff
@@ -22,7 +22,8 @@ const getClasses = ({
 	['expanded', Boolean(expanded)],
 	['item-checkbox', role === MenuItemRole.menuitemcheckbox],
 	['item-radio', role === MenuItemRole.menuitemradio],
-	['two-lines', Boolean(text?.length) && Boolean(textSecondary?.length)]
+	['two-lines', Boolean(text?.length) && Boolean(textSecondary?.length)],
+	['has-meta', !!metaSlottedContent && Boolean(metaSlottedContent.length)]
```
olaf-k, line 26:
```suggestion
	['has-meta', Boolean(metaSlottedContent) && Boolean(metaSlottedContent.length)]
```
for consistency

olaf-k:
not related to your changes at all, but would you mind taking the opportunity to remove the two useless comments in `menu-item.template.ts`? ([this one](https://github.com/Vonage/vivid-3/blob/VIV-1078-menu-item-slot/libs/components/src/lib/menu-item/menu-item.template.ts#L45) and [this one](https://github.com/Vonage/vivid-3/blob/VIV-1078-menu-item-slot/libs/components/src/lib/menu-item/menu-item.template.ts#L49).)

libs/components/src/lib/badge/badge.scss
```diff
@@ -32,10 +32,13 @@
 	color: var(#{appearance.get-appearance-token(text)});
 	column-gap: 8px;
 	font: var(#{constants.$vvd-typography-base-condensed-bold});
-	max-inline-size: 100%;
 	padding-inline: 8px;
 	vertical-align: middle;
 
+	&:not(.icon-only) {
+		max-inline-size: 100%;
+	}
```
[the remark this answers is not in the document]
rachelbt: no. It caused the badge to have no width when inside flex parent. 
It was supposed to be like this before, button is like this as well.
Didn't think that one line effort a new PR.
Same goes with fixing the menu css variables

olaf-k (approved):

PR #1164, opened 2023-05-17
fix(alert): complete component (VIV-1038)

olaf-k:


libs/components/src/lib/alert/alert.ts
```diff
@@ -74,85 +71,65 @@ export class Alert extends FoundationElement {
...
 	}
 
-	override attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
-		super.attributeChangedCallback(name, oldValue, newValue);
-		switch (name) {
-			case 'open': {
-				this.open ? this.#show() : this.remove();
-			}
-		}
+	get conditionedIcon() {
```
YonatanKra, line 116: This should be tested too (exposed API).
It has 4 cases:
1. Default
2. When connotation is set
3. When icon is set
4. When connotation and icon are set together.

YonatanKra (changes requested):
Good job. Mostly docs/test related stuff.

libs/components/src/lib/alert/README.md
```diff
@@ -10,193 +10,196 @@ Alerts are meant to be used on top of pages, outside the main content.
...
   }
 </script>
 ```
 
-### Removable
+### Icon
 
-- Type: `boolean`
-- Default: `false`
+Use the `icon` attribute to add an icon to your alert.
```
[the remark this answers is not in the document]
olaf-k: clarified in README and tests

libs/components/src/lib/alert/alert.ts
olaf-k: I rearranged the icon tests and added checks for that one

libs/components/src/lib/alert/alert.scss
```diff
@@ -60,7 +60,7 @@ $alert-transition-delay: --transition-delay;
 		transform: scale(1);
 	}
 
-	min-inline-size: var(#{variables.$alert-min-inline-size}, 420px);
+	min-inline-size: var(#{variables.$alert-inline-size}, 420px);
```
rachelbt, line 63:
```suggestion
	min-inline-size: var(#{variables.$alert-min-inline-size}, 420px);
	max-inline-size: var(#{variables.$alert-max-inline-size}, fit-content);
```
if no max-width is set if there's lots of text the alert can be beryyyy wide :)

I'm being lazy adding this to all the places that are needed - variable + readme

libs/components/src/lib/alert/alert.scss
```diff
@@ -77,17 +77,14 @@ $alert-transition-delay: --transition-delay;
...
-		justify-self: flex-start;
 		margin-inline-end: auto;
 
 		.headline {
 			font: var(#{constants.$vvd-typography-base-bold});
+			margin-block-end: 4px;
 		}
 
-		.subtitle {
+		.maintext {
```
rachelbt, line 87: when we are using 2 words we usually have "-". 
maybe `alert-message` is better?

libs/components/src/lib/alert/README.md
```diff
@@ -10,193 +10,197 @@ Alerts are meant to be used on top of pages, outside the main content.
...
-    alert = document.querySelector('vwc-alert');
-    alert.placement = placement;
+  alert = document.querySelector('vwc-alert');
+  function openAlert() {
+    alert.open = true;
   }
 </script>
 ```
 
 ## Slots
```
rachelbt, line 158: you are correct that this is the default - bus since we also have template we usually give a name to the slot.
In toggle tip theres no option adding members of text on the host - but only in the default slot. Appose to Card that has text members on the host (with pre-defined template) and other slots like body-footer.
if so, maybe the maintext is a subtitle :)
OR - the "default" behaves like main slot and removes the headling and text.
WDYS?

Also - we need to set the gaps for this scenario:
image

rachelbt (commented):
looks so much better!
Didn't check the test :) - leaving this to @YonatanKra

YonatanKra (approved):
Only thing on my end is to separate the icon and connotation icon describe, but aside from that I'm good. @rachelbt has some comments of her own.

libs/components/src/lib/alert/README.md
```diff
@@ -10,193 +10,197 @@ Alerts are meant to be used on top of pages, outside the main content.
...
-
-Use the `subtitle` attribute to set the alert's subtitle.
+- Default: `undefined`
 
 ```html preview
-<vwc-alert subtitle="What an important info!!!" open></vwc-alert>
+<vwc-alert text="An important information for you" open></vwc-alert>
 ```
 
 ### Headline
```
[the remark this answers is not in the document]
olaf-k: I disagree. `text` without `headline` makes sense, not the other way around.

libs/components/src/lib/alert/README.md
```diff
@@ -10,193 +10,197 @@ Alerts are meant to be used on top of pages, outside the main content.
...
 ```html preview
-  <vwc-alert subtitle="What an important info!!!" connotation="accent" open></vwc-alert>
+<vwc-alert text="An important information for you" connotation="accent" open></vwc-alert>
 
-  <vwc-button appearance='outlined' label="Show accent connotation" onclick="openAlert('accent')"></vwc-button>
-  <vwc-button appearance='outlined' label="Show success connotation" onclick="openAlert('success')"></vwc-button> 
-  <vwc-button appearance='outlined' label="Show warning connotation" onclick="openAlert('warning')"></vwc-button> 
-  <vwc-button appearance='outlined' label="Show alert connotation" onclick="openAlert('alert')"></vwc-button>
-  <vwc-button appearance='outlined' label="Show information connotation" onclick="openAlert('information')"></vwc-button>
+<vwc-button appearance='outlined' label="Show accent connotation" onclick="openAlert('accent')"></vwc-button>
```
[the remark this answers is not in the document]
olaf-k: probably not, I just did not want to change literally *everything* 😅
will give it a try

libs/components/src/lib/alert/alert.scss
olaf-k: good point. I'll add that.

libs/components/src/lib/alert/README.md
olaf-k: I added the default slot on top of `headline` & `text` without giving it too much thought. I took for granted that it was an "either/or" situation, but I did not make it explicit in the code indeed. will change the template.

apps/docs/_data/components.json
```diff
@@ -33,11 +33,12 @@
 	{
 		"title": "alert",
 		"markdown": "./libs/components/src/lib/alert/README.md",
-		"status": "underlying",
```
rachelbt, line 36: I think we need to make this alpha :)

rachelbt (commented):
Looks great.
Left you one last comment and lets merge :)

YonatanKra (approved):
LGTM

PR #1169, opened 2023-05-22
feat(empty-state): add empty-state (VIV-1060)

RichardHelm:


libs/components/src/lib/empty-state/README.md
```diff
@@ -0,0 +1,89 @@
...
+### Headline
+
+Use the `headline` attribute add a headline to the empty state.
+
+- Type: `string`
+- Default: `undefined`
+
+```html preview
+<vwc-empty-state headline="No results found">
+  No results match your search criteria.
```
rachelbt, line 29:
```suggestion

```
when we are presenting members we usually add in the code samples the member itself without other additions if posibile

libs/components/src/lib/empty-state/README.md
```diff
@@ -0,0 +1,89 @@
...
+### Icon
+
+Use the `icon` attribute to set the icon of the empty state.
+
+- Type: `string`
+- Default: `inbox-line`
+
+```html preview
+<vwc-empty-state icon="search-line">
+  No results match your search criteria.
```
rachelbt, line 42:
```suggestion

```
same.
You can alternatively add headline since it was presented previously.

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,58 @@
+@use "../../../../../dist/libs/tokens/scss/tokens.constants" as constants;
+
+:host {
+	display: block;
```
rachelbt, line 4: why is this needed?

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.base {
+	display: flex;
+	height: 100%;
```
rachelbt, line 9:
```suggestion
	block-size: 100%;
```
I actually not sure if we need block-size.
What about inline-size? should it be auto (100%) or fit content?

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,58 @@
...
+	flex-direction: column;
+	align-items: center;
+	justify-content: center;
+	padding: 48px;
```
rachelbt, line 13: why is the padding needed?

libs/components/src/lib/empty-state/empty-state.template.ts
```diff
@@ -0,0 +1,52 @@
...
+				<slot name="graphic">
+					<div class="icon-container">
+						<${iconTag} class="icon" name="${(x) => x.icon}" size="5"></${iconTag}>
+					</div>
+				</slot>
+			</div>
+			<div class="content">
+				${when(
+		(x) => x.headline,
+		html<EmptyState>`<h3 class="headline">${(x) => x.headline}</h3>`
```
rachelbt, line 41:
```suggestion
		html`<h3 class="headline">${(x) => x.headline}</h3>`
```
also - I know that in vivid-2 is `<h3>` but is it correct? Should it be a headline at all, and if so - do we need to allow the consumers to change it if its not the right headline?

libs/components/src/lib/empty-state/empty-state.template.ts
```diff
@@ -0,0 +1,52 @@
...
+	)}
+				<div class="body">
+					<slot>No results.</slot>
+				</div>
```
rachelbt, line 45: not sure this is needed, anyway it is not documented...

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,58 @@
...
+.actions {
+	display: flex;
+	flex-wrap: wrap;
+	align-items: center;
+	justify-content: center;
+	gap: 8px;
+
+	.base.no-actions & {
+		display: none;
+	}
```
rachelbt, line 57:
```suggestion

```
I don't think that we need this.
if there's no slotted items then the slot is not rendered.
If there was padding or margin - then this was needed

libs/components/src/lib/empty-state/empty-state.template.ts
```diff
@@ -0,0 +1,52 @@
...
+import type {
+	ElementDefinitionContext,
+	FoundationElementDefinition,
+} from '@microsoft/fast-foundation';
+import { classNames } from '@microsoft/fast-web-utilities';
+import { Icon } from '../icon/icon';
+import type { EmptyState } from './empty-state';
+
+const getClasses = (x: EmptyState) => classNames('base',
+	['no-actions', x.slottedActionItems?.length === 0]
```
rachelbt, line 12: not sure this is need - see my comment in the scss file

rachelbt (changes requested):

libs/components/src/lib/empty-state/empty-state.template.ts
olaf-k: +1. maybe `<header>` as in some other components would just do?

olaf-k (changes requested):
looks good!

libs/components/src/lib/empty-state/empty-state.scss
RichardHelm: This is so that height and margin can be set on the empty state

libs/components/src/lib/empty-state/empty-state.scss
RichardHelm: I've added this so that the content will be vertically centered if the host element is larger
I think for inline-size the default (auto) is fine, so that it gets horizontally centered

libs/components/src/lib/empty-state/empty-state.scss
RichardHelm: It's part of the Figma design and I suppose this is so that it will be appropriately padded by default

libs/components/src/lib/empty-state/empty-state.template.ts
RichardHelm: Sure, I've changed to a `<header>`
If I remove `<EmptyState>` though it doesn't typecheck  anymore so I've kept that

libs/components/src/lib/empty-state/empty-state.template.ts
RichardHelm: I've added a default because the component looks weird without body text
Changed it to `Nothing to show` now, if that's fine?

libs/components/src/lib/empty-state/empty-state.scss
RichardHelm: I've added this because otherwise it creates a gap to the empty container for the actions:
image

libs/components/src/lib/empty-state/empty-state.ts
```diff
@@ -0,0 +1,28 @@
...
+	 * An optional headline for the empty state.
+	 * @public
+	 */
+	@attr headline?: string;
+
+	/**
+	 * Icon for the empty state.
+	 * @public
+	 */
+	@attr({mode: 'fromView'}) icon: string = 'inbox-line';
```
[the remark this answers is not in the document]
RichardHelm: I've added this because I got a linter error:
```
ESLint: 'attr' decorator assigned with a default value (unless mode is set to 'fromView', or, to 'boolean' assigned to false) will mutate the custom element in the DOM light tree.(@nrwl/nx/workspace/no-attribute-default-value)
```
Don't know if there's a better way to go about it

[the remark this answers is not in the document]
olaf-k: got it. in that case it can make sense to leave the attribute optional and set a fallback in the template (something along the lines of `${(x) => x.icon || 'inbox-line'}`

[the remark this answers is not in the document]
RichardHelm: Sure 👍 , updated it now

libs/components/src/lib/empty-state/empty-state.scss
rachelbt: but why do we need this to be full width and full height?

libs/components/src/lib/empty-state/empty-state.scss
rachelbt: oh, I see.

libs/components/src/lib/empty-state/empty-state.scss
RichardHelm: So, the full width is so that it centers horizontally on the page by default
And the full height doesn't do anything unless the `:host` has become larger, in which case we want it to vertically center in there as well.

Here is how I image it to be used:
image
image

olaf-k (dismissed):

libs/components/src/lib/empty-state/empty-state.scss
rachelbt: I'm not that sure about the inline-padding but I'll leave it

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,57 @@
...
+.content {
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+}
+
+.icon-container {
+	display: flex;
+	width: 120px;
+	height: 120px;
```
rachelbt, line 26:
```suggestion
	inline-size: 120px;
	block-size: 120px;
```

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,57 @@
...
+	align-items: center;
+	justify-content: center;
+	background-color: var(#{constants.$vvd-color-neutral-100});
+	border-radius: 60px;
```
rachelbt, line 30:
```suggestion
	border-radius: 50%;
```

libs/components/src/lib/empty-state/empty-state.template.ts
```diff
@@ -0,0 +1,52 @@
...
+	context: ElementDefinitionContext
+) => {
+	const iconTag = context.tagFor(Icon);
+
+	return html<EmptyState>`
+		<div class="${getClasses}">
+			<div class="graphic">
+				<slot name="graphic">
+					<div class="icon-container">
+						<${iconTag} class="icon" name="${(x) => x.icon ?? 'inbox-line'}" size="5"></${iconTag}>
```
rachelbt, line 34: I don't think we need a default icon.
What if there's no need for an icon or no place for it?
If this is removed - then I guess we can wrap the `.icon-container` with when there's an icon

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+header {
+	font: var(#{constants.$vvd-typography-base-bold});
+	text-align: center;
+}
+
+.body {
+	font: var(#{constants.$vvd-typography-base});
+	text-align: center;
```
rachelbt, line 40: this is redundant.
you can set `text-align: center` on `.content` and it will effect both of them

libs/components/src/lib/empty-state/ui.test.ts
```diff
@@ -0,0 +1,37 @@
...
+} from '../../visual-tests/visual-tests-utils.js';
+
+const components = ['empty-state', 'button'];
+
+test('should show the component', async ({ page }: { page: Page }) => {
+	const template = extractHTMLBlocksFromReadme(
+		path.join(new URL('.', import.meta.url).pathname, 'README.md')
+	).reduce(
+		(htmlString: string, block: string) =>
+			`${htmlString} <div style="margin: 5px;">${block}</div>`,
```
rachelbt, line 17: try adding height of 380px, to the div so our PNG will be smaller

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.base {
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+	justify-content: center;
+	padding: 48px;
+	block-size: 100%;
+	gap: 24px;
```
rachelbt, line 14:
```suggestion
	gap: 24px;
	box-sizing: border-box;
```

olaf-k (dismissed):

RichardHelm:
Hey @rachelbt, please let me know if you're happy with the PR now

rachelbt (approved):

PR #1172, opened 2023-05-25
feat(elevation): moving style to :before pseudo element (VIV-1098)

rachelbt:
Elevation adds background color and drop-down shadow filter and transition to its slotted item, causing component to change slotted items with position-fixed to be fixed to the new containing block (the component with elevation) and not the view-port.

In dialog showModal the filter is hidden by overflow-hidden therefor box-shadow was added.
The dialog box-shadow has hardcoded color and not variables because there's no need for color revert in darkmode.

### DIFF
--- a/libs/components/src/lib/dialog/dialog.scss
+++ b/libs/components/src/lib/dialog/dialog.scss
@@ -10,7 +10,7 @@ $dialog-space-size: 24px;
 	box-sizing: border-box;
 	padding: 0;
 	border: none;
-	background: var(elevation.$vvd-elevation-fill);
+	background-color: transparent;
 	border-radius: 6px;
 	color: var(#{constants.$vvd-color-canvas-text});
 	max-block-size: var(#{variables.$dialog-max-block-size}, calc(100vh - 64px));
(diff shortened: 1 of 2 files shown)


libs/components/src/lib/dialog/dialog.scss
```diff
@@ -31,6 +31,9 @@ $dialog-space-size: 24px;
 	}
 
 	&.modal {
+		// reason for box-shadow & hardcoded color - PR-XXX
```
rachelbt, line 34:
```suggestion
		// reason for box-shadow & hardcoded color - PR-1172
```

libs/components/src/lib/elevation/elevation.scss
```diff
@@ -18,13 +18,35 @@ $dps: 0 2 4 8 12 16 24;
...
 
 	::slotted(*) {
-		background: var(elevation.$vvd-elevation-fill);
-		filter: var(elevation.$vvd-elevation-shadow);
-		transition: background-color 0.15s linear, filter 0.15s linear;
+		position: relative;
+
+		@include utils.spread-list-selectors(".dp", $dps, $default) using($type) {
+			@include elevation.elevation($type);
+		}
```
olaf-k, line 25: not necessary

olaf-k (dismissed):

YonatanKra (approved):
Nice and elegant solution :)
This is an example of a case in which End of Action in Thought First ;)

PR #1173, opened 2023-05-25
feat(nav-item): add meta slot (VIV-1097)

RichardHelm:


libs/components/src/lib/text-anchor/text-anchor.ts
```diff
@@ -16,6 +16,14 @@ export class TextAnchor extends Anchor {
...
 	 */
 	@attr text?: string;
+
+	/**
+	 * Allows subclasses to provide a body template that will be rendered inside the anchor.
+	 * @internal
+	 */
+	getBodyTemplate(): ViewTemplate | undefined {
+		return undefined;
+	}
```
olaf-k, line 26: I'm not super comfortable with this way of patching TextAnchor.
either we code a slot directly in `text-anchor` which `nav-item` can then use

or

`nav-item` has its own template (which admittedly will be 99% like `text-anchor` but with a slot.
@YonatanKra thoughts?

olaf-k (changes requested):

apps/docs/_includes/navigation.njk
```diff
@@ -20,20 +20,22 @@
...
     {%  for item in components | publicPageFilter | sort(false, false, 'title') %}
     <vwc-nav-item
 			href="{{ ('/components/' + item.title) | url }}"
-			text="{{ item.title | replace("-", " ") | capitalize }}
-			{% if (item.status) == "alpha" %}🧪{% endif %}
-			{% if (item.status) == "underlying" %}🔒{% endif %}"
-			{% if ('/components/' + item.title + '/') == page.url %}aria-current="page"{% endif %}></vwc-nav-item>
+			text="{{ item.title | replace("-", " ") | capitalize }}"
+			{% if ('/components/' + item.title + '/') == page.url %}aria-current="page"{% endif %}>
+				{% if (item.status) == "alpha" %}<vwc-badge slot="meta" text="alpha" appearance="subtle" shape="pill" connotation="success"></vwc-badge>{% endif %}
```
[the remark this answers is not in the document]
RichardHelm: Sure, I've changed it 👍

libs/components/src/lib/nav-item/README.md
```diff
@@ -71,3 +71,43 @@ Within a set of pagination links, set a nav item `aria-current` value to *page*
...
+	</vwc-nav-item>
+</vwc-nav>
+
+<script>
+  function onClick(event) {
+    currentNavItem = document.querySelector('vwc-nav-item[aria-current="page"]');
+    currentNavItem?.removeAttribute('aria-current');
+    event.currentTarget.setAttribute('aria-current', 'page');
+  }
+</script>
```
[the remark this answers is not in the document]
RichardHelm: Thanks, updated the examples

libs/components/src/lib/text-anchor/text-anchor.ts
YonatanKra: I agree with @olaf-k here.
I'm not too much into slots, but if this is the policy then why not add the slot to the `text-anchor`? Because adding an API to `text-anchor` for `getBodyTemplate` is just a complicated way of adding a slot :)

RichardHelm: Sure I can could add a slot to text-anchor, but's it not used as a component here, instead the template is reused.
`breadcrumb-item` reuses the template as well, so that means it would also get `meta` slot which is unintended
What do you suggest to do then?

acelletti: @YonatanKra I agree with @RichardHelm that adding a slot to the text anchor would have unintended consequences and I think the implementation here makes sense. I am approving this and unless you have any actionable changes I would be inclined to merge this.

acelletti (approved):

PR #1175, opened 2023-05-25
fix(icon): fix icon color (VIV-1101)

RichardHelm:


libs/components/src/lib/nav-item/README.md
```diff
@@ -71,3 +45,25 @@ Within a set of pagination links, set a nav item `aria-current` value to *page*
...
+```html preview
+<vwc-nav>
+	<vwc-nav-item href="#" icon="profile" text="Account"></vwc-nav-item>
+	<vwc-nav-item href="#" icon="profile" text="Account" aria-current="page"></vwc-nav-item>
+</vwc-nav>
+<h4>Icon only</h4>
+<vwc-nav>
+	<vwc-nav-item href="#" icon="profile" aria-label="Account"></vwc-nav-item>
+	<vwc-nav-item href="#" icon="profile" aria-label="Account" aria-current="page"></vwc-nav-item>
+</vwc-nav>
```
olaf-k, line 68:
```suggestion
<vwc-nav>
	<vwc-nav-item href="#" icon="profile" text="Account"></vwc-nav-item>
	<vwc-nav-item href="#" icon="profile" aria-label="Account"></vwc-nav-item>
</vwc-nav>
```
I understand the rationale to merge the two examples as they refer to the same attribute but we should keep them as short and to the point as possible. (the usage of aria-current is explained right before, there's no need to duplicate it here.)

olaf-k (changes requested):

olaf-k (approved):

rachelbt (changes requested):
this is not a fix - the icon need to be in that color - @AyalaBu please

AyalaBu:
In the nav-item the icons is supposed to have the same color as text,
since it's also clickable.

RichardHelm:
@rachelbt I've followed the Figma design where the icon has the same color as the text

rachelbt (approved):

PR #1178, opened 2023-05-26
fix(tests): flaky select ui test (VIV-000)

olaf-k:
the `select` implements a light dismiss feature that collapses it on `focusout`.
our ui tests need (at the time of writing) two open `select`s at the same time in the same page, which can sometimes work or fail depending on how fast the screenshot is taken. this PR ~~short circuits the `focusout` handler in the ui tests.~~ miserably fails to fix that and simply removes `open` in the second example.


rachelbt:
the issue with the test is FF applying the open defiantly sometimes.
That was originally fixed with setting the selects flow in one line.
A PR was added that created another line - making the test flaky again.

YonatanKra (approved):
LGTM

PR #1182, opened 2023-05-30
feat(tab): add trailing icon option (VIV-1109)

rachelbt:


libs/components/src/lib/tab/tab.spec.ts
```diff
@@ -30,6 +34,29 @@ describe('vwc-tab', () => {
 		});
 	});
 
+	describe('tab icon', () => {
```
YonatanKra, line 37:
```suggestion
	describe('icon', () => {
```

libs/components/src/lib/tab/tab.spec.ts
```diff
@@ -30,6 +34,29 @@ describe('vwc-tab', () => {
...
+		it('setting `iconTrailing` set the order of element', async () => {
+			element.icon = 'home';
+			element.iconTrailing = true;
+			await elementUpdated(element);
+
+			const trailingIcon = element.shadowRoot?.querySelector(
+				`.icon-trailing ${ICON_SELECTOR}`,
+			);
+			expect(trailingIcon).toBeInstanceOf(HTMLElement);
+		});
```
YonatanKra, line 56: Extract it to its own `describe` as this is an API unto itself.
Also, the description is vague and I can't see how the test tests it.
What should it do?
It looks as if the code just adds the class. Am I correct?

libs/components/src/lib/tab/tab.spec.ts
```diff
@@ -30,6 +34,29 @@ describe('vwc-tab', () => {
 		});
 	});
 
+	describe('tab icon', () => {
+		it('adds an icon to the tab', async () => {
```
YonatanKra, line 38:
```suggestion
		it('should add an icon to the tab when icon is set', async () => {
```

YonatanKra (changes requested):

olaf-k (dismissed):

libs/components/src/lib/tab/tab.spec.ts
rachelbt: Also - in all out unittest the trailing icon is under the icon describe

YonatanKra (approved):
LGTM

PR #1187, opened 2023-06-01
fix(components): updated CSS for disabled & readonly (VIV-1114)

olaf-k:


rachelbt (changes requested):
this change should cause UI test break - why didn't that happened?

olaf-k:
> this change should cause UI test break - why didn't that happened?

when not explicitely set, the default threshold for the change detection is not zero but `0.2` ([see here](https://playwright.dev/docs/api/class-snapshotassertions#snapshot-assertions-to-match-snapshot-1-option-threshold)), meaning a 20% color difference per pixel is acceptable. this minor change falls into this margin.

rachelbt:
> > this change should cause UI test break - why didn't that happened?
> 
> when not explicitely set, the default threshold for the change detection is not zero but `0.2` ([see here](https://playwright.dev/docs/api/class-snapshotassertions#snapshot-assertions-to-match-snapshot-1-option-threshold)), meaning a 20% color difference per pixel is acceptable. this minor change falls into this margin.

hhheeemmmmmm,
But shouldn't we generate new images (or you did and there was no new images?)

rachelbt (commented):
it doesn't make sense - there's no read-only or disable in action group.
Buttons, text-field (and all other input), checkbox + radio should be updated...

olaf-k:
> it doesn't make sense - there's no read-only or disable in action group. Buttons, text-field (and all other input), checkbox + radio should be updated...

I was checking the screenshots before pushing them all. the `action-group` is a fix for the display. some components will have updated screenshots even when not using disabled or readonly as a few bytes were shaved here and there. I've only included those where the three browsers had different files. you can consider this as a new baseline.

rachelbt (approved):
I updated the checkbox images after merge with main.
The images that do not contain change in the readonly or disable states has better images for elevation drop shadow (this is the reason they appear in this PR)

PR #1190, opened 2023-06-02
fix(button): update pending state (VIV-1113)

olaf-k:
See [Figma](https://www.figma.com/file/JJNgZvt1qf3ydYmOwbE3Jg/branch/3r5pL4wJVITLkBeOBfBPgg/Vivid-UI-Kit---3.0-WIP?type=design&node-id=22690-153037&t=USbJ7gdCs6k8TQww-0).

*tl;dr* pending indicator (progress ring) replaces icon when there is one, replaces text otherwise.

requires ~~#1188~~ and ~~#1187~~


rinaok (approved):

libs/components/src/lib/button/README.md
```diff
@@ -156,14 +156,19 @@ Add the `disabled` attribute to disable the button.
 ### Pending
 
 Add the `pending` attribute to disable the button and display a processing indicator.
+The indicator will replace the icon if one is set, or the label text for text-only buttons.
+The indicator is not displayed when using the `super-condensed` size.
```
rinaok, line 160: why?

olaf-k: that's how it's been designed. I'm guessing a11y issues with too small a progress indicator.
image

PR #1193, opened 2023-06-05
chore(data-grid): manual examples (VIV-1070)

YonatanKra:


rachelbt (commented):
mind reverting the "change" of one line space in the css file?
How come there are no PNG? Or is it not needed since we have them in the selected of single + multi?

rachelbt (approved):
LGTM

PR #1205, opened 2023-06-11
feat(tabs): cta connotation (VIV-1127)

rachelbt:
small tab bug that also be tended in this PR - add scroll to tab-list


libs/components/src/lib/tab/tab.spec.ts
```diff
@@ -70,4 +72,27 @@ describe('vwc-tab', () => {
 			expect(element.shadowRoot?.querySelector('.disabled')).toBeTruthy();
 		});
 	});
+
+	describe('ariaSelected', function () {
```
rinaok, line 76:
```suggestion
	describe('connotation', function () {
```

rinaok (approved):

rinaok (approved):

PR #1206, opened 2023-06-12
feat(file-picker): migrate from vivid-2 (VIV-962)

rinaok:


rachelbt (changes requested):
nice!
It is wrong nesting a button inside an a element.
If you can - it is best that it will be only the button.

rinaok:
@rachelbt @AyalaBu What do you think should be solution when we add a lot of files to the file picker, but the form should remain certain height?

AyalaBu:
@rachelbt 
I'm not sure that I have ever seen such a use case
I think that normally the rest of the fields just move down
Also, usually the form mentions maximum files that can be uploaded

libs/components/src/lib/components.ts
```diff
@@ -1,54 +1,55 @@
-export * from './accordion/definition';
```
acelletti, line 1: Why change the order of some of the rows in this file? Also, it looks to me like some have been removed e.g. `data-grid`.

rinaok:
> nice! It is wrong nesting a button inside an a element. If you can - it is best that it will be only the button.

@rachelbt Fixed :)

rachelbt:
> > nice! It is wrong nesting a button inside an a element. If you can - it is best that it will be only the button.
> 
> @rachelbt Fixed :)

NICE!!!!

libs/components/src/lib/file-picker/file-picker.scss
```diff
@@ -0,0 +1,143 @@
...
+
+.preview-list {
+	&:empty {
+		display: none;
+	}
+
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+	margin-block-start: 4px;
```
rachelbt, line 142:
```suggestion
	margin-block-start: 4px;
	overflow-y: auto;
```
if there are to many items :)

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+			expect(element.files.length).toEqual(1);
+
+			(getBaseElement(element).querySelector('.remove-btn') as HTMLElement).click();
+			await elementUpdated(element);
+			expect(element.files.length).toEqual(0);
+		});
+	});
+
+	describe('choose file on enter key', function () {
+		it('should open file on enter', async function () {
```
YonatanKra, line 273: The test here should most likely be that the `click` was made on the hidden element. 
But looking at the implementation I now see that `filePicker` (DZ) is exposed. It should be made private (true private) otherwise, it's an integral part of the API...

So the action items here are:
1. Make `filePicker` private.
2. The test should test what's really happening - the click event on the hiddenInput on `filePicker` should fire. That means, in the least, to mock DZ for tests purposes (using `jest.mock`) and listen to the click event on a mocked `#filePicker.hiddenFileInput`.

libs/components/src/lib/file-picker/file-picker.ts
```diff
@@ -0,0 +1,207 @@
...
+		return this.filePicker.getFilesWithStatus(Dropzone.ERROR);
+	}
+
+	addFile(file: File): void {
+		this.filePicker.addFile(file as DropzoneFile);
+	}
+
+	removeFile(file: File): void {
+		this.filePicker.removeFile(file as DropzoneFile);
+	}
```
YonatanKra, line 190: See my comments in the tests. This feature is not needed at all IMO.

libs/components/src/lib/file-picker/file-picker.template.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @public
+ */
+export const FilePickerTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<FilePicker> = (context: ElementDefinitionContext) => {
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<FilePicker>`
+	<div class="${getClasses}" aria-label="${x => x.ariaLabel ? x.ariaLabel : x.label}">
```
YonatanKra, line 30: If `label` and `ariaLabel` are not defined, what would I get in `aria-label`? Is this usecase tested?

YonatanKra (changes requested):
That's a big PR and I have quite a lot of comments (mostly due to the size of the PR and its nature).

Feel free to get in touch over slack for any questions and we can go over it together.

libs/components/src/lib/file-picker/file-picker.ts
```diff
@@ -0,0 +1,139 @@
...
+
+	constructor() {
+		super();
+		Dropzone.autoDiscover = false;
+	}
+
+	override connectedCallback() {
+		super.connectedCallback();
+
+		this.filePicker = new Dropzone(this._dz,
```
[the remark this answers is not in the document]
rinaok: @YonatanKra It's not clear what you meant, `filePicker` is private. Are you referring to _dz? The _dz can be changed to private instead of being a reference.

libs/components/src/lib/file-picker/README.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+## Slots
+
+### Default
+
+Use the default slot to set the content of the file picker.
+
+```html preview
+<vwc-file-picker>Add custom content here</vwc-file-picker>
```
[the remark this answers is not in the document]
rinaok: @YonatanKra It is an example for the users where is the slot

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+				.toEqual(labelText);
+		});
+
+		it('should show label element only if label is set', async function () {
+			const labelElement = element.shadowRoot?.querySelector('label');
+			expect(labelElement)
+				.toBeNull();
+		});
+
+		it('should set arialabel on host', async function () {
```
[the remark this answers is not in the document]
rinaok: There is already a mention of it:

Screenshot 2023-06-27 at 12 15 06

libs/components/src/lib/file-picker/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+- Type: `File[]`
+- Default: `[]`
+
+### Value
+
+Read more on [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#value).
+
+- Type: `string`
+- Default: `""`
```
rinaok, line 186: Do we need this property? Cause the value is [always the file's name prefixed with C:\fakepath\](https://html.spec.whatwg.org/multipage/input.html#fakepath-srsly), which isn't the real path of the file. This is to prevent malicious software from guessing the user's file structure.

libs/components/src/lib/file-picker/README.md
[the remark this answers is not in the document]
YonatanKra: What's the use case here? What `custom content` do you have in mind?

libs/components/src/lib/file-picker/file-picker.template.ts
rinaok: @YonatanKra The same thing applies to all our components (for example, select). Therefore, we have an explanation in the documentation:

`If no label is set - it is highly recommended that `aria-label` will be added.`

libs/components/src/lib/file-picker/file-picker.ts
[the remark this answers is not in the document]
YonatanKra: TS private is not really private - it is exposed. This is why `filePicker` isn't `true private` (unlike `#filePicker` with `#` prefix). 

`filePicker` is also DZ itself (`this.filePicker = new DropZone()`)

Regarding `_dz` - it is just an element. It is exposed, but for template needs only. 
It could be made private by just querying for it in the `connectedCallback`:
```
this.#filePicker = new DropZone(this.shadowRoot.querySelector('.control'), dropZoneConfig);
```
Or something like that.

libs/components/src/lib/file-picker/README.md
YonatanKra: That's how the spec work :)

libs/components/src/lib/file-picker/file-picker.spec.ts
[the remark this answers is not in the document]
YonatanKra: What I mean is - there's no mention that the `label` is used as `aria-label` in case the last is missing :)

PR #1207, opened 2023-06-14
feat(menu): add `auto-dismiss` attribute (VIV-1144)

acelletti:


rinaok (changes requested):
Hi @acelletti, 
The click event is still being listened to when the menu component is removed. Listeners should be removed when disconnectedCallback is invoked.

libs/components/src/lib/menu/menu.ts
```diff
@@ -49,4 +61,20 @@ export class Menu extends FastMenu {
...
+			const popup = (this._popup as Popup);
+			if (popup.open && !this.contains(e.target as HTMLElement)) {
+				popup.open = false;
+			}
+		};
+		if (prevAutoDismiss !== newAutoDismiss) {
+			if (newAutoDismiss) {
+				document.addEventListener('click', dismissOnClickOutside);
+			} else {
+				document.removeEventListener('click', dismissOnClickOutside);
```
YonatanKra, line 76: I believe this will cause a memory leak.
The callback is created anew on each `auto-dismiss` change and thus the `removeEventListener` will not remove the event, because it will not be the same function.
Here's a commit that solves this:
https://github.com/Vonage/vivid-3/commit/160087fbf009868826caa3ebfd63864c1334eab0

YonatanKra (changes requested):

libs/components/src/lib/menu/menu.ts
```diff
@@ -39,6 +39,18 @@ export class Menu extends FastMenu {
...
+	 * indicates whether the menu will automatically close when
+	 * the user clicks outside the menu
+	 *
+	 * @public
+	 * HTML Attribute: auto-dismiss
+	 */
+	@attr({
+		mode: 'boolean',
+		attribute: 'auto-dismiss'
+	}) autoDismiss? = false;
```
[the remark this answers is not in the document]
acelletti: Removed.

libs/components/src/lib/menu/menu.ts
```diff
@@ -49,4 +61,20 @@ export class Menu extends FastMenu {
...
 	};
+
+	autoDismissChanged(prevAutoDismiss: boolean, newAutoDismiss: boolean) {
+		const dismissOnClickOutside = (e: MouseEvent) => {
+			const popup = (this._popup as Popup);
+			if (popup.open && !this.contains(e.target as HTMLElement)) {
+				popup.open = false;
+			}
+		};
+		if (prevAutoDismiss !== newAutoDismiss) {
```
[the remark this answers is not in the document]
acelletti: Agreed, updated the code to use `this.autoDismiss` directly.

libs/components/src/lib/menu/menu.ts
acelletti: Nice catch, I have changed it to use a private method like you suggested.

acelletti:
> Hi @acelletti, The click event is still being listened to when the menu component is removed. Listeners should be removed when disconnectedCallback is invoked.

Added remove in `disconnectedCallback` to avoid this happening.

libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -61,6 +61,22 @@ describe('vwc-menu', () => {
 		});
 	});
 
+	describe('auto-dismiss', () => {
+		it('should set "open" to false when clicked outside', async () => {
```
YonatanKra, line 65: Please add a test that covers the memory leak prevention (both for the disconnected CB as well as inside the function).

YonatanKra (changes requested):

YonatanKra (approved):

rinaok (approved):

PR #1213, opened 2023-06-19
fix(card): inner wrapper height inherit(VIV-1151)

rachelbt:


libs/components/src/lib/card/README.md
```diff
@@ -169,3 +169,27 @@ The number of lines is controlled by css variable `--subtitle-line-clamp`.
 >
 </vwc-card>
 ```
+
+## Use case
+```html preview
+<style>
+  vwc-card {}
+</style>
```
rinaok, line 177: can be removed

rachelbt: right! thanks

rachelbt, line 177:
```suggestion
```

rachelbt:
> vwc-layout should be added to the ui.test.ts

it is there - the width of the ui-test is 400 something thats why they appear one after another...

rinaok:
> > vwc-layout should be added to the ui.test.ts
> 
> it is there - the width of the ui-test is 400 something thats why they appear one after another...

@rachelbt can you make the width bigger?

rachelbt:
> > > vwc-layout should be added to the ui.test.ts
> > 
> > 
> > it is there - the width of the ui-test is 400 something thats why they appear one after another...
> 
> @rachelbt can you make the width bigger?

changes the style of the card in layout and updated the png's

PR #1215, opened 2023-06-19
docs(icon): add docs for icon custom size (VIV-1145)

acelletti:


libs/components/src/lib/icon/README.md
```diff
@@ -49,6 +49,20 @@ Note that changing `--vvd-size-density` will affect the chosen size (the example
 <vwc-icon name="close-line" size="5"></vwc-icon>60px
 ```
 
+### Custom Size
```
rachelbt, line 52: I'm not sure about the special section for that.
It was mentioned in the size section:

> Default: undefined (the icon dimensions will default to 1em×1em)
maybe its enough to add an example in the same iframe:
image

also - new png's need to be generated :)

acelletti: @rachelbt remember that not everybody has the same knowledge of CSS as you do 😄 
I think spelling it out in its own paragraph makes sense so that people can immediately see how to do it.

acelletti: I did originally have it as a sub-section, but we don't generate anchor links for those, and I want to be able to share the anchor link in case we need to point somebody to it in the future.

libs/components/src/lib/icon/README.md
```diff
@@ -49,6 +49,20 @@ Note that changing `--vvd-size-density` will affect the chosen size (the example
...
+
+```html preview
+<style>
+	.small-icon {
+		font-size: 8px;
+	}
+</style>
+
+<vwc-icon name='close-line' class="small-icon"></vwc-icon>8px
+```
```
rinaok, line 64: @rachelbt @acelletti 

It looks strange when it is small in this example:

Screenshot 2023-06-20 at 17 22 37

```suggestion
```html preview
<style>
	.big-icon {
		font-size: 100px;
	}
</style>

<vwc-icon name='close-line' class="big-icon"></vwc-icon>
```
```

A very big icon will look better in my opinion :

![Screenshot 2023-06-20 at 17 28 39](https://github.com/Vonage/vivid-3/assets/10883919/af1422da-159e-4de2-8c78-d49f4558d20d)

rachelbt: I agree with @rinaok , also 8px is really small and I'm not so sure it good UX.
Having said that - I'm not sure 100px is the right choice - thats way too big :)
lets try 12px, and can you pease add - **custom size: font size 12px**

acelletti: Updated the example to show `12px` instead of 8.

rinaok (approved):
Updated to 30px because the 12px didn't look good in the snapshots

PR #1225, opened 2023-06-25
fix(menu): block-size (VIV-1165)

rachelbt:


libs/components/src/lib/menu/README.md
```diff
@@ -146,6 +146,31 @@ Use the `--menu-min-inline-size` variable to set the menu's inline size.
 </vwc-menu>
 ```
 
+### Menu Block Size
+
+Use the `menu-block-size-size` variable to set the menu's block size.
```
rinaok, line 151:
```suggestion
Use the `menu-block-size` variable to set the menu's block size.
```

libs/components/src/lib/menu/README.md
```diff
@@ -146,6 +146,31 @@ Use the `--menu-min-inline-size` variable to set the menu's inline size.
...
+
+- Type: [`<length>`](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
+- Default: `408px`
```
rinaok, line 154:
```suggestion
- Default: `150px`
```

Why 408? It's 150

rachelbt: its 408 same as select:
`max-block-size: var(#{variables.$menu-block-size}, 408px);`

rinaok (approved):

PR #1236, opened 2023-06-30
feat(dialog): add action-items slot (VIV-1141)

RichardHelm:


YonatanKra (changes requested):
The issue (1441) references a discussion on Slack.
I need help understanding the use case and how two slots solve it.

RichardHelm:
@YonatanKra As per the Slack discussion, the use case is that a lot of dialogs (almost all in Dashboard) have primary and secondary action buttons. Adding these slots will make sure that they are laid out in accordance with the design system without and without the need to add CSS.

acelletti:
> The issue (1441) references a discussion on Slack. I need help understanding the use case and how two slots solve it.

This has been discussed and agreed on with me @YonatanKra.

YonatanKra:
> > The issue (1441) references a discussion on Slack. I need help understanding the use case and how two slots solve it.
> 
> This has been discussed and agreed on with me @YonatanKra.

Reverting from vivid-3 API back to vivid-2 API might be a foot gun. We diverted from the way it was done in vivid-2 for a reason. Why not agree as a group after hearing possible alternatives?

rachelbt (changes requested):
I find the extra 2 slots utterly unneeded and it is making component API messy and indecisive. 

I understand the issue with the easy migration from 2 to 3, and while we need to help our users with that, we also need to provide them with better, clean, and good code and API.
I want to suggest a different approach for this PR:

1. There will be no primary-action or secondary-action, these buttons will need to be changed to slot="footer".
2. We will add a property to combine with that, named "action-footer" (or something that will make sense).  When this property is added - we will add CSS for the footer for aligning the slotted items in the footer.

@YonatanKra @RichardHelm 
What do you think?

YonatanKra:
> I find the extra 2 slots utterly unneeded and it is making component API messy and indecisive.
> 
> I understand the issue with the easy migration from 2 to 3, and while we need to help our users with that, we also need to provide them with better, clean, and good code and API. I want to suggest a different approach for this PR:
> 
> 1. There will be no primary-action or secondary-action, these buttons will need to be changed to slot="footer".
> 2. We will add a property to combine with that, named "action-footer" (or something that will make sense).  When this property is added - we will add CSS for the footer for aligning the slotted items in the footer.
> 
> @YonatanKra @RichardHelm What do you think?

This request is a feature request and should be considered as such. That is why I asked for the use case.

Slots are as good a solution as exposing all of the columns in a multi-tables/multi-DB query result instead of constructing a fine-grained JSON as a response. In this case, it happens just because the use case is not defined and we are "violating" several code smell principles (like YAGNI for instance). 

That's not good practice engineering-wise, IMO.

It's impossible to tell what's the use case from the issue and very hard to extract it from the Slack discussion. 

If the use case is a confirm dialog (as implied from the request for two buttons - "OK/Cancel" or similar), then we should talk about it as a confirm dialog and generalize it for all of Vonage. There is [a spec for confirm](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-confirm-dev) as it is a native `window` method.

In this case, the question is - is it a variant of the dialog or a new component?

rinaok:
@RichardHelm In addition to the things we discussed on the Hudlle, the UI snapshots should be updated, they are difficult to understand at the moment.

RichardHelm:
@YonatanKra In Dashboard, almost all of the modal dialogs have action button. So I would consider them a fundamental part of the dialog component, and their layout should not be left up to the user.

@rachelbt I have to say that I don't really really agree with that. I find the "dual-use" of the footer slot more indecisive than a a dedicated slot. In addition, you may want to have both action buttons as well as footer content.

An argument could be made that we should just have a single `action-items` slot instead of two, which is more in line with Vivid and makes no difference functionality wise.

But the reason to separate primary and secondary actions is because it is a contentious issue where they are placed. E.g. on Windows the primary action comes first while on other platforms in comes last. So it's up to the design system to specify the order. For that reason, I think it's slightly better to have both.

rinaok:
> @YonatanKra In Dashboard, almost all of the modal dialogs have action button. So I would consider them a fundamental part of the dialog component, and their layout should not be left up to the user.
> 
> @rachelbt I have to say that I don't really really agree with that. I find the "dual-use" of the footer slot more indecisive than a a dedicated slot. In addition, you may want to have both action buttons as well as footer content.
> 
> An argument could be made that we should just have a single `action-items` slot instead of two, which is more in line with Vivid and makes no difference functionality wise.
> 
> But the reason to separate primary and secondary actions is because it is a contentious issue where they are placed. E.g. on Windows the primary action comes first while on other platforms in comes last. So it's up to the design system to specify the order. For that reason, I think it's slightly better to have both.

@RichardHelm Vonage apps, however, lack consistency when it comes to action buttons in dialog boxes. When I helped migrate VBC, for example, they had dialogs with actions in different orders. In other apps, the dialogs are different than those in the Dashboard.

For example two different dialogs in VBC:

Screenshot 2023-07-04 at 15 08 58

Screenshot 2023-07-04 at 15 08 48

Also, we should be consistent and use the same slots in different components. Here is an example for using the same two action buttons in the toggletip component:

https://vivid.deno.dev/components/toggletip#action-items

Screenshot 2023-07-04 at 15 10 17

acelletti:
@rinaok The whole point of a design system is to achieve consistency across solutions.
From our Design System Figma:
image

The examples you mentioned can still be done (in the case of legacy migrations) using the `footer` slot but this PR adds the ability to add the buttons using slots so that they look like the picture.

I also agree with the view that although `action-items` is used in other places so it does make sense to keep it consistent, I do see the point about enforcing the order of primary / secondary positioning as it's also part of the design system.

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -343,26 +343,32 @@ describe('vwc-dialog', () => {
...
+			async (slotName) => {
+				const slottedElement = document.createElement('div');
+				slottedElement.slot = slotName;
+				element.appendChild(slottedElement);
+				await elementUpdated(element);
 
-			const baseElementClasses = element.shadowRoot?.
-				querySelector('.base')?.classList;
+				const baseElementClasses =
+					element.shadowRoot?.querySelector('.base')?.classList;
```
rinaok, line 367:
```suggestion
				const baseElementClasses =
					getBaseElement(element)?.classList;
```

RichardHelm:
@rinaok Updated the screenshots now.

Ok, it might make more sense to use `action-items` then for consistency throughout the library. We could also add an option to move them to the left, but as Andrea was saying we probably don't want to encourage that.

What do you say @rinaok @acelletti ?

rinaok:
> action-items

@RichardHelm If you ask me, I think it should be consistent. We have the `action-items` slot in:
Alert, Banner, Empty state, Header, Toggletip components

libs/components/src/lib/dialog/dialog.spec.ts
rinaok: `getBaseElement` is a global function. Should be imported from `test-utils`

RichardHelm: Sure, done 👍

RichardHelm:
@rinaok Ok, I've changed it to `action-items`

rinaok (approved):

gfrileux:
Interesting discussion team, and another good opportunity for learning here. Few thoughts:

- I think the JIRA ticket could be improved. Instead of linking to a Slack chat that people may well not have access to, it should summarise what the use case is, and how we plan to solve it. A JIRA ticket should be pretty stand alone, as in everything needed to understand the task should be in the description. Sometimes it's perfectly fine to have links to Google Doc or Confluence (even better) for extra context, but links to Slack I find a bit unhelpful (unless it's a reference at the end along the lines of "for reference, this was discussed here <link>". Let's do better next time 😄 

- Rina is right that our UIs are currently inconsistent, and Richard is right that we'll have to address that by providing an opinionated component so that as people migrate, we achieve much better consistency. But there will always be the odd exception, and that's why the slot is here. There is no ambiguity here IMO Rachel, just a recommended approach (the two buttons) and a solution for exceptions. What I would say is that it's the Documentation's job to explain this clearly, and as for everything else, the Design team will have to be the gate keeper and ensure that developers don't deviate from their designs. Yonatan: this PR is a variant of the dialog. 

- Let's use `action-items` also, it's more consistent indeed.

- with regards to @YonatanKra 's point about feature changes and the comment `Why not agree as a group after hearing possible alternatives` , that's a good point also. Here are my thoughts: for new components, we've had another chat and we're in agreement that we need a "strong process" with a team review of the design. This should also apply to large refactors of existing components, or big feature changes. For smaller changes (like this one) and bug fixes however, it's fine for smaller groups of developers to work independently and come up with their own approach without involving the entire team. It's also fine to me if not everybody always agree on everything. It's tempting but very dangerous to stipulate that everyone should agree on every PR. When the team doubles in size and we have 8 developers, I can assure you we'll never get agreement of 8 people on every single PRs. At which point we'll be completely paralysed.
So if most of the time everyone agrees - great! If sometimes we can't get to all agree - that's good too, and can trigger great conversations and improve Vivid (like it did here). If after chatting we still disagree, that happens, and it's fine. At that point it's the manager's role decide. What is also important is that when disagreement happens, then we should be open minded and be ready to take feedback and re-consider certain choices based on the outcome once code is in prod and used by our users. If we realise later that a choice yielded negative outcome, then we fix it. I don't think anyone can make the right choices 100% of the time, so iteration will always have to happen. That's ok.

So it sounds like we've come to a good place, whereby Richard's taken onboard some comments, Rina's approved the PR, Andrea looks happy too, to me it looks good also so I'll approve it too.

And thank you for all the care and attention on this, it's lovely to see, I really do mean that 😃  Let's continue the discussion tomorrow and see where we can improve for next time.

gfrileux (approved):

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -119,13 +119,25 @@ $dialog-space-size: 24px;
...
+	gap: 24px;
 	padding-inline: $dialog-space-size;
 
 	.hide-footer & {
 		display: none;
 	}
-}
 
+	.footer-content {
+		align-self: center;
```
rachelbt, line 132: not sure this is needed- this is center vertical alignment.

RichardHelm: @rachelbt Yes, I've added this on purpose to be in line with the design in Figma:
image

rachelbt: oh, ok.
I think that it will be better to set the align-items: center on the footer, its better that way - to align to the center both the action-items and the footer content rather then having them have the stretch default.

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -119,13 +119,25 @@ $dialog-space-size: 24px;
 }
 
 .footer {
+	display: flex;
+	justify-content: space-between;
```
rachelbt, line 123:
```suggestion
	justify-content: space-between;
	align-items: center
```

libs/components/src/lib/dialog/dialog.scss
RichardHelm: Sure, done 👍

libs/components/src/lib/dialog/README.md
```diff
@@ -227,29 +223,41 @@ To remove the body inline padding use `full-width-body`.
...
   headline="Dialog with footer"
-  subtitle="this is an example of the dialog with slotted buttons inside footer">
-  <div slot="footer" class="demo-footer">
-    <vwc-button appearance="outlined" label="Cancel"></vwc-button>
-    <vwc-button appearance="filled" label="Action"></vwc-button>
-  </div>
+  subtitle="this is an example of the dialog with a checkbox inside footer">
+	<vwc-checkbox slot="footer" label="I agree"></vwc-checkbox>
+	<vwc-button slot="action-items" appearance="filled" label="Ok"></vwc-button>
 </vwc-dialog>
```
rinaok, line 261: In the footer slot section should be an example for slot="footer" and not slot="action-items"

rinaok (changes requested):

rinaok (approved):

PR #1239, opened 2023-07-04
fix(data-grid): cell ellipsis and sticky header background-color variable (VIV-1118)

rachelbt:
I also fixed the ellipsis on cell item that didn't work


libs/components/src/lib/data-grid/data-grid-cell.scss
```diff
@@ -17,23 +17,24 @@
 
 @use "../focus/partials/variables" as focus-variables;
 
+:host {
+	// this is for making overflow inside flex to work
```
rachelbt, line 21:
```suggestion
	// this is for making overflow inside grid to work
```

YonatanKra (approved):
Nice. Thanks.

PR #1240, opened 2023-07-04
fix(tabs): prevent tab-panel blowout

rachelbt:
since tab-panel is a grid child this issue is pretty common:
[Preventing a Grid Blowout](https://css-tricks.com/preventing-a-grid-blowout/)


YonatanKra (approved):
Nice. Thanks :)

PR #1241, opened 2023-07-04
chore(docs): new colors (VIV-1189)

rachelbt:


YonatanKra (approved):
Nice! Thanks. Maybe `getting started` and `design system` should start `closed`?

rachelbt:
> Nice! Thanks. Maybe `getting started` and `design system` should start `closed`?

I don't know - since I'm afraid people will miss the migration, font ect.
I'll keep it as it is now.
Thanks

PR #1245, opened 2023-07-05
fix(tsdoc): lint warnings (VIV-1182)

rinaok:


acelletti (commented):
Looks good, only question that I have is does it make sense to just disable the rule `@typescript-eslint/no-non-null-assertion` as I can see we ignore it in quite a few cases.
🤔

rinaok:
> Looks good, only question that I have is: does it make sense to just disable the rule `@typescript-eslint/no-non-null-assertion` as I can see we ignore it in quite a few cases? 🤔

@acelletti Thought about it, but wasn't sure because it could be an error, and the linter wouldn't warn.

acelletti:
I would say it should be safe to disable as:

- this is finding valid instances of syntax and suggesting rephrasing it to make it more explicit
- we have unit tests that should cover all edge cases of our code

Also from the [docs](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-assertion.md#when-not-to-use-it): 
> If your project does not use the strictNullChecks compiler option, this rule is likely useless to you. If your code is often wildly incorrect with respect to strict null-checking, your code may not yet be ready for this rule.

I'd say we are a 2/2 on when not to use it. 😉

acelletti (approved):

PR #1248, opened 2023-07-05
fix(data-grid): add scroll (VIV-1180)

rachelbt:


libs/components/src/lib/data-grid/data-grid-row.scss
```diff
@@ -31,7 +31,7 @@
 
 	:host([row-type="sticky-header"]) & {
 		position: sticky;
-		z-index: 10;
+		z-index: 9;
```
rinaok, line 34: Why did it change to 9?

rachelbt: popup gets 10, and we don't want select (that has popup) to be under the sticky header

rinaok (approved):

PR #1249, opened 2023-07-06
fix(text-area): fix base width (VIV-1061)

RichardHelm:


libs/components/src/lib/text-area/text-area.scss
```diff
@@ -29,7 +29,7 @@ $outline-width: 2px;
 	@include connotation.connotation (text-area);
 
 	display: inline-grid;
-	width: inherit;
+	width: 100%;
```
rachelbt, line 32:
```suggestion
	inline-size: 100%;
```
if we are changing - lets ditch the width :)

rachelbt (approved):
LGTM

libs/components/src/lib/text-area/text-area.scss
RichardHelm: Sure, updated 👍

rachelbt (approved):

PR #1250, opened 2023-07-06
chore(ci): extract coverage upload into separate job (VIV-1136)

RichardHelm:


.github/workflows/pull-request.yml
```diff
@@ -46,11 +46,18 @@ jobs:
...
       os: ubuntu-latest
       report_coverage: true
 
+  upload-coverage:
+    if: github.event.pull_request.draft == false
+    needs: unit-tests
+    uses: vonage/vivid-3/.github/workflows/_upload-coverage.yml@039a315116ee99695d3f8f2dcdd245282aa5f02a
+    with:
+      os: ubuntu-latest
+
```
YonatanKra, line 60: We can move to relative paths so we wouldn't need to state the branch/commit hash.
See here:
https://github.com/Vonage/vivid-3/blob/main/.github/workflows/release-please.yml#L90

YonatanKra (changes requested):
Great job!
Thanks!
Just a small change :)

.github/workflows/pull-request.yml
RichardHelm: Ah, that's much better
Updated 👍

YonatanKra (approved):
Great stuff! Thanks!

PR #1253, opened 2023-07-09
fix(tabs): tabs now scroll only inside the tabs (VIV-1191)

YonatanKra:


rachelbt (commented):
if you updated the ui-test - how come there's no new png's? :)

YonatanKra:
> if you updated the ui-test - how come there's no new png's? :)

Because I didn't create a new PNG. This is a behavioral test that doesn't require a screenshot.

PR #1254, opened 2023-07-09
fix(docs): prevent caching when there's a bad fetch (VIV-1195)

YonatanKra:


rinaok (changes requested):
Screenshot 2023-07-09 at 19 06 52


There is a problem loading fonts and pictures

YonatanKra:
> Screenshot 2023-07-09 at 19 06 52
> There is a problem loading fonts and pictures

Thanks. Strangely it works locally. Looking into it. Thanks again :)

YonatanKra:
@rinaok - can you recheck? Looks good on my end. Then again, it looked good before so... ;)

rinaok:
> icon.resources.vonage.com

@YonatanKra Yes, now it is loading :)

apps/docs/assets/bundled-scripts/sw.js
```diff
@@ -2,14 +2,19 @@ const VERSION = 'SW_VERSION';
 
 console.log(`Vivid Upgraded to Version ${VERSION}!`);
 
+const iconsUrl = 'icon.resources.vonage.com';
```
rinaok, line 5: Is it possible to make this a global url? It appears in multiple places in the code.

YonatanKra: It seems like a bigger issue due to how we our build works. I'll sort it out in a different [issue](https://jira.vonage.com/browse/VIV-1198).

rinaok: You can add it to this PR: https://jira.vonage.com/browse/VIV-1128

rinaok (approved):

PR #1256, opened 2023-07-10
fix(dialog): fix modal dialog not being fixed (VIV-1193)

RichardHelm:


libs/components/src/lib/dialog/dialog.scss
```diff
@@ -26,13 +26,14 @@ $dialog-space-size: 24px;
...
 
-	&:not(.modal) {
+	&:not(:modal) {
 		z-index: var(#{variables.$dialog-z-index}, 1);
 	}
 
-	&.modal {
-		// reason for box-shadow & hardcoded color - PR-1172
+	&:modal {
+		position: fixed;
```
rachelbt, line 34: we dont need this - the dialog is natively has position:fixed when it is opened with showModal.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal
We have a polypill in the dialog which I suspect causing the issue.
@YonatanKra what do you think?

RichardHelm: You're right the polyfill is setting `position: absolute`

But even without the polyfill we have `elevation` overriding the position as well:
image

rachelbt: oh - thats a shame :(

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -26,13 +26,14 @@ $dialog-space-size: 24px;
...
 	}
 
-	&:not(.modal) {
+	&:not(:modal) {
 		z-index: var(#{variables.$dialog-z-index}, 1);
 	}
 
-	&.modal {
-		// reason for box-shadow & hardcoded color - PR-1172
+	&:modal {
```
rachelbt, line 33: I think we can stick with both of them, to make sure we are covered in all browsers.
Lets rty this:
&.modal,
&:modal {....}

libs/components/src/lib/dialog/dialog.scss
```diff
@@ -26,13 +26,14 @@ $dialog-space-size: 24px;
 		#{variables.$dialog-default-max-inline-size}: 560px;
 	}
 
-	&:not(.modal) {
+	&:not(:modal) {
```
rachelbt, line 29: I think we can stick with both of them, to make sure we are covered in all browsers.
Lets rty this:

&:not(:modal, .modal){....}

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -131,11 +131,6 @@ describe('vwc-dialog', () => {
 			await showModalDialog();
 			expect(getBaseElement(element).hasAttribute('open')).toEqual(true);
 		});
-
```
rachelbt, line 134: please bring tis back :)

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -135,9 +135,9 @@ export class Dialog extends FoundationElement {
 	}
 
 	showModal() {
+		this.#handleModal(true);
 		this.#dialog.showModal();
 		this.open = true;
-		this.#handleModal(true);
```
RichardHelm, line 140: Needs to add `modal` class before `showModal()` or it would scroll to the top anyway

libs/components/src/lib/dialog/dialog.scss
RichardHelm: Ok, but let's just revert it then as there's no point in having two paths

libs/components/src/lib/dialog/dialog.scss
YonatanKra: Dialog is supported in all modern browsers for more than a year.
Our policy so far had been to maintain support up to two years back so we still need the polyfill.

I see now that while the JS code of the polyfill is not loaded, the CSS does. This can be solved by adding the same condition in the definition file IMO:
```
export function isSupportingDialog() {
	return HTMLDialogElement && HTMLDialogElement.prototype.showModal;
}
```
Then in the definition file do this:
```
export const dialogDefinition = Dialog.compose<FoundationElementDefinition>({
	baseName: 'dialog',
	template: template as any,
	styles: !isSupportingDialog() ? [styles, dialogPolyfillStyles] : [styles],
});
```
This way the polyfill will truly work only in unsupported browsers.

RichardHelm: Sure, done 👍

rachelbt (dismissed):
LGTM

libs/components/src/lib/dialog/dialog.scss
YonatanKra: I think this makes the `scss` change obsolete? Or do we need it for older browsers?

RichardHelm: No we still need it, because vwc-elevation overrides the `position`

YonatanKra (approved):
LGTM. Thanks!

PR #1259, opened 2023-07-11
fix(popup): add namespace to open and close events (VIV-1198)

rinaok:


libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -235,7 +235,7 @@ describe('vwc-menu', () => {
 	describe('events', () => {
 		it('should fire an event on popup open', async () => {
```
YonatanKra, line 236: I actually don't understand this test.
Does it test the `popup`? What does the menu need to do in the case of `open`?

libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -245,7 +245,7 @@ describe('vwc-menu', () => {
 		it('should fire an event on popup close', async () => {
```
YonatanKra, line 245: Again, it seems like this test tests the `popup` and not the `menu`.
What does the menu need to do in the case of `popup:close`?

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -148,7 +148,7 @@ describe('vwc-popup', () => {
 
 		it('should fire open & close events', async function () {
```
YonatanKra, line 149: This tests only `open`. 
```suggestion
		it('should fire vwc-popup:open event', async function () {
```

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -169,7 +169,7 @@ describe('vwc-popup', () => {
 		it('should fire close event', async function () {
```
YonatanKra, line 169:
```suggestion
		it('should fire "vwc-popup:close" event', async function () {
```

YonatanKra (changes requested):
Really small changes.
Thanks!

libs/components/src/lib/menu/menu.spec.ts
rinaok: Updated the test

libs/components/src/lib/menu/menu.spec.ts
rinaok: Updated the test

PR #1260, opened 2023-07-12
fix(docs): refine docs (VIV-1197)

rinaok:


docs/design/foundations.md
```diff
@@ -1,34 +1,76 @@
...
+ 
+<p><vwc-note icon="info-solid" connotation="information" headline="Components">
+Vivid Components are interactive building blocks that implement Vonages design system.  
+VVC (vivid-components) are used for creating user interfaces and include a built-in system for connotations, states, and user interactions.
+</vwc-note></p>
 
-<hr>
+<p><vwc-note icon="info-solid" connotation="information" headline="Web components">
+  Web Components is a suite of different technologies allowing you to create reusable custom elements - with their functionality encapsulated away from the rest of your code - and utilize them in your web apps. Vivid Web Components (VWC) are a collection of Web Components maintained by Vivid team.
+</vwc-note></p>
```
YonatanKra, line 21: I don't understand this part. What are the `Core Variables`? How do I use them? Do I need to use them? If not, do I need to know about them? Is it meant for a consumer? A Vivid developer/contributor?

docs/design/foundations.md
```diff
@@ -1,34 +1,76 @@
...
 
-Most Templates and Pages will be built and design by the product design teams based on vivid components.
- <hr>
- 
-## Components
-Vivid Components are interactive building blocks that implement Vonages design system.  
-VVC (vivid-components) are used for creating user interfaces and include a built-in system for connotations, states, and user interactions.
+Design tokens are directly integrated into our component libraries and UI kits. They cover the various options of platform scales, color themes, component states, and more.
+
+Using design tokens allows us to manage and maintain our design system as Vonage’s design Single source of truth.
```
YonatanKra, line 32: These are 3 paragraphs that repeat the same idea: "What are design tokens?"

``` suggest
Design tokens are a set of values that maintain a design system. They represent anything defined by design: font size in pixel, a color as a RGB value, an opacity as a number, etc. They’re used instead of hard-coded values to ensure flexibility and unity across all product experiences.

Design tokens are directly integrated into our component libraries and UI kits and are the design's Single Source of Truth.
```

YonatanKra (changes requested):
A small clarification and suggestion.

docs/design/foundations.md
rinaok: @YonatanKra I think that we should remove all of this paragraph, but was afraid to erase everything lol. If you agree I can remove it, cause in my opinion nobody will read it.

docs/design/foundations.md
rinaok: @YonatanKra I think we can stay with this one and remove everything else:

`Design tokens are all the values needed to construct and maintain a design system — color, typography, spacing, object styles, sizing, etc.

Using design tokens allows us to manage and maintain our design system as Vonage’s design Single source of truth.`

WDYT?

docs/design/foundations.md
YonatanKra: Let's remove it. It means nothing to both of us :)

docs/design/foundations.md
YonatanKra: I don't like this phrasing: `Design tokens are all the values needed to construct and maintain a design system`
This one is a bit more straightforward: `Design tokens are a set of values that maintain a design system.`
(I even asked Grammarly ;) )
IMO, breaking two ideas into two sentences is better than combining them into one less understandable one.
The 2nd sentence in your version repeats on `maintain` again, which IMO is less preferable. I tried to reduce repeatition and add some more info - or go over the same one in a different way at least.
So I'll go with:
```
Design tokens are a set of values that maintain a design system. 
They represent anything defined by design: font size in pixel, a color as a RGB value, an opacity as a number, etc. 

The Design Tokens are the design system's Single Source of Truth.
```
WDYT?

YonatanKra: Definitely.
Maybe gradients belong to Colors and design tokens need to be on its own page?

rinaok: I changed it to Guidelines instead of Foundations cause it's guidelines about design-tokens.

YonatanKra (approved):
LGTM. Thanks!

PR #1268, opened 2023-07-13
fix(data-grid): prevent non-enumarable in columndefs (VIV-1208)

YonatanKra:


libs/components/src/lib/data-grid/data-grid.spec.ts
```diff
@@ -85,6 +85,42 @@ describe('vwc-data-grid', () => {
...
+
+			const expectedColumnDef = [
+				{'columnDataKey': 'id', 'gridColumn': '0'}, {'columnDataKey': 'name', 'gridColumn': '1'},
+				{'columnDataKey': 'age', 'gridColumn': '2'}, {'columnDataKey': 'address.city', 'gridColumn': '3'}
+			];
+
+			expect(DataGrid.generateColumns(rowData)).toEqual(expectedColumnDef);
+		});
+
+		it('should omit enumarable properties from columnDefs', function () {
```
RichardHelm, line 105:
```suggestion
		it('should omit non-enumerable properties from columnDefs', function () {
```

libs/components/src/lib/data-grid/data-grid.spec.ts
```diff
@@ -85,6 +85,42 @@ describe('vwc-data-grid', () => {
...
+			const rowData = {
+				id: '1',
+				name: 'Person 1',
+				age: 20,
+				'address.city': 'City 1',
+			};
+			Object.defineProperty(rowData, 'enumarable',{
```
RichardHelm, line 112:
```suggestion
			Object.defineProperty(rowData, 'nonEnumerable',{
```

PR #1269, opened 2023-07-13
chore(menu): refactor (VIV-1204)

YonatanKra:
Follow the commits


libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -109,35 +85,31 @@ describe('vwc-menu', () => {
...
+	describe('focus', () => {
+		function createMenuItem(type = 'menuitem') {
 			const div = document.createElement('div');
-			div.setAttribute('role', 'menuitem');
+			div.setAttribute('role', type);
 			element.appendChild(div);
+			return div;
+		}
+		it('should focus the first menuitem in the menu', async () => {
+			const div = createMenuItem();
```
rachelbt, line 112: the fact that this is in the HTML does not test that it is focused. Checking the menu manually discovers that there's no way to get to the menu items if you don't have hands and eyes :) :(

libs/components/src/lib/menu/menu.template.ts
```diff
@@ -24,16 +24,17 @@ export const MenuTemplate: (
 ) => ElementViewTemplate = (context: ElementDefinitionContext) => {
 	const popupTag = context.tagFor(Popup);
 
+	function handlePopupEvents(x: Menu, state: boolean) {
+		x.open = state;
+	}
 	return html<Menu>`
-		<template
```
rachelbt, line 28: 👍

rachelbt (commented):
nice idea of commits as notes (though I like also simple bullets in the PR comments :) )

libs/components/src/lib/menu/menu.spec.ts
YonatanKra: It doesn't check it is in the HTML. It checks that this is the focused (`document.activeElement`).

rinaok (approved):

rachelbt (approved):
LGTM

PR #1270, opened 2023-07-16
feat(text field): add action-items slot (VIV-1188)

rachelbt:


libs/components/src/lib/text-field/text-field.template.ts
```diff
@@ -29,7 +30,8 @@ const getStateClasses = ({
 	[`appearance-${appearance}`, Boolean(appearance)],
 	[`shape-${shape}`, Boolean(shape)],
 	['no-label', !label],
-	['success connotation-success', Boolean(successText)]
+	['success connotation-success', Boolean(successText)],
+	['hide-footer', !(actionItemsSlottedContent?.length)]
```
YonatanKra, line 34: This one's for hiding the footer, and we have another `hide` for the wrapper?
Anyway, can you please add a test for this too?

libs/components/src/lib/text-field/ui.test.ts
```diff
@@ -34,6 +34,8 @@ test('should show the component', async ({page}: { page: Page }) => {
 
 	await page.waitForLoadState('networkidle');
 
+	//await page.pause();
```
YonatanKra, line 37: Please remove this line

libs/components/src/lib/text-field/text-field.template.ts
rachelbt: this is my stupid copy paste....
I'll remove this one.

YonatanKra (approved):
Code's solid :)
If @AyalaBu is OK with the result, we can merge this.

PR #1271, opened 2023-07-17
fix(components): fix wrong appearance definition (VIV-1210)

rachelbt:


YonatanKra:
Thanks for handling this.
Please add a UI test that covers this.

rachelbt:
> Thanks for handling this. Please add a UI test that covers this.

there are both unit-test and UI test. 
image
Its like - it doesn't matter which layout is in the TS. The one that is in the css file is the one that counts...
Like - in text-field there was Layout.Outline - but it was not outlined but fieldset!

YonatanKra:
> there are both unit-test and UI test. image Its like - it doesn't matter which layout is in the TS. The one that is in the css file is the one that counts... Like - in text-field there was Layout.Outline - but it was not outlined but fieldset!

So one would expect the examples to change now that you fixed it, no?

YonatanKra (approved):
I see it's typing issues.
Thanks for correcting it :)

PR #1272, opened 2023-07-17
feat(menu-item): allow default action when role is presentation (VIV-1183)

YonatanKra:


libs/components/src/lib/menu-item/README.md
```diff
@@ -185,3 +192,20 @@ vwc-menu {
...
+2. Wrap the `vwc-menu-item` in an anchor tag.
+3. Set the `role` attribute to `menuitem` on the anchor tag.
+
+If you are using a framework, just wrap the menu item in any routing component/directive as done with the anchor tag.
+
+```html preview
+	<a role="menuitem" href="https://www.vonage.com" target="_blank" rel="noopener noreferrer">
+  	<vwc-menu-item role="presentation" text="Go to Vonage"></vwc-menu-item>
+	</a>
+```
```
rinaok, line 211: 1. The example should be inside a menu like all the other examples.
2. How the user can navigate and click on the menu-item if the menu is not focusable?

rinaok (changes requested):

libs/components/src/lib/menu-item/README.md
```diff
@@ -71,6 +71,9 @@ set `role` to change the role of the menu item
   <vwc-divider></vwc-divider>
   <vwc-menu-item role="menuitemradio" text="Radio 2.1"></vwc-menu-item>
   <vwc-menu-item role="menuitemradio" text="Radio 2.2"></vwc-menu-item>
+	<a role="menuitem" href="https://www.vonage.com" target="_blank" rel="noopener noreferrer">
+  	<vwc-menu-item role="presentation" text="Go to Vonage"></vwc-menu-item>
+	</a>
```
rinaok, line 76: Screenshot 2023-07-17 at 16 19 36

Can't see the example

rinaok (changes requested):

libs/components/src/lib/menu-item/README.md
YonatanKra: 1. Sure
2. You are right. It works, but the focus needs to be corrected visually.
Thanks!

rachelbt (commented):
we have some issues with this:
1. we can't force style on the a - and we are counting on devs to add text-decoration: none to the a.
2. We have as you said - focus from the A and not from the menu item.
:(

rinaok (approved):

PR #1274, opened 2023-07-18
chore(docs): css variable (VIV-1057)

rinaok:


apps/docs/assets/styles/main.css
```diff
@@ -352,3 +351,23 @@ article.article > h1 {
...
+
+  tbody > tr {
+    border-bottom: 1px solid var(--vvd-color-neutral-100);
+  }
+
+  td,
+  th {
+    max-inline-size: 80ch;
+    padding: 8px 16px;
+  }
```
rachelbt, line 372: this file is css and not scss - so this all need to be after the table close tag

rachelbt: also you can add style:
td {
 word-break: normal;
text-align: left;
}

rachelbt (commented):
I went over the PR but not through all of it.
The pages look more clearer with the table!
I have some general comments:
1. I'm not sure about the table inside the elevation - it looks a bit strange.
2. we should wrap the table with a div with horizontal scroll so the pages will look OK in mobile
3. I'm not sure its correct to remove the default value of most of the css variables. When setting a css-variable with a value it usually changes the coded value - I think its important to know what is the initial value.

rinaok:
> I went over the PR but not through all of it. The pages look more clearer with the table! I have some general comments:
> 
> 1. I'm not sure about the table inside the elevation - it looks a bit strange.
> 2. we should wrap the table with a div with horizontal scroll so the pages will look OK in mobile
> 3. I'm not sure its correct to remove the default value of most of the css variables. When setting a css-variable with a value it usually changes the coded value - I think its important to know what is the initial value.

@rachelbt 

1. I took it from [an old design we had in Figma](https://www.figma.com/file/EOh4rw71ku9dnRUXnhdfma/Vivid-showcase?node-id=756%3A144&mode=dev), you are welcome to change it as you wish :)
Screenshot 2023-07-18 at 16 43 58

2. I think that now after your changes it looks better in mobile even without the scroll, WDYT?
3. Fixed.

libs/components/src/lib/alert/README.md
```diff
@@ -206,19 +205,19 @@ Use the `--alert-min-inline-size` variable to set the alert's minimum inline siz
 
 Use the `--alert-max-inline-size` variable to set the alert's maximum inline size.
 
-- Type: [`<length>`](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
-- Default: `fit-content`
```
rachelbt, line 210: please add:
`Default: `fit-content``

libs/components/src/lib/dialog/README.md
```diff
@@ -341,85 +325,25 @@ vwc-dialog {
...
+<vwc-elevation>
 
-Shows the dialog. Positioned in a top position by default.
+| Name| Returns  | Description  |
+| ----------- | ------- | ---------------- |
+| `show`      | `void`  | Shows the dialog. Positioned in a top position by default. |
+| `close`     | `void`  | Closes the dialog.       |
+| `showModal` | `void`  | Shows the dialog and makes it the top-most modal dialog. Positioned in a center position by default. Interaction outside the dialog is blocked and the content outside it is rendered inert For more information, see the native [Dialog.showModal](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal).  |
 
-```html preview
```
rachelbt, line 374: I think we need to keep these examples (I use them regularly).
We can combine them into one with buttons for different use-case.
WDYT?

libs/components/src/lib/header/README.md
```diff
@@ -82,8 +82,6 @@ It is also possible to assign application context directly to the header's `app-
 
 The size of the header block is set definitively. A header's block size value is often used in conjunction with other elements in the application. The `--vvd-header-block-size` private custom property is applied internally to header styles and holds the block size value. This property isn't customizable by authors but does pierce in and can be set to apply style to assigned content.
 
-- Type: [`length`](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
-- Accessibility: `read-only`
```
rachelbt, line 86: `Accessibility: `read-only`
this is needed - since the variable can be used and its value will always be 64px.
See how we are using in the docs main.css.
maybe we need to add here a code sample for what is it for.

libs/components/src/lib/layout/README.md
```diff
@@ -203,9 +203,6 @@ Use the `gutters` attribute to set the initial preferred `margin`.
 
 Control the [grid-template-columns](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns) of the layout by setting `--layout-grid-template-columns`.
 
-- Type: `string`
-- Default: `undefined`
```
rachelbt, line 207: there's a default that this variable is running over.

libs/components/src/lib/layout/README.md
```diff
@@ -219,10 +216,6 @@ Control the [grid-template-columns](https://developer.mozilla.org/en-US/docs/Web
 
 Control the [grid-template-rows](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows) of the layout by setting `--layout-grid-template-rows`.  
 
-
-- Type: `string`
```
rachelbt, line 223: our default is auto

libs/components/src/lib/layout/README.md
```diff
@@ -241,9 +234,6 @@ Control the grid [column-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/c
 #### Layout-column-gap
 Control the grid [row-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap) by setting `--layout-row-gap`.
 
-- Type: `string`
```
rachelbt, line 244: we have a default gap - medium.

libs/components/src/lib/menu/README.md
```diff
@@ -108,13 +108,10 @@ While any DOM content is permissible as a child of the menu, only `vwc-menu-item
 
 ## CSS Variables
 
-### Max Inline Size
+### Maximum Inline Size
 
 Use the `--menu-max-inline-size` variable to set the menu's inline size.
 
-- Type: [`<length>`](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
-- Default: `auto`
```
rachelbt, line 116: I would keep the default

libs/components/src/lib/menu/README.md
```diff
@@ -127,13 +124,10 @@ Use the `--menu-max-inline-size` variable to set the menu's inline size.
 </vwc-menu>
 ```
 
-### Min Inline Size
+### Minimum Inline Size
 
 Use the `--menu-min-inline-size` variable to set the menu's inline size.
 
-- Type: [`<length>`](https://developer.mozilla.org/en-US/docs/Web/CSS/length)
-- Default: `auto`
```
rachelbt, line 135: same

libs/components/src/lib/pagination/README.md
```diff
@@ -67,17 +67,14 @@ Set the `navIcons` attribute to change the pagination's navigation buttons type.
...
-```
+| Name                | Description                                                                                  |
+| ------------------- | -------------------------------------------------------------------------------------------- |
+| `pagination-change` | Listen to the `pagination-change` event to get notified when the pagination's state changes. |
+|                     |
+
+
+</vwc-elevation>
+
+---
```
rachelbt, line 80:
```suggestion
```

libs/components/src/lib/tooltip/README.md
```diff
@@ -63,9 +63,6 @@ Use the `placement` attribute to set the placement of the tooltip around the anc
 
 Use the `--tooltip-inline-size` variable to set the tooltip's inline size.
 
-- Type: `Number`
-- Default: `auto`
```
rachelbt, line 67: keep the default please

libs/components/src/lib/accordion/README.md
```diff
@@ -55,9 +55,13 @@ Use the `expand-mode` attribute to control the expand mode of the accordion, eit
...
+| Name       | Returns | Description                                                                               |
+| ---------- | ------- | ----------------------------------------------------------------------------------------- |
+| `closeAll` | `void`  | When `expand-mode` is set to `multi`, closes all the accordion items from the open state. |
+|            |
 
-When `expand-mode` is set to `multi`, closes all the accordion items from the open state.
\ No newline at end of file
+</vwc-elevation>
+
+---
```
rachelbt, line 67:
```suggestion

```

libs/components/src/lib/alert/README.md
```diff
@@ -206,19 +205,19 @@ Use the `--alert-min-inline-size` variable to set the alert's minimum inline siz
...
+| Name    | Description                                       |
+| ------- | ------------------------------------------------- |
+| `open`  | The `open` event fires when the alert is opened.  |
+| `close` | The `close` event fires when the alert is closed. |
 
-### Close
+</vwc-elevation>
 
-The `close` event fires when the alert is closed.
+---
```
rachelbt, line 223:
```suggestion

```

libs/components/src/lib/calendar/README.md
```diff
@@ -68,9 +68,13 @@ Assign elements to a day column using `'day-0'` | `'day-1'` | `'day-2'` | `'day-
...
-- Returns: `{ day: number, hour: number }` | `null`
+| Name              | Returns                                  | Description                                                                             |
+| ----------------- | ---------------------------------------- | --------------------------------------------------------------------------------------- |
+| `getEventContext` | `{ day: number, hour: number }` \ `null` | Accepts event reference and returns an object containing _day_ & _hour_  if applicable. |
+|                   |
 
-Accepts event reference and returns an object containing _day_ & _hour_  if applicable.
+</vwc-elevation>
+
+---
```
rachelbt, line 80:
```suggestion

```

libs/components/src/lib/pagination/README.md
rachelbt, line 80:
```suggestion

```

libs/components/src/lib/tabs/README.md
```diff
@@ -170,6 +170,12 @@ Each `tab panel` has an associated [vwc-tab](../../components/tab) element, that
...
+<vwc-elevation>
 
-Fires a custom `change` event when a tab is clicked or during keyboard navigation.
+| Name     | Description                                                                        |
+| -------- | ---------------------------------------------------------------------------------- |
+| `change` | Fires a custom `change` event when a tab is clicked or during keyboard navigation. |
+
+</vwc-elevation>
+
+---
```
rachelbt, line 181:
```suggestion

```

libs/components/src/lib/tree-item/README.md
```diff
@@ -94,10 +94,13 @@ Use the `expanded` attribute to set the tree-item's open state.
...
+| Name              | Description                                                             |
+| ----------------- | ----------------------------------------------------------------------- |
+| `expanded-change` | Fires a custom `expanded-change` event when the expanded state changes. |
+| `selected-change` | Fires a custom `selected-change` event when an item has been selected.  |
 
-### Selected Change
+</vwc-elevation>
 
-Fires a custom 'selected-change' event when an item has been selected.
+---
```
rachelbt, line 106:
```suggestion

```

libs/components/src/lib/tag/README.md
```diff
@@ -135,20 +135,23 @@ Notice that it works only if the tag is `removable` and not `selectable`.
...
+| -------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `remove` | `void`  | Removes the tag from the DOM.  Fires the `removed` event and removes the tag from the DOM completely.  If you have a variable that refers to the tag element make sure to clear it otherwise it might cause a memory leak. |
 
-- Type: `function`
-- Returns: `void`
+</vwc-elevation>
 
-Removes the tag from the DOM.  Fires the `removed` event and removes the tag from the DOM completely.  If you have a variable that refers to the tag element make sure to clear it otherwise it might cause a memory leak.
\ No newline at end of file
+---
```
rachelbt, line 157:
```suggestion

```

rachelbt (commented):
NICE!
We still need a wrapper for the table - it still causing scroll in mobile

libs/components/src/lib/header/README.md
rinaok: There is a code example under the description

libs/components/src/lib/dialog/README.md
rinaok: added a use case for modal, for the regular show() - there is a form example

rinaok:
> NICE! We still need a wrapper for the table - it still causing scroll in mobile

@rachelbt 
I don't have scroll in mobile, can you send a screenshot?

IMG_79B195B41B96-1

libs/components/src/lib/pagination/README.md
```diff
@@ -67,17 +67,12 @@ Set the `navIcons` attribute to change the pagination's navigation buttons type.
 
 ## Events
 
-### pagination-change
-Listen to the `pagination-change` event to get notified when the pagination's state changes.
+<vwc-elevation>
 
-```html preview
```
rachelbt, line 73: isnt the JS here is helpful for dev? should we keep it as a JS preview?

libs/components/src/lib/dialog/README.md
```diff
@@ -432,19 +356,31 @@ Closes the dialog.
...
-</vwc-button>
+<script>
+  function openDialog() {
+    dialog = document.querySelector('vwc-dialog');
+    dialog.show();
+  }
+</script>
+```
+
+### Modal Dialog with Scrim
```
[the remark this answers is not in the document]
rinaok: @rachelbt https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/

rinaok:
> > > NICE! We still need a wrapper for the table - it still causing scroll in mobile
> > 
> > 
> > @rachelbt I don't have scroll in mobile, can you send a screenshot?
> > IMG_79B195B41B96-1
> 
> image

I don't have a scroll, if you know how to fix this I'll add it.

IMG_375622FEE4F3-1

libs/components/src/lib/dialog/README.md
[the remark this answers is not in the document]
rachelbt: sorry!!!

libs/components/src/lib/pagination/README.md
rinaok: @rachelbt I added a description now for the detail object. Tell me what you think, I think it's enough. I also think that the event should be named 'change' like in all other components, and not with prefix 'pagination-change'.

apps/docs/assets/styles/main.css
```diff
@@ -10,6 +10,12 @@ p {
   margin: unset;
 }
 
+code {
```
rachelbt, line 13: it makes the code so much more readable! 👍

PR #1275, opened 2023-07-18
fix(nav-disclosure): prevent toggle event from bubbling (VIV-1216)

RichardHelm:
nav-disclosure forwards the `toggle` event from the underlying `<details>`.
This event should, like the native one, not bubble. See https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event


libs/components/src/lib/nav-disclosure/README.md
```diff
@@ -71,11 +71,27 @@ If label is not applied.
...
+  <vwc-nav-disclosure id="disclosure" icon="profile">
+    <vwc-nav-item href="#" text="2nd level item"></vwc-nav-item>
+  </vwc-nav-disclosure>
+<vwc-nav>
+<script>
+  disclosure.addEventListener('toggle', (e) => {
+    output.innerHTML = `nav-disclosure is now ${e.target.open ? 'open' : 'closed'}`;
+  });
+</script>
+```
```
rinaok, line 95: There is no need to add an example for toggle in the documentation. Just a description is enough.

libs/components/src/lib/nav-disclosure/README.md
```diff
@@ -71,11 +71,27 @@ If label is not applied.
...
 
-## Accessibility
+## Events
 
-The nav-disclosure has a `role` button.
-Nav-disclosure has `aria-expanded` set to true when the content is visible. Otherwise, it is set to false.
-Nav-disclosure has a value specified for `aria-controls` that refers to the content.
+### Toggle
+
+The `toggle` event is dispatched when the open/closed state of the nav-disclosure is toggled.
```
YonatanKra, line 78: If it is dispatched and doesn't bubble, why should we care about it? Where does it stop?
Who dispatched the `Toggle` event?

libs/components/src/lib/nav-disclosure/nav-disclosure.spec.ts
```diff
@@ -76,16 +76,28 @@ describe('vwc-nav-disclosure', () => {
 		});
 	});
 
-	it('should remove toggle listener after disconnection', async function() {
-		const spy = jest.fn();
-		element.addEventListener('toggle', spy);
+	describe('toggle event', () => {
+		it('emits a toggle event that does not bubble when open/closed state is toggled', async function() {
```
YonatanKra, line 80:
```suggestion
		it('shuold emit a toggle event that does not bubble when open/closed state is toggled', async function() {
```

libs/components/src/lib/nav-disclosure/nav-disclosure.spec.ts
```diff
@@ -76,16 +76,28 @@ describe('vwc-nav-disclosure', () => {
...
-		element.disconnectedCallback();
-		element.connectedCallback();
+			element.details.dispatchEvent(new Event('toggle'));
 
-		element.details.dispatchEvent(new Event('toggle'));
+			expect((spy as any).mock.calls.length).toEqual(1);
+			expect((spy as any).mock.calls[0][0].bubbles).toBe(false);
+		});
+
+		it('should remove toggle listener after disconnection', async function() {
```
YonatanKra, line 90: This doesn't seem to test what is said in the description.
It adds a listener to the element and then dispatches the event.
The expectation is that an event that was fired actually fired.
I think this `it` can be removed. Am I correct?

libs/components/src/lib/nav-disclosure/README.md
RichardHelm: Sure, removed 👍

libs/components/src/lib/nav-disclosure/nav-disclosure.spec.ts
RichardHelm: I did not write this test, but sure, we can remove it

rinaok (approved):

PR #1279, opened 2023-07-19
fix(elevation, listbox): remove export from underlying components (VIV-1221)

rinaok:


YonatanKra (changes requested):
Let's just remove them and be done with it.
If we ever need them back, we'll add them ourselves.

PR #1280, opened 2023-07-20
fix(text-filed): nicer visual focus with slotted action items (VIV-1224)

rachelbt:


rachelbt:
> There's something wrong with the Firefox snapshots

yeh I know - didn't want to regenerate all.
Waiting for @AyalaBu and then I'll open this to code-review and I'll fix the png :)
Thanks!

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -638,27 +638,18 @@ describe('vwc-text-field', () => {
...
-		it('should remove hide class if action-items is slotted', async function () {
+		it('should add action-items class if action-items is slotted', async function () {
 			const slottedElement = document.createElement('div');
 			slottedElement.slot = 'action-items';
 			slottedElement.id = 'action-items';
 			element.appendChild(slottedElement);
 			await elementUpdated(element);
 
-			const divWrapper = element.shadowRoot?.querySelector('.action-items-wrapper');
+			const baseElementHasNoHeader = element.shadowRoot?.
```
YonatanKra, line 648: `baseElementHasNoHeader` - why is that named like this? What's it supposed to be?

YonatanKra (changes requested):

YonatanKra (approved):

PR #1284, opened 2023-07-20
feat(file-picker): add file-picker component (VIV-962)

RichardHelm:
I've made the following changes from the docs:
- Remove `multiple` as it is covered by `maxFiles`
- `files` only returns accepted files
- Remove the localisation strings, as we're planning to add proper localisation support to vivid soon anyways


libs/components/src/lib/file-picker/file-picker.template.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const FilePickerTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<FilePicker> = (context: ElementDefinitionContext) => {
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<FilePicker>`
+		${x => {x.setButtonTag(context.tagFor(Button));}}
+		<div class="base" aria-label="${x => x.label}">
```
rinaok, line 36: What happens if the user adds only an `aria-label`? 

Screenshot 2023-07-24 at 13 10 59

libs/components/src/lib/file-picker/file-picker.template.ts
```diff
@@ -0,0 +1,47 @@
...
+			${when(x => x.label, html<FilePicker>`<label>${x => x.label}</label>`)}
+			<div class="${getClasses}" tabindex="0" role="button"
+					 @keydown="${(x, c) => x.handleKeydown(c.event as KeyboardEvent)}">
+				<slot class="main"></slot>
```
rinaok, line 40: Maybe we should have a Placeholder instead of a slot like in other components?  https://vivid.deno.dev/components/text-field#placeholder

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+		it('should allow accessing the component in unmounted state and mounting later without error', async () => {
+			const unmountedElement = document.createElement(COMPONENT_TAG) as FilePicker;
+			unmountedElement.maxFileSize = 256;
+			unmountedElement.maxFiles = 1;
+			unmountedElement.accept = '.jpg';
+			expect(unmountedElement.files).toEqual([]);
+
+			document.body.appendChild(unmountedElement);
+			document.body.removeChild(unmountedElement);
+		});
```
rinaok, line 42: What should it expect at the end?

libs/components/src/lib/file-picker/file-picker.scss
```diff
@@ -0,0 +1,123 @@
+@use "../../../../../dist/libs/tokens/scss/tokens.constants" as constants;
+@use "../../../../../dist/libs/tokens/scss/size.variables" as size;
+
+$low-ink-color: --_low-ink-color;
+
+:host {
+	display: block;
+	min-inline-size : 400px;
```
rinaok, line 8:
```suggestion
	inline-size: 400px;
```

libs/components/src/lib/file-picker/file-picker.ts
```diff
@@ -0,0 +1,205 @@
...
+		super.disconnectedCallback();
+		this.#dropzone!.destroy();
+	}
+	/**
+	 * Used internally to set the button tag.
+	 * @internal
+	 */
+	setButtonTag(tag: string) {
+		this.buttonTag = tag;
+	}
```
rinaok, line 176: Can it be private if it is only used internally?

libs/components/src/lib/file-picker/ui.test.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test } from '@playwright/test';
+import type { Page } from '@playwright/test';
+import {
+	loadComponents,
+	loadTemplate,
+} from '../../visual-tests/visual-tests-utils.js';
+
+const components = ['file-picker', 'button', 'layout', 'divider'];
```
rinaok, line 8:
```suggestion
const components = ['file-picker', 'button', 'layout', 'text-field'];
```

apps/docs/_data/components.json
```diff
@@ -457,5 +457,16 @@
...
 		]
+	},
+	{
+		"title": "file-picker",
+		"status": "alpha",
+		"markdown": "./libs/components/src/lib/file-picker/README.md",
+		"modules": [
+			"/assets/modules/components/file-picker/index.js",
+			"/assets/modules/components/button/index.js",
+			"/assets/modules/components/divider/index.js",
```
rinaok, line 468:
```suggestion
			"/assets/modules/components/text-field/index.js",
```

rinaok (changes requested):
The validation message (https://vivid.deno.dev/components/text-field#validation) is missing.
If the file picker is empty when you click submit, but you must choose a file, what happens?

libs/components/src/lib/file-picker/file-picker.template.ts
```diff
@@ -0,0 +1,47 @@
...
+	'dz-default',
+	'dz-message',
+	[`size-${size}`, Boolean(size)],
+);
+
+
+/**
+ * The template for the FilePicker component.
+ *
+ * @param context
```
rinaok, line 25: Screenshot 2023-07-24 at 14 00 34

libs/components/src/lib/file-picker/file-picker.template.ts
RichardHelm: When putting `aria-label` on vwc-file-picker element:
image

I removed adding it to base as well, as we would then have 2 `aria-label`

libs/components/src/lib/file-picker/file-picker.template.ts
RichardHelm: So when I did the `empty-state` we decided to not put any default text inside as it's not in line with other components, and I think the same reasoning would apply here

libs/components/src/lib/file-picker/file-picker.spec.ts
RichardHelm: Expecting no error, I updated the test to make it explicit

libs/components/src/lib/file-picker/file-picker.scss
RichardHelm: Oh I meant to put `max-inline-size`. I don't think it should have a fixed size

libs/components/src/lib/file-picker/file-picker.ts
RichardHelm: Unfortunately not as it's used by the template

RichardHelm:
@rinaok I made some changes, please have another look
I looked into the validation issue but because the file-picker is not a form element, it's difficult to add validation support

rinaok:
> @rinaok I made some changes, please have another look I looked into the validation issue but because the file-picker is not a form element, it's difficult to add validation support

Approved :) Can you just open a Jira issue to add a validation error to the file picker?

rinaok (dismissed):

RichardHelm:
> > @rinaok I made some changes, please have another look I looked into the validation issue but because the file-picker is not a form element, it's difficult to add validation support
> 
> @RichardHelm Approved :) Can you just open a Jira issue to add a validation error to the file picker?

Sure, let's use the old ticket we already have: https://jira.vonage.com/browse/VIV-802

rinaok (approved):

PR #1285, opened 2023-07-20
feat(menu): add header and action items slots (VIV-947)

rinaok:


libs/components/src/lib/menu/README.md
```diff
@@ -106,6 +116,65 @@ While any DOM content is permissible as a child of the menu, only `vwc-menu-item
...
+</style>
+
+<vwc-menu open>
+  <vwc-text-field slot="header" placeholder="Search" icon="search"></vwc-text-field>
+  <vwc-menu-item text="Menu item 1"></vwc-menu-item>
+  <vwc-menu-item text="Menu item 2"></vwc-menu-item>
+</vwc-menu>
+```
+
+### Action Items
```
rachelbt, line 137: why do we nned both action-items and footer?

libs/components/src/lib/menu/menu.scss
```diff
@@ -2,6 +2,12 @@
 @use "partials/variables" as variables;
 
 .base {
+	display: flex;
+	box-sizing: border-box;
+	flex-direction: column;
```
rachelbt, line 7:
```suggestion
	flex-direction: column;
	gap: 8px;
```

libs/components/src/lib/menu/menu.scss
```diff
@@ -22,3 +28,29 @@
 	outline: 2px solid currentColor;
 	outline-offset: -2px;
 }
+
+.header {
+	padding: 24px 24px 0;
```
rachelbt, line 33:
```suggestion
	padding-inline: 16px;
	padding-block-start: 8px;
```

libs/components/src/lib/menu/menu.scss
```diff
@@ -2,6 +2,12 @@
...
 .base {
+	display: flex;
+	box-sizing: border-box;
+	flex-direction: column;
+}
+
+.body {
 	max-block-size: var(#{variables.$menu-block-size}, 408px);
 	max-inline-size: var(#{variables.$menu-max-inline-size}, 100%);
 	min-inline-size: var(#{variables.$menu-min-inline-size});
```
rachelbt, line 13: I'm not sure this need to be here - this needs to be on the base

rachelbt: also the overflow

rachelbt: and remove the `padding-block: 8px;`

libs/components/src/lib/menu/menu.scss
```diff
@@ -22,3 +28,29 @@
...
+	.hide-header & {
+		display: none;
+	}
+}
+
+.footer {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	padding: 0 24px 24px;
```
rachelbt, line 44:
```suggestion
	padding-inline: 16px;
	padding-block-end: 8px;
```

libs/components/src/lib/menu/menu.template.ts
```diff
@@ -37,14 +54,27 @@ export const MenuTemplate: (
...
 				</div>
+				<div class="footer">
+					<div>
+						<slot name="footer" ${slotted('footerSlottedContent')}></slot>
+					</div>
+					<div class="actions">
+						<slot name="action-items" ${slotted('actionItemsSlottedContent')}></slot>
+					</div>
+				</div>
+			</div>
```
rachelbt, line 77: I don't think this is needed - we only need a footer slot.
maybe a validation components for cancel submit is needed...\

libs/components/src/lib/menu/README.md
```diff
@@ -106,6 +116,65 @@ While any DOM content is permissible as a child of the menu, only `vwc-menu-item
...
+<style>
+  html { /* for demo purposes */
+		block-size: 200px;
+	}
+</style>
+
+<vwc-menu open>
+  <vwc-menu-item text="Menu item 1"></vwc-menu-item>
+  <vwc-menu-item text="Menu item 2"></vwc-menu-item>
+  <vwc-checkbox slot="footer" label="I agree"></vwc-checkbox>
```
rachelbt, line 172:
```suggestion
```
Not sure this has to do with the menu-items :)

rachelbt (commented):
went over the template and the css - left some comments :)

libs/components/src/lib/menu/README.md
rinaok: Similar to dialog. The examples all had action items at the bottom. Do you think we don't need it?

libs/components/src/lib/menu/menu.template.ts
rinaok: @rachelbt But the same as in dialog all the examples have two action-buttons at the bottom :( 
https://docs.google.com/document/d/1Ag5Nmu53-v6YgqWpN_Uz_K4-KSOYWzRGLhWqrvT6GXU/edit#heading=h.36st4rl33pcx

libs/components/src/lib/menu/README.md
rinaok: @rachelbt Just wanted to show an example for the footer combined with the action buttons, do you have another idea?

libs/components/src/lib/menu/menu.scss
```diff
@@ -2,12 +2,15 @@
...
+	display: flex;
+	box-sizing: border-box;
+	flex-direction: column;
+	gap: 8px;
 	max-block-size: var(#{variables.$menu-block-size}, 408px);
 	max-inline-size: var(#{variables.$menu-max-inline-size}, 100%);
 	min-inline-size: var(#{variables.$menu-min-inline-size});
 	overflow-x: hidden;
 	overflow-y: auto;
-	padding-block: 8px;
```
RichardHelm, line 10: The menu without header/footer is now missing the padding: 
image

RichardHelm (changes requested):

libs/components/src/lib/menu/README.md
```diff
@@ -188,3 +257,27 @@ Menu component is a low level element, unaware of its document context, but is,
 A common practice used in apps / frameworks to promote a Menu component to top other elements z-axis, is to utilise a service that dynamically appends a Menu component to the end of the body element, when called for.
 
 This helps ensure elements don't render over top a Menu undesirebly.
+
+## Use Cases
```
RichardHelm, line 261: Can I put arbitrary content in default slot? Like data-grid?
image

Can we add a use case for that please?

rinaok: @RichardHelm This is an example from vivid-2. We don't have checkboxes inside a data-grid in vivid-3. And also, I'm not sure if it is correct to do it. We have menu-items for checkboxes.

libs/components/src/lib/menu/README.md
rinaok: Updated to only action-items, removed the footer

libs/components/src/lib/menu/README.md
RichardHelm: It's just an example. What I mean is you are supposed to be able to put anything inside menu, right?
And if you do that, will it break accessibility?

So we should have an example for that.

E.g. in dashboard we have this (nav-items):
image

rachelbt (commented):
css + HTML looks good!
The only thing that worth considering in a different ticket is the scroll - if there's a need for sticky header or footer or both.

RichardHelm (approved):

PR #1287, opened 2023-07-23
chore(workspace): upgrade nx to 16.5.5 (VIV-1181)

YonatanKra:


libs/components/src/lib/accordion/ui.test.ts
```diff
@@ -18,8 +18,8 @@ test('should show the component', async ({ page }: { page: Page }) => {
 		''
 	);
 
-	page.setViewportSize({ width: 500, height: 720 });
-
+	await page.setViewportSize({ width: 500, height: 720 });
+	await page.pause();
```
rachelbt, line 22:
```suggestion

```

rachelbt: why is this needed? you didn't add this in the other ui tests

libs/components/src/lib/accordion/ui.test.ts
```diff
@@ -32,7 +32,7 @@ test('should show the component', async ({ page }: { page: Page }) => {
 	const testWrapper = await page.$('#wrapper');
 
 	await page.waitForLoadState('networkidle');
-
+	await page.evaluate(() => document.fonts.ready);
```
rachelbt, line 35: why only this has it and not all other components?

libs/components/src/lib/action-group/action-group.ts
```diff
@@ -60,7 +60,7 @@ export class ActionGroup extends FoundationElement {
 		mode: 'boolean',
 	}) tight = false;
 
-	@attr() role: string | undefined;
+	@attr() override role: string | null = null;
 	@attr({attribute: 'aria-label'}) override ariaLabel: string | null = null;
 }
```
rachelbt, line 65: action-group is missing the safari png

rachelbt (commented):
Left you 2 comments.
went over all the snap shote - all seems OK. Spezia is smaller and more condensed. 
I looked at the other files - there are a lot! guess its upgrading all those files.
I think it is OK.
The docs look OK

rachelbt (commented):
Went over all the PNG's they look OK. The font had changed to Spezia and render slightly different in the terms of stretch and line height.
UI - wise it looks OK.

RichardHelm (approved):

PR #1290, opened 2023-07-26
fix(nav-item): word-break (VIV-1238)

rinaok:


libs/components/src/lib/nav/README.md
```diff
@@ -63,3 +63,43 @@ When the nav has focus:
...
+### Navigation inside a Side Drawer
+
+```html preview full
+<style>
+  html { /* for demo purposes */
+    block-size: 250px;
+  }
+  vwc-nav-item {
+    padding: 4px;
+  }
```
rachelbt, line 78:
```suggestion

```

libs/components/src/lib/nav/README.md
```diff
@@ -63,3 +63,43 @@ When the nav has focus:
...
+  vwc-fab {
+    position: fixed;
+    inset: auto auto 8px 8px;
+    z-index: 2;
+  }
+</style>
+
+<vwc-side-drawer id="sideDrawer" alternate open>
```
rachelbt, line 86:
```suggestion
<vwc-side-drawer id="sideDrawer" alternate open>
<vwc-layout gutters="small" column-basis="block">
```

libs/components/src/lib/nav/README.md
```diff
@@ -63,3 +63,43 @@ When the nav has focus:
...
+  <vwc-nav id="sideNav">
+      <vwc-nav-item href="#" text="Calls" icon="call-line" data-value="Calls" onclick="onClick(this)" aria-current="page"></vwc-nav-item>
+      <vwc-nav-item href="#" text="Voicemail" icon="voicemail-line" data-value="Voicemail" onclick="onClick(this)"></vwc-nav-item>
+      <vwc-nav-item href="#" text="SMS" icon="chat-line" data-value="SMS" onclick="onClick(this)"></vwc-nav-item>
+  </vwc-nav>
```
rachelbt, line 91:
```suggestion
  </vwc-nav>
  </vwc-layout>
```

libs/components/src/lib/nav/ui.test.ts
```diff
@@ -1,17 +1,25 @@
...
 	loadTemplate,
 } from '../../visual-tests/visual-tests-utils.js';
 
-const components = ['nav', 'nav-item'];
+const components = ['nav', 'nav-item', 'badge', 'icon'];
 
 test('should show the component', async ({ page }: { page: Page }) => {
-	const template = extractHTMLBlocksFromReadme(path.join(new URL('.', import.meta.url).pathname, 'README.md'))
-		.reduce((htmlString: string, block: string) => `${htmlString} <div style="margin: 5px;">${block}</div>`, '');
+	const template = `<vwc-nav>
```
rachelbt, line 11: please add the margin

libs/components/src/lib/nav/ui.test.ts
```diff
@@ -1,17 +1,25 @@
...
+	  </vwc-nav-item>
+	<vwc-nav-item href="#" icon="ai" text="AI Studio" onclick="onClick(this)">
+		  <vwc-badge slot="meta" text="new" connotation="success" appearance="subtle" shape="pill"></vwc-badge>
+	  </vwc-nav-item>
+	  <vwc-nav-item href="#" icon="books-line" text="Documentation" onclick="onClick(this)">
+		  <vwc-icon slot="meta" name="open-solid"></vwc-icon>
+	  </vwc-nav-item>
+  </vwc-nav>`;
 
 	page.setViewportSize({ width: 200, height: 720 });
```
rachelbt, line 24: 200px is not enough - see the image is cut at the right?

rachelbt (commented):
I liked the use cases that you added - like the use of the nav inside side-drawer in the nav readme.

I'm wondering if its a breaking change - if there's a menu using long words that used to break now they dont.... I really don't know why Yinon added that in the first place. On the other hand - it is quite stupid here.... 

Regardless of the word-break, We can also do what we do in menu-item - by default its one line (can be more by default - like - auto) but you can set a css-var to set how many lines you want:
https://vivid.deno.dev/components/menu-item#css-variables

I'm not sure about the example in the side-drawer though:
- first there's a jump when opening and closing - this need to be seen by Ayala
- we need to add aria-label if there's no text (this one is an easy fix)
- what if there's a nav-disclosure - we have an issue with the selected state: https://jira.vonage.com/browse/VIV-971
I like the idea that we provide a sample for this use case - but I feel its not complete because of the way our components are (padding is different when only icon, the only text manipulation is either have it or not and more)

rachelbt (approved):
LGTM
@AyalaBu what do you think about this use-case example?
https://vivid-zghvm7sg39a0.deno.dev/components/side-drawer#collapsible-side-drawer

AyalaBu:
Something about the close/open animation looks weird to me....
Sorry but I don't like it much

PR #1291, opened 2023-07-27
chore(ui-tests): use self-hosted Spezia font for ui-tests (VIV-1243)

YonatanKra:


rachelbt (commented):
you are missing 2 images in action-group - FF & Safari

PR #1294, opened 2023-08-01
fix(button): set title on button when is set on host (VIV-1249)

rachelbt:


libs/components/src/lib/button/README.md
```diff
@@ -171,6 +171,10 @@ The indicator is not displayed when using the `super-condensed` size.
 <vwc-button appearance="outlined" icon="check-line" label="outlined with icon" pending></vwc-button>
 ```
 
+## Accessibility 
+When setting an `aria-label`, `aria-labelledby` or `title`, mind that `aria-label`, `aria-labelledby` are better for a11y.  
+When both `aria-label` and `title` are set, `aria-label` text will be read by the assistive technologies and not `title`.
```
rinaok, line 176:
```suggestion
When setting an `aria-label`, `aria-labelledby` or 'title', keep in mind that `aria-label`, `aria-labelledby` are better for A11Y.  
The assistive technology will read the `aria-label` text rather than the `title` if both are set.
```

libs/components/src/lib/button/button.spec.ts
```diff
@@ -177,4 +179,22 @@ describe('vwc-button', () => {
 			expect(control).toBeInstanceOf(Element);
 		});
 	});
+	describe('aria-label', function () {
```
YonatanKra, line 182: Do we want to take the label from the `label` property if `aria-label` is not set?

YonatanKra (approved):
Looks good. Look at my suggestion. Can be in a different issue.

PR #1297, opened 2023-08-03
feat(split-button): add new component (VIV-786)

rinaok:


libs/components/src/lib/split-button/README.md
```diff
@@ -0,0 +1,191 @@
...
+Note: Icon, by its own, doesn't make a discernible text. An `aria-label`, `aria-labelledby` or `title` must be provided to ensure that the user can understand the indicator's purpose.
+
+- Type: `string`
+- Default: `chevron-down-line`
+
+```html preview
+<vwc-split-button appearance="filled" label='Split Indicator' split-indicator="more-vertical-solid" aria-label="Show more options"></vwc-split-button>
+```
+
+### Icon with Label
```
YonatanKra, line 92: This is not an API. It better fits into the `Use Cases` section IMO.

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+	describe('disabled', function () {
+		it('should set disabled class when disabled is true', async () => {
+			const appearance = 'filled';
+			(element as any).appearance = appearance;
+			element.toggleAttribute('disabled', true);
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.appearance-${appearance}.disabled`);
+			expect(control).toBeInstanceOf(Element);
```
YonatanKra, line 119: I think this is more explicit:
```suggestion
			const controlHasDisabledClass = getControlElement(element).classList.contains(`appearance-${appearance}.disabled`);
			expect(controlHasDisabledClass).toBeTruthy();
```

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	describe('size', () => {
+		it('sets correct internal size style', async () => {
+			const size = 'condensed';
+			(element as any).size = size;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.size-${size}`);
+			expect(control?.classList.contains(`size-${size}`)).toBeTruthy();
```
YonatanKra, line 107:
```suggestion
			const control = getControlElement(element);
			expect(control?.classList.contains(`size-${size}`)).toBeTruthy();
```

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	describe('appearance', () => {
+		it('sets correct internal appearance style', async () => {
+			const appearance = 'filled';
+			(element as any).appearance = appearance;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.appearance-${appearance}`);
+			expect(control).toBeInstanceOf(Element);
```
YonatanKra, line 96:
```suggestion
			const control = getControlElement(control); 
			expect(control.classList.contains(`appearance-${appearance}`).toBeTruthy();
```

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	describe('shape', () => {
+		it('sets correct internal shape style', async () => {
+			const shape = 'pill';
+			(element as any).shape = shape;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.shape-${shape}`);
+			expect(control).toBeInstanceOf(Element);
```
YonatanKra, line 85:
```suggestion
			const control = getControlElement(element); 
			expect(control.classList.contains(`shape-${shape}`).toBeTruthy();
```

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	describe('connotation', () => {
+		it('sets correct internal connotation style', async () => {
+			const connotation = 'cta';
+			(element as any).connotation = connotation;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector(`.control.connotation-${connotation}`);
+			expect(control).toBeInstanceOf(Element);
```
YonatanKra, line 74:
```suggestion
			const control = getControlElement(element); 
			expect(control.classList.contains(`connotation-${connotation}`).toBeTruthy();
```

libs/components/src/lib/split-button/split-button.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+		});
+	});
+
+	describe('label', () => {
+		it('set label property to node', async () => {
+			const label = 'lorem';
+			element.label = label;
+			await elementUpdated(element);
+
+			const control = element.shadowRoot?.querySelector('.control');
```
YonatanKra, line 62:
```suggestion
			const control = getControlElement(element);
```

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function actionButton(context: ElementDefinitionContext) {
+	const affixIconTemplate = affixIconTemplateFactory(context);
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<SplitButton>`
+	<button
+		${ref('_action')}
+		class="control ${getClasses}"
+		?autofocus="${(x) => x.autofocus}"
```
YonatanKra, line 27: Can you please add a test for this?

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+@use "../../../../../dist/libs/tokens/scss/size.variables" as size;
+
+.base {
+	display: inline-flex;
+}
+
+.control,
+.indicator {
+	@include connotation.connotation(split-button);
+	@include appearance.appearance;
```
rachelbt, line 28: isn't it better to set this on the base?

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+		/* stylelint-disable value-no-vendor-prefix */
+		display: -webkit-box;
+		/* stylelint-enable value-no-vendor-prefix */
+		overflow: hidden;
+		-webkit-box-orient: vertical;
+		inline-size: 100%;
+		-webkit-line-clamp: var(#{variables.$split-button-line-clamp}, 1);
+	}
+
+	&:not(.icon-only) {
```
rachelbt, line 53: icon-only class is never set

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<SplitButton>`
+	<button
+		${ref('_indicator')}
+		class="indicator ${getClasses}"
+		?autofocus="${(x) => x.autofocus}"
+		?disabled="${(x) => (x.disabled || x.disabledIndicator)}"
+		title="Open for more actions"
+		aria-haspopup="true"
```
YonatanKra, line 58: This is tricky.
We assume it has a popup because that's our use case. Might it actually not have a popup (user didn't set it)?

Two points here:
1. Note that `menu` controls the `aria-haspopup` of its anchor. Maybe it should be delegated to the `popup` itself? What do you think?
2.  If we choose to leave this `true`, as this is the use case, I think we should mention this in the a11y section of the docs => `if you are using a split button, you must set the menu to its indicator`

What do you think?

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @param context - element definition context
+ * @public
+ */
+export const SplitButtonTemplate: (
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<SplitButton> = (context: ElementDefinitionContext) => {
+	return html<SplitButton>`
+	<template>
+		<div class="base">
```
rachelbt, line 80: I wonder if we need to add this: `role="group"`

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+
+	&:not(:disabled) {
+		cursor: pointer;
+	}
+
+	&:disabled {
+		cursor: not-allowed;
+	}
+
+	&.icon-only {
```
rachelbt, line 75: this is not working - no class like this is added in the template

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+			border-radius: 6px 0 0 6px;
+		}
+
+		&.shape-pill {
+			border-radius: 24px 0 0 24px;
+		}
+	}
+
+	&.indicator {
+		border-left: 0;
```
rachelbt, line 105: why is this needed?

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		&.shape-pill {
+			border-radius: 0 24px 24px 0;
+		}
+	}
+
+	/* Size */
+	&.size-super-condensed {
+		@include mixins.get-size-properties(
```
rachelbt, line 117: control need all this, but indicator does not.

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+			#{size.$vvd-size-super-condensed},
+			#{constants.$vvd-typography-base-condensed-bold},
+			4px,
+			8px
+		);
+		&.indicator {
+			width: 22px;
```
rachelbt, line 124: try using the size property and not fixed px.

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function indicatorButton(context: ElementDefinitionContext) {
+	const affixIconTemplate = affixIconTemplateFactory(context);
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<SplitButton>`
+	<button
+		${ref('_indicator')}
+		class="indicator ${getClasses}"
+		?autofocus="${(x) => x.autofocus}"
```
YonatanKra, line 55: I already mentioned this above for testing, but now I see that we have two elements with `autofocus`. It's kind of meaningless, because only the first one is taken into account according to the spec. Is there a reason why this `autofocus` is here?

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+);
+
+function actionButton(context: ElementDefinitionContext) {
+	const affixIconTemplate = affixIconTemplateFactory(context);
+	const focusTemplate = focusTemplateFactory(context);
+
+	return html<SplitButton>`
+	<button
+		${ref('_action')}
+		class="control ${getClasses}"
```
rachelbt, line 26: it feels like its more clean to move this to base - then you have the class only once.

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+		aria-atomic="${(x) => x.ariaAtomic}"
+		aria-busy="${(x) => x.ariaBusy}"
+		aria-current="${(x) => x.ariaCurrent}"
+		aria-disabled="${(x) => x.disabled || x.disabledAction}"
+		aria-expanded="${(x) => x.ariaExpanded}"
+		aria-hidden="${(x) => x.ariaHidden}"
+		aria-invalid="${(x) => x.ariaInvalid}"
+		aria-label="${(x) => x.ariaLabel}"
+		aria-live="${(x) => x.ariaLive}"
+		aria-pressed="${(x) => x.ariaPressed}"
```
YonatanKra, line 38: A few questions here:
1. What's the meaning of `aria-invalid` in the case of `split-button`?
2. I think we should add an a11y section in the docs according to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live). This is important as the value might change according to the value in the menu(?).
3. These values are not tested and are not even mentioned in the `basic` section. Can you add a test for them as well?
4. Do we have a `pressed` state on this button?

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -0,0 +1,85 @@
...
+		${ref('_indicator')}
+		class="indicator ${getClasses}"
+		?autofocus="${(x) => x.autofocus}"
+		?disabled="${(x) => (x.disabled || x.disabledIndicator)}"
+		title="Open for more actions"
+		aria-haspopup="true"
+		aria-expanded="${(x) => x.ariaExpanded}"
+		aria-disabled="${(x) => x.disabled || x.disabledIndicator}"
+		aria-label="${(x) => x.ariaLabel}"
+		aria-labelledby="${(x) => x.ariaLabelledBy}">
```
YonatanKra, line 62: Can you make sure these are tested, please?

libs/components/src/lib/split-button/split-button.ts
```diff
@@ -0,0 +1,171 @@
...
+	@attr({ attribute: 'split-indicator', mode: 'fromView' }) splitIndicator: string = 'chevron-down-line';
+
+	/**
+	 * Indicates if the split button's popup is open.
+	 *
+	 * @public
+	 * @remarks
+	 * HTML Attribute: open
+	 */
+	@attr({ mode: 'boolean' }) open = false;
```
rachelbt, line 136: is this needed?

YonatanKra (changes requested):
Great job! I like how it was done!

I've written 12 comments. Many of them are testing style suggestions I think make the code clearer and easier to read.

I've written a few that require a bit more thinking. Please let me know when you're on it, and we can discuss it online.

rachelbt (commented):
I didn't see that @YonatanKra was also reviewing the PR. I checked only the template + css.
the split button looks good, left some comments to make it better :)
some general comments:
1. the aria-label that is added - is implemented on both buttons - but it should be only on the more option. Having said that- we need to think tow to set an aria-label if there's only icon in the control button.
2. the outline divider is on top of the outer border
3. this one is for @AyalaBu - the half disable state is strange - especially the outline one.
4. is it possible to place the menu to the bottom-start of the indicator button? I think a use case for split-button and menu will be nice (and remove it from the first preview - so users will not expect a menu in the template)
5. really sorry - have some more css comments - but I want to wait with them after you finish what I left for now :)

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
+@use "../../../../../dist/libs/tokens/scss/tokens.constants" as constants;
+@use "partials/variables" as variables;
+@use "partials/mixins" as mixins;
+@use "../focus/partials/variables" as focus-variables;
+
+@use "../../../../shared/src/lib/sass/mixins/connotation/config" with (
+	$connotations: accent cta success alert,
```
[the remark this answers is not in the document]
AyalaBu: There was no half disabled design for a split button.
I don't think it should be allowed.
Also, I didn't see it even once.
If it's disabled, then it's wholly disabled.

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -0,0 +1,187 @@
...
+	$shades: primary primary-text primary-increment contrast fierce firm soft faint,
+	$default: accent
+);
+@use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
+
+@use "../../../../shared/src/lib/sass/mixins/appearance/config" as appearance-config with (
+	$appearances: filled outlined ghost,
+	$states: idle hover disabled active,
+	$default: ghost
```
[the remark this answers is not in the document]
AyalaBu: I believe it is filled, but it really can depend on the designer's decision.

libs/components/src/lib/split-button/split-button.scss
rinaok: why not?

libs/components/src/lib/split-button/split-button.template.ts
rinaok: are you talking about control?

libs/components/src/lib/split-button/split-button.template.ts
rinaok: @YonatanKra Basically the split button's structure should be like this:

`<div role="group" aria-label="Split Button Example">
  <button id="main-button">Main Action</button>
  <button aria-haspopup="true" aria-expanded="false" aria-controls="dropdown-menu">▼</button>
  <ul id="dropdown-menu" role="menu" aria-labelledby="main-button">
    <li role="menuitem"><a href="#">Option 1</a></li>
    <li role="menuitem"><a href="#">Option 2</a></li>
    <li role="menuitem"><a href="#">Option 3</a></li>
  </ul>
</div>`

That's why I wanted to add the menu with a slot inside to the template. If that was the case we could handle the expanded, I mentioned this issue in the doc.

1. aria-haspopup should be on menu not on the popup.
2. I can add it to the a11y section.

libs/components/src/lib/split-button/split-button.template.ts
rinaok: added

libs/components/src/lib/split-button/split-button.scss
rinaok: the two buttons will not get the appearance

libs/components/src/lib/split-button/split-button.template.ts
YonatanKra: 1. `has-popup` let's split the question:

- The attribute should be on the anchor, or should it not? 
- What I meant by delegating the `has-popup` to the popup was that currently menu is doing that. Other components that are using the `popup` don't. Is there a use case of using the popup where we don't want the anchor to have `has-popup`?

2. The `menu` auto adds the `has-popup` attribute to the anchor, so there's no need to hard-code it to `true` (IMO). This way, if someone adds a menu, the menu will add `has-popup`. If no menu was added, there's no popup... do you think this covers our use cases?

rachelbt (commented):
@rinaok @AyalaBu 
I changed the css, I think now it fits the design better.
@AyalaBu the indicator width of the condensed and expanded are not as in Figma since there's no values of size of 38px or 26px in our sizes.
Also - there was difference in the separator opacity. I gave all 0.35.

rinaok:
> 

The outlined hover doesn't look good now:

image

AyalaBu:
Looks great!

Aside of one issue:
Outline button hover state looks kind of off,
but only when you hover the more actions part.
Screenshot 2023-08-10 at 13 44 28

rachelbt:
> Looks great!
> 
> Aside of one issue: Outline button hover state looks kind of off, but only when you hover the more actions part. Screenshot 2023-08-10 at 13 44 28

@AyalaBu @rinaok 
I fixed the hover issue :)

libs/components/src/lib/split-button/README.md
```diff
@@ -0,0 +1,160 @@
...
+    <vwc-menu-item text="Menu item 1"></vwc-menu-item>
+    <vwc-menu-item text="Menu item 2"></vwc-menu-item>
+  </vwc-menu>
+</vwc-split-button>
+
+<script>
+  window.onload = function(){ 
+    menu.anchor = splitButton.indicator;
+    splitButton.action.onclick  = () => {
+      alert("clicked on action");
```
rachelbt, line 29: this is kind of nagging, and seems a bit unrelated to the documentation

rinaok: But it shows the users how to use it, maybe I can move it to use cases

rinaok: @rachelbt Updated

rachelbt (approved):

PR #1312, opened 2023-08-07
feat(data-grid): custom columns widths (VIV-1252)

YonatanKra:
See the test and documentation example.
In essence, setting `grid.gridTemplateColumns` to a valid value will result in custom column widths.


libs/components/src/lib/data-grid/data-grid-cell.scss
```diff
@@ -59,11 +59,11 @@
 slot {
 	display: block;
 	overflow: hidden;
+	inline-size:100%;
 	text-overflow: ellipsis;
-	white-space: nowrap;
+	white-space: nowrap
```
rachelbt, line 64:
```suggestion
	white-space: nowrap;
```

libs/components/src/lib/data-grid/README.md
```diff
@@ -176,6 +176,30 @@ See the [use cases](#sortable-columns) for more information.
 </script>
 ```
 
+### gridTemplateColumns
+
+- Type: `string`
+- Default: `undefined`
+
+The grid-template-columns property defines the line size of every grid column. Values match the CSS grid-template-columns property.
```
rachelbt, line 184: is this can be used not in the JS but as setting on the `<vwc-data-grid>`?
If yes - is it set on the `vwc-data-grid` or `vwc-data-grid-row`.
What is the property that is added? `grid-template-columns` or `gridTemplateColumns`?
This needs to be more clear in the docs

libs/components/src/lib/data-grid/README.md
```diff
@@ -176,6 +176,30 @@ See the [use cases](#sortable-columns) for more information.
...
+	}
+</style>
+<vwc-data-grid></vwc-data-grid>
+<script>
+	grid = document.querySelector('vwc-data-grid');
+	grid.rowsData = [
+		{data1: 'data11', data2: 'data12'},
+		{data1: 'data21', data2: 'data22'},
+	];
+	grid.gridTemplateColumns = '50px 200px';
```
rachelbt, line 199: this does not work properly because we have a min-width on the cell host for the ellipsis :) 
see:
image
the min-width of the cell need to be changed as well

libs/components/src/lib/data-grid/README.md
YonatanKra: This is added on the `vwc-data-grid` and added for every `vwc-data-grid-row`.
The property is `gridTemplateColumns` and the attribute is `grid-template-columns`.

libs/components/src/lib/data-grid/README.md
YonatanKra: Why do we need `min-width`? For the case the parent doesn't have a width to spread the `fr`'s?

rachelbt: exactly - without it if there's a long text - the text will overflow the next cell

libs/components/src/lib/data-grid/README.md
YonatanKra: Should be able to, yes.

rachelbt (approved):
LGTM
We need to think about a better and clearer data-grid doc page. but this is a task by itself

PR #1320, opened 2023-08-09
feat(menu): add non-bubbling open and close events (VIV-1258)

RichardHelm:


libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -395,6 +395,32 @@ describe('vwc-menu', () => {
 		});
 	});
 
+	describe('open event', () => {
```
YonatanKra, line 398: Can you also test it doesn't bubble? I can only tell this from the title of the PR :)
It's actually the same test only you make sure the event doesn't fire in a parent element.

YonatanKra (changes requested):

libs/components/src/lib/menu/menu.spec.ts
RichardHelm: Sure done 👍 
Just checking it on the event

PR #1321, opened 2023-08-10
feat(nav-disclosure): add aria-current (VIV-893)

rinaok:


rachelbt (commented):
didn't go over all the files - why are this an option and not added by default to the nav-disclosure template?

rinaok:
> 

@rachelbt Exactly like in the nav-items, it is also an option and not by default

libs/components/src/lib/nav-disclosure/nav-disclosure.scss
```diff
@@ -66,10 +66,6 @@
 .icon {
 	font-size: 20px;
 	line-height: 1;
-
```
rachelbt, line 69: @AyalaBu id the icon suppose to be lighter then the text?
It isn't in the nav item, but it is in the nav-disclosure. 
@rinaok if this is needed then you can add:
`.control:not(.icon-only, [aria-current="true"]) &`

libs/components/src/lib/nav-disclosure/README.md
```diff
@@ -71,6 +71,22 @@ If label is not applied.
 <vwc-nav>
 ```
 
+### Aria Current
+
+- Type: `boolean`
```
rachelbt, line 76: aria current is not boolean - it can get other options then `true` and `false`. (like - `page` in the nav-item).
maybe its better to ask consumers to add only the property - `aria-current`, which will be false by default - as you wrote and when its closed we will change it to true.
WDYT?

libs/components/src/lib/nav-item/README.md
```diff
@@ -49,18 +49,8 @@ Within a set of pagination links, set a nav item `aria-current` value to *page*
 
 ```html preview
 <vwc-nav>
-  <vwc-nav-item href="#" text="Account" onclick="onClick(this)"></vwc-nav-item>
```
rachelbt, line 52: I liked this example - maybe we can add this as a usage example?

rachelbt (commented):
all looks OK.
Added small comments and one conceptual question about setting the aria-current

libs/components/src/lib/nav-disclosure/nav-disclosure.scss
AyalaBu: I believe we discussed it
Since the whole component is clickable the icon is clickable as well
Therefore it's supposed to be 1000 Neutral

rinaok: @AyalaBu @rachelbt What is `Neutral 1000`? We don't have it..

libs/components/src/lib/nav-disclosure/README.md
rinaok: @rachelbt In nav disclosure it'n only `true` or `false`. It is not an anchor, so it can't be `page`.

libs/components/src/lib/nav-disclosure/nav-disclosure.scss
AyalaBu: It's black #000000

libs/components/src/lib/nav-item/README.md
rinaok: i@rachelbt We have this example in nav. I think that the nav-item and nav-disclosure should be as minimal as possible .

libs/components/src/lib/nav-disclosure/README.md
rachelbt: but it is not defined as boolean - it can also get page if someone sets it there

rachelbt: I'm not sure.
If it is boolean then lets set it as one in `nav-disclosure.ts`, currently its set as a string:
`@attr({attribute: 'aria-current'}) override ariaCurrent: string | null = null;`
if this is not possible - we need to think about the API - how will we name it correctly

rinaok: I can't set it to boolean so I updated it to be string in the docs.

rachelbt (approved):
LGTM

PR #1324, opened 2023-08-10
feat(date-picker): add date-picker component (VIV-1244)

RichardHelm:
This adds a first version of the single date picker.
- `min` and `max` features not implemented, can be easily added later: [VIV-1273](https://jira.vonage.com/browse/VIV-1273)
- Does not work properly inside a form. This is because the text-field will have a different value than the date picker. Maybe we can use [ElementInternals API](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) to add this functionality later. Ticket: [VIV-1272](https://jira.vonage.com/browse/VIV-1272)


libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+}
+
+.months-separator {
+	background-color: var(#{constants.$vvd-color-neutral-200});
+	block-size: 1px;
+	margin-inline: -12px;
+}
+
+.month-grid {
+	display: flex;
```
rachelbt, line 160: why not use `display: grid `- and then you can also remove the `months-row` wrapper div.

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	.title {
+		display: flex;
+		flex-grow: 1;
+		justify-content: center;
+	}
+
+	.title-action {
+		position: relative;
```
rachelbt, line 35: why is this needed?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		padding: 0;
+		border: none;
+		background: none;
+		block-size: 24px;
+		border-radius: 6px;
+		cursor: pointer;
+		font: var(#{constants.$vvd-typography-base-extended});
```
rachelbt, line 42: please add color - in all the buttons - otherwise it gets button default of browser or user agent (in my safari  iphone it get s blue)

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+			display: none;
+		}
+
+		&:hover {
+			background: var(#{constants.$vvd-color-neutral-50});
+		}
+
+		&:active {
+			background: var(#{constants.$vvd-color-neutral-100});
+		}
```
rachelbt, line 59: why not use appearance ghost + connotation?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	.footer {
+		display: flex;
+		align-items: center;
+		justify-content: flex-end;
+		block-size: 48px;
```
rachelbt, line 66: the buttons sets the block-side - why is this needed?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		gap: 8px;
+	}
+}
+
+.calendar {
+	display: flex;
```
rachelbt, line 72: you can use `display: grid` and remove all un-needed row wrappers and padding or margin on the calendar children

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+	flex-direction: column;
+	margin-top: 7px;
```
rachelbt, line 74: we try to avoid odd numbers - 8px is better :)

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+	gap: 4px;
+
+	.calendar-separator {
```
rachelbt, line 77: again - cant this be done with `::before`?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		justify-content: center;
+		block-size: 16px;
+		font: var(#{constants.$vvd-typography-base-condensed});
+		inline-size: 40px;
+	}
+
+	.calendar-week {
+		display: flex;
+		justify-content: center;
+	}
```
rachelbt, line 100: this will be un-needed if you use `grid`

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+
+	.calendar-day {
+		position: relative;
```
rachelbt, line 103:
```suggestion

```
what is this needed for?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		--focus-inset: -2px;
+
+		&:hover {
+			background: var(#{constants.$vvd-color-neutral-50});
+		}
+
+		&:active {
+			background: var(#{constants.$vvd-color-neutral-100});
+		}
+
```
rachelbt, line 138: why not use appearance states?

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		display: inline-flex;
+		align-items: center;
+		justify-content: center;
+		padding: 0;
+		border: none;
+		background: none;
+		block-size: 28px;
+		border-radius: 50%;
+		cursor: pointer;
+		font: var(#{constants.$vvd-typography-base});
```
rachelbt, line 113: also - add color :)

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+		align-items: center;
+		justify-content: center;
+		padding: 0;
+		border: none;
+		background: none;
+		block-size: 48px;
+		border-radius: 50%;
+		cursor: pointer;
+		font: var(#{constants.$vvd-typography-base-extended});
+		inline-size: 48px;
```
rachelbt, line 182: you can use `aspect-ratio`
https://caniuse.com/mdn-css_properties_aspect-ratio

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,212 @@
...
+			background: var(#{constants.$vvd-color-neutral-100});
+		}
+
+		&.current {
+			border: 1px solid var(#{constants.$vvd-color-neutral-200});
+		}
+
+		&.selected {
+			background: var(#{constants.$vvd-color-neutral-200});
+		}
```
rachelbt, line 210: again - cant we use appearance?

rachelbt (commented):
WOW - this is one of a big component!! 👏 👏 
Went over css - a little on the HTML structure in the template.
In general - it looks very good - left some comments regarding colors, appearance and using grid over flex.

libs/components/src/lib/date-picker/date-picker.template.ts
```diff
@@ -0,0 +1,287 @@
...
+			<div class="dialog" role="dialog" ${ref(
+		'dialogEl'
+	)} aria-modal="true" aria-label="${(x) =>
+	x.locale.datePicker.chooseDateLabel}">
+				${renderDialogHeader(context)}
+				${when(
+		(x) => x.inMonthPicker,
+		html<DatePicker>`
+						<div class="months-separator"></div>
+						${renderMonthPickerGrid(context)}
```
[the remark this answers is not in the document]
RichardHelm: I tried this but it didn't work well with the layout, so changed it to `vwc-divider` with `role="presentation"` now

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: It seems the aria `grid` requires the use of a row element with `role="row"`
Therefore leaving this

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Sure, added 👍

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Good point, I added it now

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: I changed it to a padding now 👍

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Sure 👍

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Same here

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Same here

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Same as the other buttons

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Done

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Done

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Sure, done 👍

libs/components/src/lib/date-picker/date-picker.scss
RichardHelm: Yes, done

YonatanKra:
When I select a date, I'm focused inside the input. Clicking the input again does not open the calendar to pick another date. I need to blur and refocus in order to get the date picker again.

YonatanKra (changes requested):
Wow! That's a big one!
Nice work! A few comments and discussions. Less than I'd expect for such a big PR :)

libs/components/src/lib/components.ts
```diff
@@ -1,53 +1,54 @@
...
 export * from './button/definition';
-export * from './calendar/definition';
 export * from './calendar-event/definition';
+export * from './calendar/definition';
 export * from './card/definition';
 export * from './checkbox/definition';
 export * from './combobox/definition';
-export * from './empty-state/definition';
+export * from './data-grid/definition';
+export * from './date-picker';
```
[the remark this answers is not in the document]
RichardHelm: Yes, fixed 👍

libs/components/src/lib/date-picker/calendar/calendarGrid.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+} as DatePickerLocale;
+
+const enGbLocale = { ...enUsLocale, firstDayOfWeek: 1 };
+
+describe('buildCalendarGrid', () => {
+	it('should include a cell for dates within the month', () => {
+		const grid = buildCalendarGrid({ month: 7, year: 2023 }, enUsLocale);
+		expect(grid.grid[0][2].date).toBe('2023-08-01');
+		expect(grid.grid[0][2].label).toBe('1');
+		expect(grid.grid[0][2].isOutsideMonth).toBe(false);
```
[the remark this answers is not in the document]
RichardHelm: Sure, done

libs/components/src/lib/date-picker/calendar/calendarGrid.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+	});
+
+	it('should include a cell for dates in the week after the last of the month', () => {
+		const grid = buildCalendarGrid({ month: 7, year: 2023 }, enUsLocale);
+		expect(grid.grid[4][6].date).toBe('2023-09-02');
+		expect(grid.grid[4][6].label).toBe('2');
+		expect(grid.grid[4][6].isOutsideMonth).toBe(true);
+	});
+
+	describe('for August 2023 with en-US locale', () => {
```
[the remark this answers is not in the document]
RichardHelm: It's just an example, changed it to `for the example month August 2023`

libs/components/src/lib/date-picker/calendar/calendarGrid.ts
```diff
@@ -0,0 +1,98 @@
...
+	label: string;
+	isOutsideMonth: boolean;
+}
+
+export type CalendarGrid = {
+	weekdays: Weekday[];
+	grid: CalendarGridDate[][];
+};
+
+export const buildCalendarGrid = (
```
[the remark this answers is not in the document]
RichardHelm: Done 👍

libs/components/src/lib/date-picker/date-picker.spec.ts
```diff
@@ -0,0 +1,550 @@
...
+		)) as DatePicker;
+		textField = element.shadowRoot!.querySelector('.control') as TextField;
+		popup = element.shadowRoot!.querySelector('.popup') as Popup;
+		titleAction = element.shadowRoot!.querySelector(
+			'.title-action'
+		) as HTMLButtonElement;
+
+		// In jsdom does not currently support delegatesFocus, see https://github.com/jsdom/jsdom/issues/3418
+		// Workaround to allow us to test focus behaviour:
+		// Unfortunately it causes jest to throw "TypeError: Converting circular structure to JSON" when a test fails
```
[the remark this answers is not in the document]
RichardHelm: Yes, when a test fails jest will throw that error if the workaround is in place. I don't know why

In our case we're focussing an element is inside the shadow dom, so `shadowRoot.activeElement` should be set to the element. But JSDOM does not do that, which breaks the code and we cannot test it

libs/components/src/lib/date-picker/date-picker.spec.ts
```diff
@@ -0,0 +1,550 @@
...
+
+		it('should switch to the new month when navigating to a date of a different month', async () => {
+			getDateButton('2023-08-01').focus();
+			pressKey('ArrowLeft');
+			await elementUpdated(element);
+
+			expect(getDialogTitle()).toBe('July 2023');
+		});
+
+		it('should not prevent default when pressing any other key', async () => {
```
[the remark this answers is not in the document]
RichardHelm: Sure done 👍

libs/components/src/lib/date-picker/date-picker.spec.ts
```diff
@@ -0,0 +1,550 @@
...
+			await elementUpdated(element);
+
+			expect(getDialogTitle()).toBe('January 2023');
+		});
+
+		it('should close the month picker when clicking on the title action', async () => {
+			titleAction.click();
+			await elementUpdated(element);
+
+			expect(getDialogTitle()).toBe('August 2023');
```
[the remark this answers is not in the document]
RichardHelm: Ok, updated with `expect(element.shadowRoot!.querySelector('.month-grid')).toBeNull();`

libs/components/src/lib/date-picker/date-picker.ts
```diff
@@ -0,0 +1,584 @@
...
+			// Ensure we are switched to the month of the new selected date
+			this.selectedMonth = monthOfDate(validatedValue);
+		} else {
+			this.presentationValue = '';
+		}
+	}
+
+	#updateValueDueToUserInteraction(newValue: DateStr | undefined) {
+		this.value = newValue;
+		this.$emit('change');
```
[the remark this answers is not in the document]
RichardHelm: No, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) it should be fired here:
> Depending on the kind of element being changed and the way the user interacts with the element, the change event fires at a different moment:
...
> When the user commits the change explicitly (e.g., by selecting a value from a <select>'s dropdown with a mouse click, by selecting a date from a date picker for <input type="date">

libs/components/src/lib/date-picker/date-picker.ts
```diff
@@ -0,0 +1,584 @@
...
+			return this.value;
+		}
+		return null;
+	}
+
+	/**
+	 * The month the calendar is currently showing.
+	 * @internal
+	 */
+	@observable selectedMonth = getCurrentMonth();
```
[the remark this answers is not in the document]
RichardHelm: I have considered this, but I think it would be overkill and make testing etc. harder

Instead I've now prefixed all the internal members with `_`, WDYT?

libs/components/src/lib/date-picker/date-picker.ts
```diff
@@ -0,0 +1,584 @@
...
+			this.textFieldEl.focus();
+			this.#isClosingPopup = false;
+		}
+	}
+
+	/**
+	 * On keydown anywhere in the date picker.
+	 * @internal
+	 */
+	onBaseKeyDown(event: KeyboardEvent) {
```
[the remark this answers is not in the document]
RichardHelm: See other comment

libs/components/src/lib/date-picker/date-picker.ts
```diff
@@ -0,0 +1,584 @@
...
+			this.monthPickerYear = null;
+		} else {
+			this.monthPickerYear = this.selectedMonth.year;
+		}
+	}
+
+	/**
+	 * @internal
+	 */
+	onPrevYearClick() {
```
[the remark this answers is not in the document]
RichardHelm: See other comment

libs/components/src/lib/date-picker/ui.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import { expect, test } from '@playwright/test';
+import {
+	loadComponents,
+	loadTemplate,
+	useFakeTime,
+} from '../../visual-tests/visual-tests-utils.js';
+
+const components = ['date-picker'];
+
+test('should show the component', async ({ page }: { page: Page }) => {
```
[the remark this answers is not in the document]
RichardHelm: Sure, I've added an e2e test now for basic select functionality

RichardHelm:
@YonatanKra , please have another look.

I've changed the functionality to open the popup again when clicking on the text field, or, for keyboard users, when pressing enter

libs/components/src/lib/date-picker/date-picker.scss
```diff
@@ -0,0 +1,163 @@
...
+
+	position: relative;
+	padding: 0;
+	border: none;
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	color: var(#{appearance.get-appearance-token(text)});
+	cursor: pointer;
+
+	.focus-indicator {
+		color: var(#{constants.$vvd-color-canvas-text});
```
rachelbt, line 149:
```suggestion
		color: var(#{constants.$vvd-color-canvas-text});
		#{focus-variables.$focus-stroke-gap-color}: transparent;
```
this is to make the focus outline lighter.
You may also need to add:
`@use "../focus/partials/variables" as focus-variables;
`

rachelbt (commented):
CSS - LGTM
I wonder if we need to consider adding a button to open the date-picker (like in t[his example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/)) because in mobile its a bit annoying that pressing the text-field open both the keyboard and the the popup.
@AyalaBu WDYT? maybe only in mobile we need to do this change.
This can be in a separate PT I think :)

libs/components/src/lib/date-picker/date-picker.ts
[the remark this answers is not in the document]
YonatanKra: Still unsure if this is the right way to go exposing all that stuff. 
I also think it's a bigger discussion and shouldn't hold this PR :)

YonatanKra (approved):
Nice work! Thanks!
Can you just connect the form association ticket from JIRA to this PR for later reference please?

RichardHelm:
Sure, it's [VIV-1272](https://jira.vonage.com/browse/VIV-1272)

libs/components/src/lib/date-picker/date-picker.scss
AyalaBu: MUI updated their datepicker and actually changed it a bit and added a button to open a datepicker only.
I really like it and I think that we should consider this behavior.
I think it is a better UX.

https://mui.com/x/react-date-pickers/date-picker/#basic-usage

WDYT?

PR #1328, opened 2023-08-14
feat(pagination): add shape (VIV-1062)

rinaok:


libs/components/src/lib/pagination/pagination.scss
```diff
@@ -32,3 +31,12 @@
...
 }
+
+/* Shape */
+
+.control:not(.shape-pill) {
+	border-radius: 4px;
+}
+.control.shape-pill {
+	border-radius: 14px;
+}
```
rachelbt, line 42: you can move this to be inside the control and use:

&:not(.shape-pill) {
	border-radius: 4px;
}
&.shape-pill {
	border-radius: 14px;
}

libs/components/src/lib/pagination/pagination.spec.ts
```diff
@@ -1,8 +1,10 @@
...
-import {Size} from '@vonage/vivid';
-import type {Button} from '../button/button';
-import {Pagination, PaginationSize} from './pagination';
+import { elementUpdated, fixture, getControlElement } from '@vivid-nx/shared';
+import { FoundationElementRegistry } from '@microsoft/fast-foundation';
+import type { Button } from '../button/button';
+import { Shape, Size } from '../enums';
+import { Pagination, PaginationShape, PaginationSize } from './pagination';
 import '.';
+import { paginationDefinition } from './definition';
```
rachelbt, line 7: I think we need to add the shape export in the definition as well

rachelbt (commented):
LGTM
@AyalaBu if pill is set - are the next + previous button should also be pill?

libs/components/src/lib/pagination/pagination.spec.ts
rinaok: @rachelbt added export to definition

rachelbt (approved):
LGTM

libs/components/src/lib/pagination/pagination.spec.ts
AyalaBu: I believe so.
Putting together two different shapes in the pagination component
can create some real ugliness

PR #1331, opened 2023-08-15
fix(pagination): outline  use-case (VIV-1063)

rinaok:


rachelbt (commented):
as we talked - the outline is not an appearance since it doesn't effect the buttons same as shape or size.
We will add an outline to use case. adding the code:
```
<style>
vwc-pagination {
    border: 1px solid var(--vvd-color-neutral-400);
    padding: 6px;
    border-radius: 24px;
    display: inline-block;
}
</style>

<vwc-pagination total="10" shape="pill"></vwc-pagination>
```
In addition please fix this in the doc page:

- Nav Icons - wrong attribute name and poor description
- Pages List -> Properties and add a table

rachelbt (approved):
Look so much better then adding the appearance!

PR #1332, opened 2023-08-16
fix(typography): fix font-setting typography css (VIV-1262)

rachelbt:
the fix was initially only adding documentation however it let to:
- fixing wrong setting of font for sub sup text
- adding class .base for use to text (equivalent to p element)
- adding option for removing margin-block with tight class

the examples of the typography are here:
https://vivid-hzbyy08etfm0.deno.dev/designs/typography


docs/getting-started/advanced.md
```diff
@@ -32,6 +32,10 @@ These **core styles** rely on the [tokens and fonts to be loaded](/getting-start
 
 Styles **require** a `vvd-root` class selector to be present on a wrapping element. [Follow the instruction to set a vvd-root class](/getting-started/quick-start/#setting-vivid-class).
 
+### Using the typography
+
+[see all usage examples in the Typography](/designs/typography/#Using-Vivid-Typography).
```
RichardHelm, line 37:
```suggestion
[See typography usage examples](/designs/typography/#Using-Vivid-Typography).
```

libs/components/package-lock.json
RichardHelm, line 1: Why was this file added? Can we remove it?

docs/design/typography.md
```diff
@@ -90,4 +88,67 @@ The following CSS custom properties can be referred to style the text in your ap
...
-```
\ No newline at end of file
+```
+
+
+
+## Using Vivid Typography
+The typography style target both HTML tags and css classes.  
+All `headlines` and `p` elements also get `margin-block`.  
+Use the class `tight` to remove `margin-block`.
```
RichardHelm, line 98: I think we should add here that the css needs to be included for this
```suggestion
## Using Vivid Typography

The Vivid typography css is optional. To use it, you will need to [include the typography styles](/getting-started/advanced).

The typography style target both HTML tags and css classes.  
All `headlines` and `p` elements also get `margin-block`.  
Use the class `tight` to remove `margin-block`.  
```

RichardHelm (changes requested):

libs/components/package-lock.json
rachelbt: by mistake - I'll revert it

RichardHelm (approved):

PR #1333, opened 2023-08-16
feat(select): add placeholder (VIV-1242)

rinaok:


libs/components/src/lib/select/select.spec.ts
```diff
@@ -537,4 +538,34 @@ describe('vwc-select', () => {
...
+			element = (await fixture(
+				`<${COMPONENT_TAG}>
+					<option value="1">1</option>
+					<option value="2" selected>2</option>
+					<option value="3">3</option>
+				</${COMPONENT_TAG}>`
+			)) as Select;
+
+			element.placeholder = 'placeholder';
+			await elementUpdated(element);
```
RichardHelm, line 566: Why does this test not have the placeholder here?

It doesn't seem to work when setting both the placeholder and `selected` / `value`

```suggestion
		it('should change selectedindex -1 when option selected', async () => {
			element = (await fixture(
				`<${COMPONENT_TAG} placeholder="placeholder">
					<option value="1">1</option>
					<option value="2" selected>2</option>
					<option value="3">3</option>
				</${COMPONENT_TAG}>`
			)) as Select;

			await elementUpdated(element);
```

libs/components/src/lib/select/select.spec.ts
```diff
@@ -537,4 +538,34 @@ describe('vwc-select', () => {
 			expect(element.selectedOptions[0]).toEqual(element.querySelector('option:nth-child(2)'));
 		});
 	});
+
+	describe('placeholder', function () {
```
RichardHelm, line 542: Can we add some more tests here please?
Maybe that the placeholder is displayed and that, like selectedIndex, value will be as expected

libs/components/src/lib/select/select.spec.ts
```diff
@@ -537,4 +538,34 @@ describe('vwc-select', () => {
...
+		it('should set selectedindex -1 when placeholder', async () => {
```
RichardHelm, line 543:
```suggestion
		it('should set selectedindex -1 when placeholder is set', async () => {
```

libs/components/src/lib/select/select.ts
```diff
@@ -28,20 +28,55 @@ export type SelectShape = Extract<Shape, Shape.Rounded | Shape.Pill>;
...
+	 * The placeholder attribute.
 	 *
-	 * Slot observer:
+	 * @public
+	 * HTML Attribute: placeholder
+	 */
+	@attr placeholder: string | undefined;
+
+	/**
+	 * The ref to the internal `.control` element.
```
RichardHelm, line 67:
```suggestion
	 * The ref to the internal placeholder option.
```

RichardHelm (changes requested):

RichardHelm (commented):
Looks good now!
Just the screenshots seem to be failing now

rinaok:
> Looks good now! Just the screenshots seem to be failing now

@RichardHelm They didn't pass because of the date-picker.

RichardHelm (approved):

PR #1334, opened 2023-08-16
feat(tokens): add compatibility css to fix token collision with vivid-2 (VIV-1265)

RichardHelm:


docs/getting-started/migration.md
```diff
@@ -81,60 +81,56 @@ After:
...
+</html>
 ```
 
 It is recommended to add the class to the [:root](https://developer.mozilla.org/en-US/docs/Web/CSS/:root) element, this is the easiest way to make your app benefit from tokens and fonts.
 
-Alternatively, you can add it to any other wrapping element in order to restrict the styles's scope to specific parts of your application, but please note that this may negatively impact font scaling.
+Alternatively, you can add it to any other wrapping element in order to restrict the style's scope to specific parts of your application, but please note that this may negatively impact font scaling.
 
-<vwc-note connotation="warning" icon="warning-solid" headline="Avoid Tokens Collisions">As long as you use vivid-2 components in the project it is best to add the vvd-root class to each vivid-3 component, and not to the html or body.</vwc-note>
+Wherever you choose to add the class, you will need to add vivid-2's `vvd-scheme-main` to a descendant element. If you are using the `:root` element, you can add it to the body.
```
rachelbt, line 120: since this is really important maybe we should add this to a note as it was, or we can ad sub-heading the the 4th section something like:

```
## Step 4: Setting Vivid Class
### Vivid-3 class
### Vivid-2 class
```
not sure about the text :)

rachelbt (commented):
Looks great, and I guess you tested it!
I wonder if we need to test it also with other customers.
We can ask for volunteers in ask-vivid, WDYT?

rachelbt (approved):
LGTM!
Now we need more people trying this :)

PR #1340, opened 2023-08-21
feat(checkbox): add default slot (VIV-1088)

rinaok:


libs/components/src/lib/checkbox/README.md
```diff
@@ -139,3 +139,42 @@ Use the `value` member to set the checkbox's value.
...
+  }
+  a:hover:visited {
+    color: var(--vvd-color-information-700);
+  }
+  a:active {
+    color: var(--vvd-color-cta-800);
+  }
+  a:active:visited {
+    color: var(--vvd-color-information-800);
+  }
```
rachelbt, line 166: maybe we can add style for slotted - like general color - so that will align to the label color and with dark-light theme. And for a tag.
I'm wondering if this is the right way to do it for general slotted items:
```
::slotted(*) {
}
```
and this is for the a:
```
::slotted(a),
::slotted(a:visited) {
	color: var(#{constants.$vvd-color-cta-600});
}
::slotted(a:hover),
::slotted(a:visited:hover) {
	color: var(#{constants.$vvd-color-cta-700});
}

```

libs/components/src/lib/checkbox/README.md
```diff
@@ -139,3 +139,42 @@ Use the `value` member to set the checkbox's value.
...
+  <a href="#" onclick="openLink(event)">Vonage Terms of Service</a>
+</vwc-checkbox>
+
+<script>
+  function openLink(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    window.open('https://www.vonage.com/legal/', '_blank');
+  }
+</script>
```
rachelbt, line 179: this is not needed...

libs/components/src/lib/checkbox/README.md
```diff
@@ -139,3 +139,42 @@ Use the `value` member to set the checkbox's value.
...
+  a:active {
+    color: var(--vvd-color-cta-800);
+  }
+  a:active:visited {
+    color: var(--vvd-color-information-800);
+  }
+</style>
+
+<vwc-checkbox label="I agree to" error-text="You need to accept the Terms of service">
+  <a href="#" onclick="openLink(event)">Vonage Terms of Service</a>
```
rachelbt, line 170:
```suggestion
  <a href="https://www.vonage.com/legal" target="_blank">Vonage Terms of Service</a>
```
a tag can't have on-click:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events

I understand the problem - I'm wondering if we need to deal with it or consumers.
It feels like we need this.
I guess the big question is what the slot is for. 
Does it replace the label? 
Does it added an optional link to the label?

rachelbt (commented):
wondering about how this need to be implemented - regarding the functionality of checking the checkbox

libs/components/src/lib/checkbox/README.md
rinaok: @rachelbt If we will remove it, when you click on the link the checkbox will be selected

libs/components/src/lib/checkbox/README.md
rinaok: I don't think that we should style the slot cause maybe the consumers don't need this specific style

libs/components/src/lib/checkbox/README.md
rachelbt: I know - see my comment here:
https://github.com/Vonage/vivid-3/pull/1340#discussion_r1300004584

libs/components/src/lib/checkbox/README.md
rinaok: anchor ca have an onclick event, you can find examples here:

https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR35

libs/components/src/lib/checkbox/README.md
rachelbt: yes, you are right.
maybe we should consider adding this as a class of `vvd-link` in the typography file.

rachelbt (commented):
we need to add `flex-wrap: wrap` to the control so we will not have this:
image

libs/components/src/lib/checkbox/README.md
rachelbt: OK.
That's still do not solve the issue.
if all slots are meant to contain links - will they need always to add this script?
Or its is not our concern at all?

rinaok: They will always add this script.

rinaok:
> we need to add `flex-wrap: wrap` to the control so we will not have this: image

Done

libs/components/src/lib/checkbox/README.md
rachelbt: a,
a:visited {
	color: var(#{constants.$vvd-color-cta-600});
}
a:hover,
a:visited:hover {
	color: var(#{constants.$vvd-color-cta-700});
}

rinaok: @rachelbt But `hover` and `visited` should be `information` and not `cta`. Also, what about `active`?

rachelbt: what?! visited is information? this is soooo strange - it's like the opposite of the native link :)

rachelbt (commented):
Seems better this way and easier for consumers.

if there's no label then there's an extra space at the right:
image

Also -there's  package-lock file

libs/components/src/lib/checkbox/checkbox.scss
```diff
@@ -48,12 +48,17 @@ $gutter: 2px;
...
+	.hide-label & {
+		display: none;
+	}
 }
 
 .control {
 	position: relative;
 	display: flex;
 	flex-shrink: 0;
+	flex-wrap: wrap;
```
rachelbt, line 61:
```suggestion

```
this is not needed any more but - you need to add this to the label:
```
display: flex;
    align-items: center;
    column-gap: 4px;
    flex-wrap: wrap;
```

libs/components/src/lib/checkbox/checkbox.spec.ts
```diff
@@ -50,64 +50,50 @@ describe('vwc-checkbox', () => {
...
-			const base = element.shadowRoot?.querySelector('.base.readonly');
-			expect(base).toBeInstanceOf(Element);
+			expect(element.shadowRoot?.querySelector('.readonly')).toBeTruthy();
 		});
 	});
 
 	describe('indeterminate', () => {
-		it('should set checked class when indeterminate is true', async () => {
-
+		it('should set readonly class when readonly is true', async () => {
```
rachelbt, line 79:
```suggestion
		it('should set checked class when readonly is true', async () => {
```

libs/components/src/lib/checkbox/checkbox.scss
```diff
@@ -48,12 +48,17 @@ $gutter: 2px;
 		cursor: not-allowed;
 		pointer-events: none;
 	}
+
+	.hide-label & {
+		display: none;
+	}
```
rachelbt, line 54: this need to be inside label :)

libs/components/src/lib/checkbox/checkbox.template.ts
```diff
@@ -26,46 +25,37 @@ const getClasses = ({
 		['checked', Boolean(checked) || Boolean(indeterminate)],
 		['disabled', Boolean(disabled)],
 		['error connotation-alert', Boolean(errorValidationMessage)],
-		['success connotation-success', !!successText]
+		['success connotation-success', !!successText],
+		['hide-label', !label && !slottedContent?.length],
```
rachelbt, line 29: this has no test

libs/components/src/lib/checkbox/checkbox.spec.ts
```diff
@@ -46,68 +46,69 @@ describe('vwc-checkbox', () => {
 			const labelEl = element.shadowRoot?.querySelector('label');
 			expect(labelEl?.textContent?.trim()).toEqual(label);
 		});
+
+		it('should not add hide-label class to .base if label', async function () {
```
rachelbt, line 50:
```suggestion
		it('should not add hide-label class to .base if there's no label', async function () {
```

rinaok: @rachelbt Your fix is not correct, the test checks if the hide-label class is not added when there is a label.

rachelbt (approved):

PR #1343, opened 2023-08-21
fix(tokens): fix vvd-theme-alternate class applied to vvd-component (VIV-1275)

RichardHelm:
This fixes an issue when both classes are applied like this:
```html
<vwc-menu class="vvd-component vvd-theme-alternate">...
```
Because both try to set the variables


rachelbt (approved):
LGTM

PR #1345, opened 2023-08-22
feat(listbox):  add shape and orientation (VIV-863)

rinaok:


rinaok:
> I realise this is some work started by Yinon, and while the code is fine, the component raises a lot of questions:
> 
> * The component does not exist in Figma. Should it be added then?
> * When used horizontally, isn't the multiple mode just the same as a select with multiple?
> * Maybe this whole component should just be a variant of select?
> * When used vertically, isn't it the same as [radio group example](https://vivid-5hxj38p4qb70.deno.dev/components/action-group#radio-group) of action-group?
> * If we add this input component, shouldn't it also have `label`, `value`, `helper-text` etc. and events, like other input elements?
> 
> Basically this PR adds a new component that hasn't actually be designed by anyone and has a lot of overlap with other components. So I'm not sure if this is really the right way to go.

@RichardHelm  I can remove the export and alpha tag from the PR and add only the horizontal changes until we decide what to do. Alternatively, we could just use it internally, like the popup, for example inside select - instead of a div class="listbox">

RichardHelm:
@rinaok Sure that sound good! Let's keep it internal for now

libs/components/src/lib/listbox/listbox.spec.ts
```diff
@@ -51,6 +53,44 @@ describe('vwc-listbox', () => {
...
+			expect(getBaseElement(element).classList.contains(`orientation-${orientation}`)).toBeTruthy();
+		});
+
+		it('should remove event listener when orientation is vertical', async function () {
+			const orientation = 'vertical';
+			element.orientation = orientation;
+			await elementUpdated(element);
+
+			expect(element.hasAttribute('orientation')).toBeTruthy();
+			expect(element.hasAttribute('keydown')).toBeFalsy();
```
RichardHelm, line 71: I don't think this works

rinaok: You are right, copied it from the original PR and haven’t noticed this test.

libs/components/src/lib/listbox/listbox.spec.ts
```diff
@@ -51,6 +53,53 @@ describe('vwc-listbox', () => {
...
+	describe('orientation', function () {
+		it('should set correct internal orientation style', async function () {
+			const orientation = 'horizontal';
+			(element as any).orientation = orientation;
+			await elementUpdated(element);
+
+			expect(getBaseElement(element).classList.contains(`orientation-${orientation}`)).toBeTruthy();
+		});
+
+		it('should not remove keydown event listener when orientation is vertical', async function () {
```
RichardHelm, line 65:
```suggestion
		it('should not remove keydown event listener when orientation is horizontal', async function () {
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1350, opened 2023-08-24
feat(data-grid): add filter button to columns (VIV-1073)

YonatanKra:
Things left to do:

- [x] Design
- [x] Documentation
- [x] a11y about aria-rowcount
- [x] a11y about `haspopup`
- [x] filterable with sortable - prevent sorting action when filter is clicked


libs/components/src/lib/data-grid/data-grid-cell.template.ts
```diff
@@ -26,13 +39,29 @@ function renderSortIcons<T extends DataGridCell>(c: ElementDefinitionContext) {
 		`;
 }
 
+function renderFilterIcons<T extends DataGridCell>(c: ElementDefinitionContext) {
+	const buttonTag = c.tagFor(Button);
+	return html<T>`
+			${when(shouldShowFilterIcons, html<T>`
+				<${buttonTag} @click="${handleFilterClick}" class="header-icon filter-button" icon="${getFilterIcon}"></${buttonTag}>
```
rachelbt, line 46:
```suggestion
				<${buttonTag} @click="${handleFilterClick}" class="header-icon filter-button" icon="${getFilterIcon}" size="condensed"></${buttonTag}>
```

rachelbt: you can add to` :host([cell-type="columnheader"]) .base` : `column-gap: 4px;`

rachelbt (commented):
went over only on the visual side :)
You also need to add this under `.base` in `data-grid-cess.scss`
```
	&:focus-visible {
		outline: none;
	}
```

libs/components/src/lib/data-grid/data-grid-cell.scss
```diff
@@ -35,6 +35,10 @@
 	color: var(#{appearance.get-appearance-token(text)});
 	font: var(#{constants.$vvd-typography-base});
 
+	&:focus-visible {
+		outline: none;
+	}
+
```
rachelbt, line 41:
```suggestion

```
I'm sorry - you need to add this after the line 23
```
:host(:focus-visible) {
	outline: none;
}
```

RichardHelm (commented):
I reviewed the code and it's fine.
However, I noticed that the ticket is still blocked by https://jira.vonage.com/browse/VIV-1284
I think this is because it's not yet clear what exactly should happen when the button is clicked.
The use case looks like this, but maybe it should be attached to the button instead?
image

rinaok:
> I reviewed the code and it's fine. However, I noticed that the ticket is still blocked by https://jira.vonage.com/browse/VIV-1284 I think this is because it's not yet clear what exactly should happen when the button is clicked. The use case looks like this, but maybe it should be attached to the button instead? image


@RichardHelm I'm not sure, do we need @AyalaBu for this?

YonatanKra:
Closing due to inconclusive design use case. @AyalaBu , let me know if this changes

PR #1351, opened 2023-08-24
feat(text-field): leading action-items slot (VIV-1239) 

rachelbt:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -155,18 +174,23 @@ $low-ink-color: --_low-ink-color;
 
 }
 
-.icon {
-	// to avoid autofill style hiding other elements
+.leading-items-wrapper {
 	position: absolute;
```
RichardHelm, line 178: I see no reason why leading and trailing action items need to be position absolute? Only the icon
These slots are really awkward to use otherwise, because you need to provide the exact width.

I tried laying them out normally and I see no problem:

https://github.com/Vonage/vivid-3/assets/86777660/e04bc8ef-b3d5-42ce-9f3a-aac82a758816

rachelbt: thank for this!
fixed it - not slots position: absolute!

libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -550,6 +550,21 @@ describe('vwc-text-field', () => {
...
+				querySelector('.base')?.classList.contains('has-icon');
+			expect(baseElementHasIconClass).toEqual(true);
+
+		});
+		it('should remove .no-leading class to .base if icon is set', async function () {
+			element.icon = 'home';
+			await elementUpdated(element);
+			const baseElementHasIcon = element.shadowRoot?.
+				querySelector('.base')?.classList.contains('no-leading');
+			expect(baseElementHasIcon).toEqual(false);
```
RichardHelm, line 566: Just updating the wording
```suggestion
		it('should remove .no-leading class from .base if icon is set', async function () {
			element.icon = 'home';
			await elementUpdated(element);
			const baseElementHasNoLeadingClass = element.shadowRoot?.
				querySelector('.base')?.classList.contains('no-leading');
			expect(baseElementHasNoLeadingClass).toEqual(false);
```

libs/components/src/lib/text-field/README.md
```diff
@@ -167,25 +167,56 @@ You can add action items elements using the `action-items` slot.
...
+</vwc-text-field>
+```
+
+
 ## CSS Variables
-### Inline end-Padding
-Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-tiems` slot.
+### Inline end-Padding - **Deprecated**
+This css variable is not in use anymore. Don't worry if it is set - the design still be the same :)
+Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-items` slot.
```
rachelbt, line 202:
```suggestion
~~Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-items` slot.~~
```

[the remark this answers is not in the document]
rachelbt: went for the strike-through.

libs/components/src/lib/text-field/README.md
```diff
@@ -167,25 +167,56 @@ You can add action items elements using the `action-items` slot.
...
+</vwc-text-field>
+```
+
+
 ## CSS Variables
-### Inline end-Padding
-Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-tiems` slot.
+### Inline end-Padding - **Deprecated**
+This css variable is not in use anymore. Don't worry if it is set - the design still be the same :)
+~~Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-items` slot.~~
```
rachelbt, line 202:
```suggestion
This css variable is not in use anymore. Don't worry if it is set - the design still be the same :). 
~~Use `--text-field-inline-end-padding` variable to set the text-field inline-end padding when using the `action-items` slot.~~
```

RichardHelm (approved):

RichardHelm (approved):

RichardHelm (changes requested):
I just realised we're missing the `@slot` annotations for both leading and normal action items.
Could you please add them as well?

RichardHelm (approved):

PR #1359, opened 2023-09-03
feat(menu-item): add submenu (VIV-1230)

rinaok:


RichardHelm:
Very cool! I didn't review the code but noticed some things when trying it out:

- When using keyboard you can open multiple submenus at once:
image
- When moving the mouse to click on an item in the submenu further down, the item loses hover so it closes / changes the submenu

https://github.com/Vonage/vivid-3/assets/86777660/9dfcf72d-52a9-46c6-8bae-0a33754d4e74

Maybe we need to look into something like https://smashingmagazine.com/2023/08/better-context-menus-safe-triangles

rinaok:
> Very cool! I didn't review the code but noticed some things when trying it out:
> 
> * When using keyboard you can open multiple submenus at once:
>   image
> * When moving the mouse to click on an item in the submenu further down, the item loses hover so it closes / changes the submenu
> 
>  Screen.Recording.2023-09-06.at.15.39.47.mov 
> Maybe we need to look into something like https://smashingmagazine.com/2023/08/better-context-menus-safe-triangles

@RichardHelm Fixed everything but the Safe Triangles. Can you check again? :) I know there are some teams waiting for this, maybe for the Safe Triangles we can open another issue.

libs/components/src/lib/menu-item/menu-item.spec.ts
```diff
@@ -267,26 +292,6 @@ describe('vwc-menu-item', () => {
...
-		it('should fire "change" event when enter is pressed', async () => {
-			const spy = jest.fn();
-			element.addEventListener('change', spy);
-			const event = new KeyboardEvent('keydown', {
-				key: keyEnter,
-			} as KeyboardEventInit);
-			element.dispatchEvent(event);
-			expect(spy).toHaveBeenCalled();
-		});
-
```
RichardHelm, line 289: Can we bring these tests back please? I think this should work again now that we're using fast's logic

libs/components/src/lib/menu-item/menu-item.spec.ts
```diff
@@ -324,4 +335,90 @@ describe('vwc-menu-item', () => {
...
+
+			await elementUpdated(menuElement);
+			expect(menuitem.expanded).toEqual(false);
+
+			menuitem.focus();
+			menuitem.dispatchEvent(new KeyboardEvent('keydown', { key: ' ' }));
+
+			await elementUpdated(menuElement);
+			expect(menuitem.expanded).toEqual(true);
+		});
```
RichardHelm, line 390: As this is basically testing the same thing with different keys, maybe we can write it like this instead?

```suggestion
		it.each(['Enter', ' '])('should expand first menuitem when "%s" is pressed', async (key) => {
			const menuitem = menuElement.querySelector('#menuitem') as MenuItem;

			menuitem.focus();
			menuitem.dispatchEvent(new KeyboardEvent('keydown', { key }));

			await elementUpdated(menuElement);
			expect(menuitem.expanded).toEqual(true);
		});
```

libs/components/src/lib/menu-item/menu-item.template.ts
```diff
@@ -12,43 +14,62 @@ const getCheckIcon = (affixIconTemplate: any, x: MenuItem, iconType: string) =>
...
+) => ViewTemplate<MenuItem> = (context: ElementDefinitionContext) => {
 	const affixIconTemplate = affixIconTemplateFactory(context);
 	const focusTemplate = focusTemplateFactory(context);
+	const iconTag = context.tagFor(Icon);
 
 	return html<MenuItem>`
 	<template
 		role="${x => (x.role ? x.role : MenuItemRole.menuitem)}"
+		aria-haspopup="${x => (x.hasSubmenu ? 'menu' : void 0)}"
 		aria-checked="${x => (x.role !== MenuItemRole.menuitem ? x.checked : void 0)}"
```
RichardHelm, line 73: Not sure who wrote `void 0` to begin with, but let's use undefined instead

```suggestion
		aria-haspopup="${x => (x.hasSubmenu ? 'menu' : undefined)}"
		aria-checked="${x => (x.role !== MenuItemRole.menuitem ? x.checked : undefined)}"
```

RichardHelm (changes requested):
@rinaok Thanks, it works great now! I just left some minor comments

libs/components/src/lib/menu-item/menu-item.template.ts
rinaok: @RichardHelm The void 0 is copied from FAST code

libs/components/src/lib/menu-item/menu-item.spec.ts
rinaok: Yes, I forgot to revert it

RichardHelm (approved):

PR #1369, opened 2023-09-07
fix(tabs): style (VIV-1170)

rinaok:


rinaok:
@AyalaBu Can you please review? :)

PR #1374, opened 2023-09-12
chore(docs): speed up build (VIV-1312)

RichardHelm:
This speeds up the 11ty build:
```diff
- [11ty] Copied 410 files / Wrote 69 files in 8.05 seconds (116.7ms each, v2.0.1)
+ [11ty] Copied 410 files / Wrote 69 files in 0.89 seconds (12.9ms each, v2.0.1)
```

Instead of transforming the built pages with JSDOM, code / variables previews are now handled by a markdown-it rule.

Also, instead of specifying all components used by an example, they are now determined from the example code. This means we no longer need to specify `modules` at all.


rinaok (approved):
WOW! That's amazing! What a great improvement :)

YonatanKra:
Great stuff!!! Thanks!

PR #1375, opened 2023-09-12
docs(code block): add banner (VIV-862)

rinaok:


apps/docs/assets/styles/main.css
```diff
@@ -253,6 +253,12 @@ article.article > h1 {
 	align-items: center;
 }
 
+.cbd-live-hint {
+  color: #abb2bf;
```
rachelbt, line 257: why not use vivid-variables?

apps/docs/assets/styles/main.css
```diff
@@ -253,6 +253,12 @@ article.article > h1 {
 	align-items: center;
 }
 
+.cbd-live-hint {
+  color: #abb2bf; 
+  background-color: #282c34;
```
rachelbt, line 258: is there a way we can change the hint + the code area to have vivid-background? 
I remember that @olaf-k said its not that easy, I see that the class name is strange: `.ͼo`

apps/docs/code-example-preview/createCodeExample.js
```diff
@@ -73,6 +73,7 @@ const renderiFrame = (
 			</div>
 			<details class="${CBD_DETAILS}" slot="main">
 				<summary></summary>
+				<div class="font-base tight cbd-live-hint"> // Enter HTML Source Code</div>
```
rachelbt, line 76: // Enter HTML Source Code - this is unrelated here,  why not: you can change the code or edit it - or that you wrote in the banner for the first place :)

rachelbt (commented):
this looks nice in dark mode!

apps/docs/assets/styles/main.css
rinaok: @rachelbt Cause I need the exact color of the background, it's not a vivid color.

apps/docs/assets/styles/main.css
rinaok: @rachelbt  I don't think we can

rachelbt (approved):
Look sooooo good!

PR #1383, opened 2023-09-18
feat(accordion-item and tree-itemadd icon slot to accordion-item

rinaok:


libs/components/src/lib/accordion-item/ui.test.ts-snapshots/snapshots-accordion-item-Desktop-Firefox-linux.png
RichardHelm, line 1: This screenshot is broken

package.json
```diff
@@ -27,6 +27,7 @@
     "dialog-polyfill": "^0.5.6",
     "dropzone": "^6.0.0-beta.2",
     "happy-dom": "^10.8.0",
+    "jsdom": "^22.1.0",
```
RichardHelm, line 30: Why did you add this?

libs/components/src/lib/accordion-item/accordion-item.spec.ts
```diff
@@ -54,6 +54,12 @@ describe('vwc-accordion-item', () => {
 	});
 
 	describe('icon', () => {
+		it('should have an icon slot', async () => {
+			element.icon = 'chat-solid';
+			await elementUpdated(element);
```
RichardHelm, line 59: I think this doesn't add to the test
```suggestion
```

RichardHelm (changes requested):

package.json
rinaok: @RichardHelm The ui tests didn't work and I had an error that jsdom doesn't exist. Only after I installed it they passed.

libs/components/src/lib/accordion-item/ui.test.ts-snapshots/snapshots-accordion-item-Desktop-Firefox-linux.png
rinaok: fixed

libs/components/src/lib/accordion-item/accordion-item.spec.ts
rinaok: removed

RichardHelm (approved):

YonatanKra (approved):
LGTM.
One thing to note - the `ui-tests` changed (probably because you changed the docs).
Can you check if they are also passing before the change to the docs but with the code change? To make sure everything's good?

PR #1384, opened 2023-09-18
fix(number-field): can now accept decimals and negative values from user

YonatanKra:


rinaok (dismissed):

libs/components/src/lib/number-field/number-field.spec.ts
```diff
@@ -464,6 +463,43 @@ describe('vwc-number-field', () => {
 			expect(activeClassWhenDisabled)
 				.toEqual(true);
 		});
+
+		it('should allow negative sign as first char', async function () {
```
RichardHelm, line 467: It doesn't seem to allow typing `-0` (e.g. `-0.5`)

libs/components/src/lib/number-field/number-field.spec.ts
```diff
@@ -464,6 +463,43 @@ describe('vwc-number-field', () => {
...
+
+			element.value = '5-';
+			await elementUpdated(element);
+			const valueWhenNegativeIsNotFirstChar = element.value;
+
+			expect(valueWhenNegativeIsNotFirstChar).toEqual('5');
+			expect(valueWhenNegativeIsFirstChar).toEqual('-');
+		});
+
+		it('should allow for decimal values', async function () {
```
RichardHelm, line 480: While decimal values can be entered now, they immediately cause an error to be displayed, even if step is set to `0.1`
image

RichardHelm (changes requested):

RichardHelm (dismissed):

PR #1386, opened 2023-09-19
feat(banner and badge): add icon slot (VIV-1302)

rinaok:


libs/components/src/lib/banner/banner.spec.ts
```diff
@@ -213,7 +213,11 @@ describe('vwc-banner', () => {
 		let getIcon: () => Icon;
 
 		beforeEach(function () {
-			getIcon = () => element.shadowRoot?.querySelector('.icon > vwc-icon') as Icon;
+			getIcon = () => element.shadowRoot?.querySelector('slot[name="icon"] > vwc-icon') as Icon;
+		});
```
YonatanKra, line 217: This is wrong... the `getIcon` shouldn't be initialized in a `beforeEach`. Like this:
``` suggestion
const getIcon = () => element.shadowRoot?.querySelector('slot[name="icon"] > vwc-icon') as Icon;
```

YonatanKra (changes requested):
Nice catch with the `nav-disclosure`, `option`, and `select` :)
One really small comment and we're done :)

P.S.
Also reminding you to run the visual tests locally without the e2e/readme changes to make sure the code didn't break the old look :)

PR #1388, opened 2023-09-20
chore(number-field): add missing docs for number field (VIV-1328)

YonatanKra:
* Added relevant tests for `valueAsNumber()`, `stepUp` and `stepDown`
* Added methods to Readme


rinaok (dismissed):

libs/components/src/lib/number-field/README.md
```diff
@@ -120,6 +120,19 @@ Add the `readonly` attribute to restrict user from changing the number field's v
...
 
+## Methods
+
+<div class="table-wrapper">
+
+| Name            | Returns  | Description                                   |
+|-----------------|----------|-----------------------------------------------|
+| `stepUp`        | `void`   | Increase value by step (1 if step is not set) |
+| `stepDown`      | `void`   | Decrease value by step (1 if step is not set) |
+| `valueAsNumber` | `number` | Returns the value as a number                 |
```
RichardHelm, line 131: `valueAsNumber` should be under properties

RichardHelm (changes requested):

libs/components/src/lib/number-field/README.md
YonatanKra: Should it be at all?
It's part of the HTML spec...

RichardHelm (approved):

libs/components/src/lib/number-field/README.md
RichardHelm: Ah I didn't know that, alright then.

However, and this is a different discussion, maybe yes we should actually include things like this in our docs as it will make it easier and clearer for users.

PR #1390, opened 2023-09-20
feat(nav-item): add icon slot(VIV-1303)

rinaok:


libs/components/src/lib/text-anchor/text-anchor.template.ts
```diff
@@ -5,13 +5,13 @@ import type {
...
 import { classNames } from '@microsoft/fast-web-utilities';
-import { affixIconTemplateFactory } from '../../shared/patterns/affix';
+import { affixIconTemplateFactory, IconWrapper } from '../../shared/patterns/affix';
 import { focusTemplateFactory } from './../../shared/patterns/focus';
 import type { TextAnchor } from './text-anchor';
 
 const getClasses = ({text, icon}: TextAnchor) => classNames(
 	'control',
-	['icon-only', !text && !!icon],
+	['icon-only', (!text && !!icon) || (!text && !!IconWrapper.Span)],
```
RichardHelm, line 14: What does this logic do? `!!IconWrapper.Span` is always going to be true

So it's equivalent to this?
```suggestion
	['icon-only', !text],
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1392, opened 2023-09-27
fix(docs): customization tabs (VIV-1264)

rinaok:


apps/docs/variables-preview/index.js
```diff
@@ -119,7 +119,7 @@ const getCssPropertiesForComponent =(componentData) => {
 
 const renderVariablesStylesheet = (cssProperties) => {
 	return `<style>
-    .vvd-root {
+    vwc-tab[aria-selected="true"] {
```
RichardHelm, line 122: What does this do?

From my understanding, the issue is only with the customization example for tabs

RichardHelm (changes requested):

apps/docs/variables-preview/index.js
RichardHelm: No it does work, it overrides the connotations colors with accouncement so they are all pink

But with your change the selector doesn't match any more and has no effect

I think the purpose of this ticket is just to fix the tabs example to color the tab components as well

rinaok: I didn't understand the ticket right, fixed :)

RichardHelm (approved):

PR #1396, opened 2023-09-28
chore(docs): adds missing component docs comments (VIV-997)

TaylorJ76:


RichardHelm (changes requested):
Looks good, thanks!

Can you please remove the change to package-lock though?

RichardHelm (approved):

PR #1399, opened 2023-10-01
fix(dialog): prevent close when clicking inside and mouseup outside (VIV-1346)

YonatanKra:


libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -159,27 +180,41 @@ describe('vwc-dialog', () => {
...
+			const elementOpenStateAfterClick = element.open;
+
+			expect(elementOpenStateAfterMouseUp).toEqual(true);
+			expect(elementOpenStateAfterClick).toEqual(true);
+		});
+
+		it('should leave the dialog open when mousedown on a non dialog element', async function () {
+			const otherElement = document.createElement('div');
+			const event = createMouseEventOutsideTheDialog('mousedown');
+			Object.defineProperty(event, 'target', {value: otherElement, enumerable: true});
```
RichardHelm, line 201: Isn't `target` just going to get overridden by `dispatchEvent`?

Maybe append the div to the dialog and dispatchEvent on it directly instead

RichardHelm (changes requested):

RichardHelm (approved):

PR #1403, opened 2023-10-02
feat(button): add icon slot (VIV-1340)

rinaok:


libs/components/src/lib/button/button.template.ts
```diff
@@ -34,24 +34,19 @@ function renderIconOrPending(
...
 			[Size.Normal]: '-5',
 			[Size.Expanded]: '-4',
 		};
-		content = `<${progressTag} size="${progressSize[size]}"></${progressTag}>`;
-		classes += ' pending';
-	} else if (icon) {
-		const iconTag = context.tagFor(Icon);
-		content = `<${iconTag} name="${icon}"></${iconTag}>`;
-	}
+		return html`<span class="icon pending">${`<${progressTag} size="${progressSize[size]}"></${progressTag}>`}</span>`;
```
RichardHelm, line 44: We can do it like this, right?
```suggestion
		return html`<span class="icon pending"><${progressTag} size="${progressSize[size]}"></${progressTag}></span>`;
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1405, opened 2023-10-04
chore(docs): improve docs (VIV-000)

RichardHelm:


RichardHelm:
@TaylorJ76 Yes, I agree.
I'm going to go through all the components and fix it. In this PR I'm only doing accordion -> breadcrumb.

So I'm going to fix them when I get to them

PR #1409, opened 2023-10-10
chore(a11y): component a11y tests (VIV-1354)

TaylorJ76:
Adds an axe test to each component unit test.
I have put fixes in where they are quick and easy, others I have skips the test with a comment on why it is failing.
Also added some specific test for a11y attributes where they were not previously covered.


RichardHelm:
Very cool 👍 

I've noticed that some of tests do an additional `expect(await axe(element)).toHaveNoViolations();` at the end while others don't. Is there some kind of logic behind this?

Have you considered adding the check to `elementUpdated()` instead so that they run whenever some UI changes?

libs/components/src/lib/tag-group/tag-group.spec.ts
```diff
@@ -21,4 +21,22 @@ describe('vwc-tag-group', () => {
...
+			element = (await fixture(
+				`<${COMPONENT_TAG}>
+					${`
+						<vwc-tag label="Label 1"></vwc-tag>
+						<vwc-tag label="Label 2"></vwc-tag>
+						<vwc-tag label="Label 3"></vwc-tag>
+					`}
+				</${COMPONENT_TAG}>`
+			)) as TagGroup;
+			element.ariaLabel = 'Tag group';
```
RichardHelm, line 36: I think this can be a bit simpler
```suggestion
				`<${COMPONENT_TAG} aria-label="Tag group">
					<vwc-tag label="Label 1"></vwc-tag>
					<vwc-tag label="Label 2"></vwc-tag>
					<vwc-tag label="Label 3"></vwc-tag>
				</${COMPONENT_TAG}>`
			)) as TagGroup;
```

libs/components/src/lib/tag/tag.spec.ts
```diff
@@ -327,4 +327,30 @@ describe('vwc-tag', () => {
...
+			element.selectable = false;
+			element.removable = true;
+			await elementUpdated(element);
+			exposedHTMLString = `
+				<div role="listbox" aria-label="tag group">
+					${element.shadowRoot?.innerHTML}
+				</div>
+			`;
+
+			expect(await axe(exposedHTMLString)).toHaveNoViolations();
```
RichardHelm, line 353: Again this should be a separate test

RichardHelm (changes requested):

RichardHelm:
> Yes, I have been using the `elementUpdated` to wait for UI changes. In some tests I do more than one axe test to test the component in different states (ini the Avatar component for example)

I get that, I was thinking to run the test as part of `elementUpdated`. But that would probably flag a lot of issues when the component does not have all the required attributes etc....

package.json
```diff
@@ -10,7 +10,8 @@
     "nx": "nx",
     "test": "npx nx run components:test",
     "test:ui": "npx nx run components:e2e",
-    "test:tools": "jest --projects tools"
+    "test:tools": "jest --projects tools",
+    "test:pa11y": "pa11y-ci"
```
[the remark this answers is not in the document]
TaylorJ76: Not needed. Removed

RichardHelm (approved):

PR #1415, opened 2023-10-17
feat(date-range-picker): dual calendar view (VIV-1339)

RichardHelm:
This changes the date-range-picker to always show 2 calendars:
image

The date-picker is unaffected.


rachelbt (approved):
LGTM (as far as I got most of it...).
Not entirely connected to this PR, but why is the date-picker folder under shared? it seems to me that its not belongs there.

RichardHelm:
@rachelbt The base class is there because it is shared between `date-picker` and `date-range-picker`

What's the accessibility issue?

rachelbt:
> @rachelbt The base class is there because it is shared between `date-picker` and `date-range-picker`

yes, I got that - but its like a "sub component" that only 2 components are using it. 
I wonder if its not more correct to place it under components as underlying (like elevation or popup).
If we decide to move it - maybe its better to do it in another PR.

> What's the accessibility issue?
maybe It is me but I didn't managed to use keyboard to select range of dates :(

RichardHelm:
@rachelbt Sure, we can move it in another PR

For me it works fine with the keyboard:
https://github.com/Vonage/vivid-3/assets/86777660/84037ceb-ded4-4aff-bfad-90184663e248

rachelbt (approved):
LGTM

PR #1417, opened 2023-10-18
feat(selectable-box): add component (VIV-1173)

TaylorJ76:


libs/shared/src/lib/sass/mixins/connotation/_mixins.scss
```diff
@@ -19,7 +19,7 @@
 
 	$connotation-mapping: utils.pick-map(config.$connotations, variables.$connotation-mapping);
 
-  @include utils.spread-list-selectors('.connotation', config.$connotations, config.$default) using ($type) {
+  	@include utils.spread-list-selectors('.connotation', config.$connotations, config.$default) using ($type) {
```
rachelbt, line 22: there was no change here - mind reverting it?

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+    <vwc-selectable-box control="radio" checked style="max-inline-size: 450px">
+        <div>Checked radio box</div>
+    </vwc-selectable-box>
+</vwc-layout>
+```
+
+### Clickable
+
+By default, the card's control element (checkbox or radio) is the clickable element. This allows you to use other clickable elements within the box.
+Setting the `clickable` attribute makes the whole card clickable, just make sure the card does not contain other clickable elements.
```
rachelbt, line 77: this is really important, maybe we need to add an alert icon at the beginning of the line?

rachelbt (commented):
went over the doc, html + css.
look really nice :)
I think it will be nice if we will add some use case for it in a group of items.
This can also be good for a good accessibility use.

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.base:not(.clickable):not(.active) {
+	&:hover {
+		background-color: transparent;
+	}
+
+	&:active {
+		background-color: transparent;
+	}
+}
```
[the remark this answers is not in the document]
TaylorJ76: Yeah, I tried it inside and it caused issues.

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.base {
+	@include connotation.connotation(selectable-box);
+	@include appearance.appearance;
+
+	--spacing-unit: 16px;
+
+	position: relative;
+	min-height: calc(var(--spacing-unit) * 2 + 22px);
```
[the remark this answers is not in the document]
TaylorJ76: The min-height is the minimum needed to display the control (radio / checkbox) with the correct spacing.

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
+# Selectable Box
+
+Represents a selectable-box custom element.
```
RichardHelm, line 3: This should be a more meaningful sentence

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+```js
+<script type="module">import '@vonage/vivid/selectable-box';</script>
+```
+
+```html preview
+<vwc-selectable-box aria-label="Box 1">
+    <div>Box content</div>
```
RichardHelm, line 11: Minor, but I guess we don't need to wrap the content in a div
```suggestion
    Box content
```

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+    <vwc-selectable-box connotation="accent" style="max-inline-size: 450px">
+        <div>Accent box</div>
+    </vwc-selectable-box>
+    <vwc-selectable-box connotation="cta" style="max-inline-size: 450px">
+        <div>CTA box</div>
+    </vwc-selectable-box>
+</vwc-layout>
+```
+
+### Control
```
RichardHelm, line 36: While control is a very good name, in other components `control` is the control element and that is assumed to be the case by some code like FormAssociated.
So to avoid a possible name conflicts in the future I would change it to `control-type`

```suggestion
### Control type
```

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Set the `control` attribute to change the box's selectable control
+It accepts a subset of predefined values.
+When `control` is set to `radio`, it is the consuming app's responsibility to ensure only 1 selectable box in a group is checked at a time.
```
RichardHelm, line 40:
```suggestion
When `control` is set to `radio`, it is the consuming app's responsibility to ensure only one selectable box in a group is checked at a time.
```

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Type: `'checkbox'`, `'radio'`
+- Default: `'checkbox'`
+
+```html preview
+<vwc-layout gutters="small" row-spacing="small" column-basis="block">
+    <vwc-selectable-box control="checkbox" style="max-inline-size: 450px">
+        <div>Checkbox control</div>
+    </vwc-selectable-box>
+    <vwc-selectable-box control="radio" style="max-inline-size: 450px">
```
RichardHelm, line 50: There's a problem with the radio, it cannot be unchecked

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Type: `boolean`
+- Default: `false`
+
+```html preview
+<vwc-layout gutters="small" row-spacing="small" column-basis="block">
+    <vwc-selectable-box clickable style="max-inline-size: 450px">
+        <div>Clickable accent box</div>
+    </vwc-selectable-box>
+    <vwc-selectable-box clickable connotation="cta" style="max-inline-size: 450px">
```
RichardHelm, line 87: There's a problem with the focus ring here:
image

libs/components/src/lib/selectable-box/definition.ts
```diff
@@ -0,0 +1,25 @@
...
+	SelectableBox.compose<FoundationElementDefinition>({
+		baseName: 'selectable-box',
+		template: template as any,
+		styles,
+	});
+
+/**
+ * @internal
+ */
+export const selectableBoxRegistries = [selectableBoxDefinition()];
```
RichardHelm, line 18: If your template uses other components you need to add them here (check other components)
That would be: focus, checkbox, radio

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,83 @@
...
+	$shades: firm soft faint,
+	$default: accent
+);
+@use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
+
+@use "../../../../shared/src/lib/sass/mixins/appearance/config" as appearance-config with (
+	$appearances: outlined,
+	$states: idle hover active,
+	$default: outlined
+);
```
RichardHelm, line 15: More of a design problem, but when the card is `clickable`, IMO it should have a hover, active (pressed), and checked state but these have not been defined in Figma.
Also the current appearance does not match up with what's in Figma color wise

libs/components/src/lib/selectable-box/selectable-box.template.ts
```diff
@@ -0,0 +1,75 @@
...
+ * 
+ */
+function renderControl(x: SelectableBox, c: ElementDefinitionContext) {
+	const Control = x.control === 'radio' ? Radio : Checkbox;
+	const tagName = c.tagFor(Control);
+	return html<SelectableBox>`
+		<${tagName} 
+			${x.ariaLabel !== null && !x.clickable ? `aria-label="${x.ariaLabel}"` : ''}
+			${x.ariaLabelledby !== null && !x.clickable ? `aria-labelledby="${x.ariaLabelledby}"` : ''}
+			${x.clickable ? 'tabindex="-1" aria-hidden="true"' : ''}
```
RichardHelm, line 37: Something is getting messed up here:
image

libs/components/src/lib/selectable-box/selectable-box.template.ts
```diff
@@ -0,0 +1,75 @@
...
+		tabindex="${x => x.clickable ? '0' : null}"
+		role="${x => x.clickable ? x.control || 'checkbox' : null}"
+		aria-checked="${x => x.clickable && x.checked ? x.checked : null}"
+		aria-label="${x => x.clickable ? x.ariaLabel : null}"
+		aria-labelledby="${x => x.clickable ? x.ariaLabelledby : null}"
+		@keydown="${(x, c) => x.handleKeydown(c.event as KeyboardEvent)}"
+		@click="${x => x.clickable ? x.handleCheckedChange() : null}"
+	>
+		${(x) => x.clickable ? focusTemplate : ''}
+		${x => renderControl(x, context)}
```
RichardHelm, line 71: This will recreate the control every time something changes. It's better to have the functions on individual attributes so that only those will update when they change
```suggestion
		${renderControl(context)}
```

libs/components/src/lib/selectable-box/selectable-box.ts
```diff
@@ -0,0 +1,91 @@
...
+	*
+	* @public
+	* HTML Attribute: no-padding
+	*/
+	@attr({ mode: 'boolean'}) clickable = false;
+
+	handleCheckedChange() {
+		if (this.control === 'radio' && this.checked) return;
+		this.checked = !this.checked;
+		this.$emit('change', { checked: this.checked });
```
RichardHelm, line 84: This is reasonable but in other components we're not emitting the checked value, so for consistency I would remove it
```suggestion
		this.$emit('change');
```

libs/components/src/lib/selectable-box/selectable-box.ts
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	handleKeydown(event: KeyboardEvent) {
+		if (event.code === 'Space' && this.clickable) this.handleCheckedChange();
```
RichardHelm, line 88: Should this accept Enter as well?

libs/components/src/lib/selectable-box/selectable-box.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+	describe('clickable', () => {
+		it('should set clickable class on the base element', async function () {
+			const baseElement = getBaseElement(element);
+			element.clickable = true;
+			await elementUpdated(element);
+			expect(baseElement?.classList?.contains('clickable')).toBe(true);
+		});
+	});
+
+	describe('changed event', () => {
```
RichardHelm, line 107:
```suggestion
	describe('change event', () => {
```

libs/components/src/lib/selectable-box/selectable-box.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+	describe('tight', () => {
+		it('should set tight class on the base element', async function () {
+			const baseElement = getBaseElement(element);
+			element.tight = true;
+			await elementUpdated(element);
+			expect(baseElement?.classList?.contains('tight')).toBe(true);
+		});
+	});
+
+	describe('checked', () => {
```
RichardHelm, line 82: I think it's never getting tested that checked is updated when clicking on the card / control

libs/components/src/lib/selectable-box/selectable-box.template.ts
```diff
@@ -0,0 +1,75 @@
...
+	definition: FoundationElementDefinition
+) => ViewTemplate<SelectableBox> = (
+	context: ElementDefinitionContext
+) => {
+	const focusTemplate = focusTemplateFactory(context);
+	return html<SelectableBox>`
+	<div
+		class="${getClasses}"
+		tabindex="${x => x.clickable ? '0' : null}"
+		role="${x => x.clickable ? x.control || 'checkbox' : null}"
```
RichardHelm, line 63: Mhm, maybe we should use a different role. Not sure if making the whole thing a checkbox/radio is really accurate.
E.g. Google Meet backgrounds use role=button with aria-pressed:
image

libs/components/src/lib/selectable-box/selectable-box.ts
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	* Controls the selected state of the box
+	*
+	* @public
+	* HTML Attribute: no-padding
+	*/
+	@attr({ mode: 'boolean'}) clickable = false;
+
+	handleCheckedChange() {
```
RichardHelm, line 81: This (and keydown handler) needs internal annotation or it will be exported as part of the custom element manifest.
Additionally, it is preferable to prefix it with an underscore to discourage users from calling it anyway.
```suggestion
/**
 * @internal
 */
	_handleCheckedChange() {
```

libs/components/src/lib/selectable-box/selectable-box.ts
```diff
@@ -0,0 +1,91 @@
...
+export type SelectableBoxSize = Size.Normal | Size.Condensed;
+
+export type SelectableBoxControl = 'checkbox' | 'radio';
+
+
+/**
+ * Base class for selectable-box
+ *
+ * @public
+ */
```
RichardHelm, line 26: This needs annotation for the default slot and change event (see other components). It will cause these things to be added to the custom element manifest, which is then used to generate the vue wrappers.

RichardHelm (changes requested):
Looks great! I added some comments, most of which should be minor

libs/components/src/lib/selectable-box/selectable-box.scss
rachelbt: I agree. 
Something is not fully aligned here with our appearances, design-wise - hence, you also needed to add the transparent background instead of using the automated colors of fill , outline, text. @AyalaBu ?

TaylorJ76: The reason for adding the transparent background bit was to override the hover styles when only the checkbox/radio is clickable.

TaylorJ76: I will reach out to @AyalaBu to get confirmation on the appearance because it hover state was accounted for in the design.

RichardHelm (changes requested):

libs/components/src/lib/selectable-box/README.md
```diff
@@ -0,0 +1,206 @@
...
+        <div>Clickable CTA box</div>
+    </vwc-selectable-box>
+</vwc-layout>
+```
+
+### Size
```
rachelbt, line 93: I'm sorry - but this is not size, size effect the sizes of the component.
I'm sorry if I was not clear at the first time.
I would stay with `spacing` but maybe use the small - medium like in column spacing of layout.
Again raising the issue of discarding the tight and adding `none` to spacing.
@RichardHelm @rinaok WDYT?

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,77 @@
...
+@use "../../../../shared/src/lib/sass/mixins/appearance" as appearance;
+
+@supports selector(:focus-visible) {
+	.base:focus-visible {
+		outline: none;
+	}
+}
+
+:host {
+	display: flex;
```
rachelbt, line 26: why is this needed?

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	&.clickable {
+		cursor: pointer;
+	}
+
+	&::slotted(img) {
+		vertical-align: middle;;
+	}
+}
```
rachelbt, line 60: this does nothing :) 
It needs to be extracted to be outside of `.base `
I'm not sure that we need to add this at all - this should be the consumers concern.

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -0,0 +1,77 @@
...
+	.base:not(:focus-visible) > & {
+		display: none;
+	}
+}
+
+.control {
+	position: absolute;
+	inset-block-start: var(#{variables.$size-unit});
+	inset-inline-end: var(#{variables.$size-unit});
+}
```
rachelbt, line 77:
```suggestion
}

	::slotted(*) {
		border-radius: 6px;
	}
```

rachelbt (commented):
O love the use examples!!
left some comments about the api naming :)

libs/components/src/lib/selectable-box/selectable-box.scss
RichardHelm: Not sure if this is good, as it will add the radius to all slotted children.

But the issue needs to be fixed somehow:
image

RichardHelm (approved):

RichardHelm (approved):

PR #1420, opened 2023-10-20
fix(components): fix min/maxlength validation (VIV-1174)

RichardHelm:
This fixes the issue with min/maxlength for the components with it.

The problem is that constraint validation on min/maxlength in HTML only applies after user interaction with the field (which is insane behaviour). See: https://stackoverflow.com/questions/66896018/html-input-checkvalidity-always-returns-true-even-with-minlength-violations

The proxy element that we use has it's value set programmatically and therefore is never interacted with. There is also appears no way to do this programatically.

Instead we will need to get the validity from the actual control. This is also how Google's MWC3 does it.

We also need to make sure we're not unnecessary setting the value programatically because it will reset the "user interacted flag" in the browser.

In this PR we do it for text field and area but not number-field, because number fields sets the value programatically as part of it's normal working operation.
However min/maxlength are of little use with the number-field anyway, as you would use min/max instead.


libs/components/src/shared/utils/Reflector.ts
```diff
@@ -0,0 +1,88 @@
...
+				case 'boolean-attribute':
+					DOM.setBooleanAttribute(this.#target, reflectedProperty.destination, Boolean(value));
+					break;
+				case 'attribute':
+					DOM.setAttribute(this.#target, reflectedProperty.destination, value);
+					break;
+				case 'property':
+					if (reflectedProperty.skipIfEqual && (this.#target as any)[reflectedProperty.destination] === value) {
+						return;
+					}
```
RichardHelm, line 82: This part is needed because even when setting an input's value to the value it already is, it would count as programatically setting it and break the validation.

libs/components/src/lib/number-field/README.md
RichardHelm, line 1: The number-field is not changed, just updating the docs

TaylorJ76 (approved):

PR #1423, opened 2023-10-24
fix(checkbox): add display: flex to host (VIV-1293)

rachelbt:


libs/components/src/lib/checkbox/README.md
```diff
@@ -153,3 +153,16 @@ The default slot allows you to use rich content as the checkbox's label.
 	I agree to <a href="https://www.vonage.com/legal/" target="_blank">Vonage Terms of Service</a>
 </vwc-checkbox>
 ```
+
+## Use case
```
RichardHelm, line 157: Not a fan of adding this weird looking use case just to test it. Maybe it can just be in the UI test?

RichardHelm (approved):

PR #1426, opened 2023-10-27
fix(a11y): adds a11y attributes to fulfil axe tests (VIV-1361)

TaylorJ76:
Adds localized `aria-label` text for internal buttons which can be overridden by passing a specific value as an attribute.
Also adds `aria-label`, `aria-labelledby` and `aria-describedby` attributes to certain interactive elements whcih needed them.


libs/components/jestGlobalSetup.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = async () => {
+	process.env.TZ = 'UTC';
```
TaylorJ76, line 2: This is to stop date based tests failing because of daylight savings time in the UK

libs/components/src/lib/checkbox/checkbox.template.ts
```diff
@@ -41,6 +41,9 @@ export const CheckboxTemplate: FoundationElementTemplate<ViewTemplate<Checkbox>,
 
 	return html`<div class="${getClasses}"
 			role="checkbox"
+			aria-label="${x => x.ariaLabel}"
+			aria-labelledby="${x => x.ariaLabelledby}"
+			aria-describedby="${x => x.ariaDescribedby}"
```
RichardHelm, line 46: This does not work, you cannot refer to to an element outside the shadow DOM by ID

I would remove it as it's not currently possible
```suggestion
```

libs/components/src/lib/checkbox/checkbox.ts
```diff
@@ -29,6 +29,9 @@ export type CheckboxConnotation = Extract<Connotation, | Connotation.Accent | Co
 @errorText
 @formElements
 export class Checkbox extends FoundationCheckbox {
+	@attr({attribute: 'aria-label'}) override ariaLabel: string | null = null;
```
RichardHelm, line 32: Not necessary to override this

```suggestion
```

libs/components/src/lib/checkbox/checkbox.template.ts
```diff
@@ -41,6 +41,9 @@ export const CheckboxTemplate: FoundationElementTemplate<ViewTemplate<Checkbox>,
 
 	return html`<div class="${getClasses}"
 			role="checkbox"
+			aria-label="${x => x.ariaLabel}"
```
RichardHelm, line 44: We use this pattern of passing ariaLabel to a child in several places.

However, this has the problem that the aria-label attribute also takes effect on the component itself and labels it.
image

I tested it in Voiceover and it will announce "A default checkbox, group" and only when going inside the group "A default checkbox, unticked tickbox..."

In Material Design 3 it's solved by adding `role=presentation` on the component itself.

libs/components/src/lib/number-field/number-field.ts
```diff
@@ -46,6 +47,8 @@ function makeStep(element: NumberField, direction: number) {
 @errorText
 @formElements
 export class NumberField extends FastNumberField {
+	@attr({attribute: 'add-button-aria-label'}) addButtonAriaLabel: string | null = null;
+	@attr({attribute: 'subtract-button-aria-label'}) subtractButtonAriaLabel: string | null = null;
```
RichardHelm, line 51: I think, like their label, increment / decrement is a more fitting name
```suggestion
	@attr({attribute: 'increment-button-aria-label'}) addButtonAriaLabel: string | null = null;
	@attr({attribute: 'decrement-button-aria-label'}) subtractButtonAriaLabel: string | null = null;
```

libs/components/src/lib/progress-ring/progress-ring.template.ts
```diff
@@ -29,6 +29,9 @@ export const ProgressRingTemplate: (
 ) => html`
 <div
         role="progressbar"
+        aria-label="${x => x.ariaLabel}"
+        aria-labelledby="${x => x.ariaLabelledby}"
+        aria-describedby="${x => x.ariaDescribedby}"
```
RichardHelm, line 34: Same concerns here

libs/components/src/lib/radio/README.md
```diff
@@ -72,3 +72,9 @@ Use the `value` member to set the radio's value.
 ```html preview
 <vwc-radio value="my-value"></vwc-radio>
 ```
+
+## Accessibility
+
+- If a label is not provided either through the `label` attribute, then the checkbox will needs accessible label.
```
RichardHelm, line 78:
```suggestion
- If a label is not provided either through the `label` attribute, then the checkbox needs an accessible label.
```

libs/components/src/lib/radio/radio.template.ts
```diff
@@ -27,6 +27,9 @@ export const RadioTemplate: (context: ElementDefinitionContext) => ViewTemplate<
 	return html<Radio>`
 	<div class="${getClasses}"
 		role="radio"
+		aria-label="${x => x.ariaLabel}"
+		aria-labelledby="${x => x.ariaLabelledby}"
+		aria-describedby="${x => x.ariaDescribedby}"
```
RichardHelm, line 32: And here

libs/components/src/lib/slider/slider.template.ts
```diff
@@ -38,7 +38,10 @@ export const SliderTemplate: (context: ElementDefinitionContext) => ViewTemplate
 	<div
 		role="slider"
 		tabindex="${x => (x.disabled ? null : 0)}"
-		aria-valuetext="${x => x.valueTextFormatter(x.value)}"
+		aria-label="${x => x.ariaLabel}"
+		aria-labelledby="${x => x.ariaLabelledby}"
+		aria-describedby="${x => x.ariaDescribedby}"
+		aria-valuetext="${x => x.ariaValuetext || x.valueTextFormatter(x.value)}"
```
RichardHelm, line 44: And here

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -41,7 +41,9 @@ describe('vwc-slider', () => {
 	describe('disabled', () => {
 		it('should set disabled class when disabled is true', async () => {
 			const classes = await setBoolAttributeOn(element, 'disabled');
+			const control = await getControlElement(element);
 			expect(classes).toContain('disabled');
+			expect(control.getAttribute('aria-disabled')).toBe('true');
```
RichardHelm, line 46: Maybe this should be it's own test

RichardHelm (changes requested):

libs/components/src/lib/checkbox/checkbox.ts
TaylorJ76: When I remove this the a11y tests are failing.

RichardHelm (approved):

RichardHelm (approved):

PR #1429, opened 2023-10-30
feat(slider): add cta connotation (VIV-1373)

rinaok:


libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -139,6 +140,31 @@ describe('vwc-slider', () => {
 		});
 	});
 
+	describe('connotation', function () {
+		const possibleConnotations = [Connotation.Accent,
+			Connotation.CTA
+		];
```
RichardHelm, line 146:
```suggestion
		] as const;
```

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -139,6 +140,31 @@ describe('vwc-slider', () => {
...
+
+		it('should leave connotation class empty if not set', async function () {
+			possibleConnotations.forEach(connotation => {
+				expect(getControlElement(element)
+					?.classList
+					.contains(connotation))
+					.toEqual(false);
+			});
+		});
```
RichardHelm, line 155: Not sure if this test is really needed
```suggestion
		it('should not set any connotation classes when no connotation is set', async function () {
			possibleConnotations.forEach(connotation => {
				expect(getControlElement(element)
					?.classList
					.contains(`connotation-${connotation}`))
					.toEqual(false);
			});
		});
```

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -139,6 +140,31 @@ describe('vwc-slider', () => {
...
+		it('should set a connotation class', async function () {
+			const connotation = possibleConnotations[1];
+			(element.connotation as Connotation) = connotation;
+			await elementUpdated(element);
+			expect(getControlElement(element)
+				?.classList
+				.contains(`connotation-${connotation}`))
+				.toEqual(true);
+		});
+	});
```
RichardHelm, line 166: I think we can write it like this for all connotations
```suggestion
		it.each(possibleConnotations)('should set the connotation class for "%s"', async function (connotation) {
			element.connotation = connotation;
			await elementUpdated(element);
			expect(getControlElement(element)
				?.classList
				.contains(`connotation-${connotation}`))
				.toEqual(true);
		});
	});
```

RichardHelm (changes requested):
Nice 👍  Got some minor comments

rachelbt (commented):
LGTM,
the only thing is that is missing is the export of the connotation in the definition.ts:
export type { SliderConnotation } from './slider';

RichardHelm (approved):

PR #1432, opened 2023-10-31
feat(split-button): add attribute for indicator aria-label and localize default (VIV-1261)

RichardHelm:


libs/components/src/lib/split-button/README.md
```diff
@@ -157,8 +157,33 @@ Add the `disabled` attribute to disable the split button.
...
+<style>
+  html { /* for demo purposes */
+    block-size: 100px;
+		text-align: center;
+  }
+</style>
+
+<vwc-split-button id="splitButton" appearance="filled" icon="compose-line" aria-label="Write a new message">
+</vwc-split-button>
+<vwc-tooltip id="tooltip" text="Write a new message"></vwc-tooltip>
```
rachelbt, line 172: I wonder if the tooltip need to be inside the split-button - same as menu. WDYT?

rachelbt (commented):
Nice! That was quick :) 👍 
I think we should add tests for the aria-label same as we have in vwc-button.

libs/components/src/lib/split-button/split-button.template.ts
```diff
@@ -67,9 +65,11 @@ export const SplitButtonTemplate: (
 	definition: FoundationElementDefinition
 ) => ViewTemplate<SplitButton> = (context: ElementDefinitionContext) => {
 	return html<SplitButton>`
-		<div class="base" role="group">
-			${actionButton(context)}
-			${indicatorButton(context)}
-			<slot></slot>
-		</div>`;
+		<template role="presentation">
```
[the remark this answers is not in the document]
RichardHelm: Otherwise specifying aria-label with label the host element itself

libs/components/src/lib/split-button/README.md
RichardHelm: Sure, moved it inside 👍
I don't think it's reachable though, that's why it has the aria-label

rachelbt: its still outside...

RichardHelm: Yes, I forgot to push it... Done now

TaylorJ76 (approved):

rachelbt (approved):
LGTM

PR #1436, opened 2023-11-03
fix: fixes issue with keypress when not clickable

TaylorJ76:
- Fixes a bug with keypress when box is not clickable
- Fixes a11y issue with checkbox when box is clickable
- Follows the same method for `aria-label` as other components and removes `aria-labelledby`

### DIFF
--- a/libs/components/src/lib/checkbox/checkbox.template.ts
+++ b/libs/components/src/lib/checkbox/checkbox.template.ts
@@ -39,7 +39,7 @@ export const CheckboxTemplate: FoundationElementTemplate<ViewTemplate<Checkbox>,
 	const focusTemplate = focusTemplateFactory(context);
 	const iconTag = context.tagFor(Icon);
 
-	return html`<template role="${x => x.ariaLabel ? 'presentation' : null}">
+	return html`<template role="presentation">
 		<div class="${getClasses}"
 			role="checkbox"
 			aria-label="${x => x.ariaLabel}"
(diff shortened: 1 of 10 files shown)


libs/components/src/lib/selectable-box/selectable-box.template.ts
```diff
@@ -87,5 +84,5 @@ export const SelectableBoxTemplate: (
 		${radio(context)}
 		<slot></slot>
 	</div>
-`;
+<//template>`;
```
RichardHelm, line 87:
```suggestion
</template>`;
```

libs/components/src/lib/checkbox/checkbox.template.ts
```diff
@@ -39,15 +39,15 @@ export const CheckboxTemplate: FoundationElementTemplate<ViewTemplate<Checkbox>,
...
+	return html`<template role="presentation">
 		<div class="${getClasses}"
 			role="checkbox"
 			aria-label="${x => x.ariaLabel}"
 			aria-checked="${x => x.checked}"
 			aria-required="${x => x.required}"
 			aria-disabled="${x => x.disabled}"
 			aria-readonly="${x => x.readOnly}"
-			tabindex="${x => (x.disabled ? null : 0)}"
+			tabindex="${x => (x.tabindex || x.disabled ? null : 0)}"
```
RichardHelm, line 50: This doesn't seem to work. When tabindex is present, is doesn't allow tabbing the checkbox

libs/components/src/lib/selectable-box/selectable-box.ts
```diff
@@ -98,8 +82,9 @@ export class SelectableBox extends FoundationElement {
 	 * @internal
 	 */
 	_handleKeydown(event: KeyboardEvent) {
+		if (!this.clickable) return true;
 		if (event.code === 'Space' || event.code === 'Enter' && this.clickable)
```
RichardHelm, line 86: I guess is should've been like this
```suggestion
		if ((event.code === 'Space' || event.code === 'Enter') && this.clickable) 
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1440, opened 2023-11-10
fix(select): aria-label repeating the label text in screenreader and not updating when changed dynamically (VIV-1379)

TaylorJ76:


libs/components/src/lib/select/select.spec.ts
```diff
@@ -99,12 +99,12 @@ describe('vwc-select', () => {
 				.toBeNull();
 		});
 
-		it('should set arialabel on host', async function () {
+		it('should not set aria-labelledby on host', async function () {
```
rinaok, line 102:
```suggestion
		it('should not set aria-label on host', async function () {
```

rinaok (approved):

PR #1443, opened 2023-11-13
fix(alert): accessability (VIV-1381)

rinaok:


libs/components/src/lib/alert/alert.ts
```diff
@@ -91,29 +95,41 @@ export class Alert extends FoundationElement {
...
 	 * HTML Attribute: open
 	 */
-	@attr({ mode: 'boolean'	}) open = false;
+	@attr({ mode: 'boolean' }) open = false;
 	openChanged(oldValue: boolean, newValue: boolean): void {
 		if (oldValue === undefined) return;
 		this.$emit(newValue ? 'open' : 'close');
 		this.#setupTimeout();
+		if (newValue) {
+			this.style.display = 'block';
```
rachelbt, line 116:
```suggestion
			this.style.display = 'inline';
```
the initial display of a host is inline.
I don't see a justification here to change it to block (even if it does not effect on alert visually)

rachelbt (commented):
Look 👍 
Left one comment

TaylorJ76 (approved):

PR #1446, opened 2023-11-14
fix(selectable-box): a11y issues when clickable (VIV-1392)

TaylorJ76:
When the selectable-box is clickable it creates an a11y issue using the checkbox / radio components inside. (Nesting interactive elements).

This PR makes the control components inert when the box is clickable.
It also update jest-axe to the latest version.

It also deprecates `clickable` and adds `clickable-box` to be more clear of its intention.


RichardHelm:
Wouldn't it be possible to just use the `inert` attribute with checkbox/radio instead? It will remove the element from the accessibility tree so it should have the same effect

> It also renames clickable to container-clickable to be more clear of its intention.
2 things about this:
- I'm not sure if `container-clickable` is much clearer than `clickable`. In AWS cloudscape they call it `entireCardClickable`, so maybe we could call it `entire-box-clickable`?
- It's a breaking change, maybe we should keep `clickable` as a deprecated attribute for backwards compatibility

rachelbt:
> Wouldn't it be possible to just use the `inert` attribute with checkbox/radio instead? It will remove the element from the accessibility tree so it should have the same effect
> 
> > It also renames clickable to container-clickable to be more clear of its intention.
> > 2 things about this:
> 
> * I'm not sure if `container-clickable` is much clearer than `clickable`. In AWS cloudscape they call it `entireCardClickable`, so maybe we could call it `entire-box-clickable`?
> * It's a breaking change, maybe we should keep `clickable` as a deprecated attribute for backwards compatibility

Since its under alpha there's no issue with breaking change :)
I think clickable is so much better then long properties. `entire-box-clickable` is sooooooo long :( 
How about `clickable-box`?

libs/shared/src/lib/sass/mixins/appearance/_mixins.scss
```diff
@@ -23,8 +23,8 @@ $selectors: (
   hover: '&:where(#{$hover}):where(:not(#{$disabled}, #{$readonly}))',
   disabled: '&:where(#{$disabled})',
   readonly: '&:where(#{$readonly}):where(:not(#{$disabled}))',
-  selected: '&:where(#{$selected}):where(:not(#{$disabled}, #{$hover}))',
```
rachelbt, line 26: since we currently do not have hover state UI test - this may effect places where there's no definition for selected+hover (like subtle).

libs/components/src/lib/selectable-box/selectable-box.scss
```diff
@@ -4,15 +4,15 @@
...
+	$shades: firm pale fierce faint,
 	$default: accent
 );
 @use "../../../../shared/src/lib/sass/mixins/connotation" as connotation;
 
 @use "../../../../shared/src/lib/sass/mixins/appearance/config" as appearance-config with (
-	$appearances: selectablebox,
-	$states: idle hover checked checkedAndHover,
-	$default: selectablebox
+	$appearances: duotone,
```
rachelbt, line 13: please also remove the selectablebox from the mapping in` _vriables.scss`

rachelbt (commented):
👍

libs/shared/src/lib/sass/mixins/appearance/_mixins.scss
TaylorJ76: I have gone through the components where this could have been an issue and I'm pretty sure they all behave as they should.
The alternative to this is to override the hover styles on a selected card, but I don't really want to do that anything that hacky.

libs/components/src/lib/selectable-box/selectable-box.spec.ts
```diff
@@ -133,78 +156,132 @@ describe('vwc-selectable-box', () => {
...
 				it('should not emit the change event when the radio is already checked', async () => {
-					element.controlType = 'radio';
-					element.checked = true;
-					await elementUpdated(element);
+					element = (await fixture(
+						`<${COMPONENT_TAG} control-type="radio" checked clickable-box></${COMPONENT_TAG}>`
+					)) as SelectableBox;
 					baseElement.click();
 
 					expect(spy).not.toHaveBeenCalled();
```
RichardHelm, line 283: This doesn't test the right thing, spy is still registered on the original element not the new one

RichardHelm (changes requested):
Nice! Just one small thing

RichardHelm (approved):

PR #1447, opened 2023-11-16
fix(typography): bold style using b or strong tag (VIV-1317)

rachelbt:


libs/styles/src/lib/core/typography.scss
```diff
@@ -76,68 +76,63 @@ $selectors: (
...
+	h2,
+	.heading2,
+	h3,
+	.heading3,
+	h4,
+	.heading4 {
+
+		#{map.get($selectors, "bold")} {
+			font-weight: 500;
+		}
```
RichardHelm, line 123: Why add this? Headings have weight 500, so then the bold tags are not going to do anything inside headings

libs/styles/src/lib/core/typography.scss
```diff
@@ -66,7 +66,7 @@ $selectors: (
 	}
 
 	#{map.get($selectors, "bold")} {
-		font: var(#{constants.$vvd-typography-base-bold});
+		font-weight: 600;
```
RichardHelm, line 69: Let's leave this for now, but this should probably be a token

RichardHelm (changes requested):

libs/styles/src/lib/core/typography.scss
rachelbt: but they do!
If this is bot here - it gets 600 - and its extra bold in the headers

libs/styles/src/lib/core/typography.scss
RichardHelm: Yes I understand. I mean the value 600 should be from a token as it should be configurable.

So if someone wants their bold text to be 700 they can change `typography-base-bold` but here it remains 600

libs/styles/src/lib/core/typography.scss
RichardHelm: Yes it becomes extra bold as expected when using bold tags.

But I can see the argument headings are already considered bold, so bold tags shouldn't do anything so I will approve this

RichardHelm (approved):

libs/styles/src/lib/core/typography.scss
rachelbt: bold is 600, 700 is extra-bold.
Either way - if someone is using <b> and want it to be 700, then he or she can run this over with css.
I'm not sure how common is the use of <b> or <strong> anyway.

PR #1448, opened 2023-11-16
feat(badge): add min-inline-size (VIV-1403)

rinaok:


libs/components/src/lib/badge/ui.test.ts
```diff
@@ -35,6 +35,7 @@ test('should show the component', async ({ page }: { page: Page }) => {
 	<vwc-badge text='warning' appearance='duotone' connotation='warning'></vwc-badge>
 	<vwc-badge text='alert' appearance='duotone' connotation='alert'></vwc-badge>
 	<vwc-badge text='large overflowing text' style="display: block; inline-size: 60px;"></vwc-badge>
+	<vwc-badge text='min-inline-size' style="--badge-min-inline-size: 500px;"></vwc-badge>
```
rachelbt, line 38: the image that is generated is only 424px - maybe here the min-size can be 400 or 300 so the badge will not be cut off.

rachelbt (changes requested):
I don't think we need the css variable.
I think its better to set the `:host` with `display: inline-block` and set `width 100%` on `.base`.
This is better for several reasons:
1. if badge has flex parent, or is inside grid - its stretches to get the width of the space / column.
2. we have text-ellipsis that works if there's a width on the host (+ adding display on the host - the 2 of them are the consumer's responsibility). Adding `display: inline-block` on the host will make this easier to use.
3. its more dynamic - the consumer can set either min width, max-width, or width or all of them.
4. less code on our side :)

libs/components/src/lib/badge/badge.scss
```diff
@@ -24,6 +27,7 @@
 	#{variables.$badge-block-size}: #{size.$vvd-size-ultra-condensed};
 
 	display: inline-flex;
+	width: 100%;
```
rachelbt, line 30:
```suggestion
	inline-size: 100%;
```
it might cause you stylelint issue

rachelbt (commented):
looks so much better.
WDYT about adding usa case - one for long text that is trimmed and one for badge with a set min-width?

libs/components/src/lib/badge/README.md
```diff
@@ -128,6 +128,18 @@ If set, the `icon` attribute is ignored.
 </vwc-badge>
 ```
 
+## Use Cases
+
+### Trim Text
+```html preview
+<vwc-badge text='with overflowing text' style="display: block; inline-size: 60px;"></vwc-badge>
```
rachelbt, line 135:
```suggestion
<vwc-badge text='with overflowing text' style="inline-size: 60px;"></vwc-badge>
```
the display:block is not needed - we have diplay: inline-block now on the host :)

rachelbt (commented):
sorry - one last comment 🙏

rachelbt (approved):
LGTM

PR #1451, opened 2023-11-20
fix(text-field): fix errors in non-DOM or JSDOM environments (VIV-1393)

RichardHelm:
Two things:
* lazily create stylesheet: prevent work done by simply importing the file
* do nothing instead of error if `replaceSync` or `adoptedStyleSheets` is not supported, like when using JSDOM

### DIFF
--- a/libs/components/src/lib/text-field/text-field.ts
+++ b/libs/components/src/lib/text-field/text-field.ts
@@ -3,6 +3,7 @@ import {
 	TextField as FoundationTextfield,
 } from '@microsoft/fast-foundation';
 import { attr, observable } from '@microsoft/fast-element';
+import { memoizeWith } from 'ramda';
 import type { Appearance, Shape } from '../enums';
 import {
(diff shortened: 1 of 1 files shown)


libs/components/src/lib/text-field/text-field.ts
```diff
@@ -27,18 +28,37 @@ export type TextFieldShape = Extract<Shape, Shape.Rounded | Shape.Pill>;
 // See bug: https://bugs.webkit.org/show_bug.cgi?id=223814
 // As a workaround we add a stylesheet to root of text-field to apply the styles
 const safariWorkaroundClassName = '_vvd-3-text-field-safari-workaround';
```
rachelbt, line 30:
```suggestion
// Will be removed once VIV-1413 is done
const safariWorkaroundClassName = '_vvd-3-text-field-safari-workaround';
```

rachelbt (approved):
👍

PR #1452, opened 2023-11-21
feat(appearance): opacity POC (VIV-1383)

rachelbt:


rachelbt:
this PR was open in order to allow us to check the ability of generating opacity in appearances.
Closing the PR since there's not design yet.

PR #1453, opened 2023-11-21
feat(dialog): add non-dismissive attribute (VIV-1351)

RichardHelm:
- adds `non-dismissible` attribute
  - consumers may need to fully disable dismissing, not just on clicking scrim 
  - having non-dismissible be the default would be a breaking change
  - other components follow the convention of a `removable` prop so we could call it `non-removable` instead. However I prefer `non-dismissible` because:
    - tag/banner do actually remove themselves from the DOM (alert does not)
    - removable is not correct language to use with dialog
- fixes focus style. This is because Chrome decides to show focus when pressing esc on non-dismissible dialog 
- try to make docs clearer


TaylorJ76 (changes requested):

rinaok (changes requested):
Commented in the issue: https://jira.vonage.com/browse/VIV-1351

rachelbt (changes requested):
I'm aware of the face that 2 people already commented, I didn't go over the code - but I do have a comment to raise regarding the change, something that we need to consider before we change the API.

It feels to me that the change is really different from other components in vivid.
Maybe we need to split it into 2 properties:
1. removable - as all of our other components - that controls the X button and its removing the component.
2. dismissible, false by default - hat deals with the scrim.
I think that there may be cases where there will be no need for light dismiss but the X button will still be needed.

libs/components/src/lib/dialog/README.md
```diff
@@ -1,8 +1,7 @@
 # Dialog
 Represents a part of an application that a user interacts with to perform a task.
 
-All [native attributes of `dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) are supported as well as some enhancements.
-
+The dialog is uses the native [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element.
```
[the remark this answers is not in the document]
RichardHelm:
```suggestion
The dialog uses the native [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element.
```

RichardHelm:
@rinaok Copying your comment here for completeness:

> The native dialog does not close when the Scrim is clicked, but in vivid it does by design: https://jira.vonage.com/browse/VIV-1186
All dialogs have a close button by default.
> In my opinion, the fix should be as follows:
> Using the Scrim button will not close the dialog - it will not be a breaking change since users still have the option to close it using the X button.
When dismissible attribute (as in other vivid components) is true, the dialog will be closed (even the regular dialog and not just the one with the scrim).

So, I've done some more research:

### Vivid conventions

* tag, banner:
  * **removable**: adds X button which removes the element from the DOM
* alert:
  * **removable**: adds X button which closes the alert
* popup (internal component):
  * **dismissible**: adds X button which closes the popup
    * attribute is internal (and unused), so not actually part of our API
* menu:
  * **auto-dismiss**: closes on click outside

### Other libraries

Library | Close on scrim by default | scrim API | close button
--- | --- | --- | ---
Microsoft FluentUI | ✅ | `isBlocking` | unremovable
Atlassian | ✅ | `shouldCloseOnOverlayClick`, `shouldCloseOnEscapePress` | no close button
Ant | ✅ | `maskClosable` | `closeIcon=null`
AWS Cloudscape | ✅ | no | unremovable
Adobe Spectrum React | 🚫 | `isDismissible` | `isDismissible`
Github Primer | 🚫 | no | unremovable
MDC2 (vivid 2) | ✅ | `scrimClickAction: "close" \| ""`, `escapeKeyAction: "close" \| ""` | `closeButton`

### Questions

1. Should dismiss-on-scrim be default? **Yes**
- In [VIV-1186](https://jira.vonage.com/browse/VIV-1186) Ayala said it was by design.
- Most other libraries have it as default
- It would be a change in behaviour. Maybe not breaking
2. Should we support the ability to remove close button and dismiss-on-esc? **Yes**
- I know that this is at least used by dashboard in one place

### API

We cannot use `removable`, `dismissible` or `auto-dismiss` as they are the default behaviour.

In Vivid, generally `removable` refers to removing from DOM, `dismiss` refers to closing something. (Except alert which is a mistake). Additionally, `removing` is not a word associated with closing a dialog. So we should stay with `dismiss`.

Let's consider use cases for light dismiss (default), no dismiss and explicit dismiss. 

1. Current
```html
<vwc-dialog>
<vwc-dialog non-dismissible>
<vwc-dialog non-dismissible="scrim">
```

2. Rachel but inverted
```html
<vwc-dialog>
<vwc-dialog non-removable non-dismissible>
<vwc-dialog non-dismissible>
```

3. New idea
```html
<vwc-dialog> <!-- defaults to dismissible="light" -->
<vwc-dialog dismissible="false"> <!-- maybe instead dismissible="non-dismissable" dismissible="blocking" -->
<vwc-dialog dismissible="explicit">
```

4. Separate
```html
<vwc-dialog>
<vwc-dialog no-close-button no-auto-dismiss no-dismiss-on-esc>
<vwc-dialog no-auto-dismiss>
```

Of the three I think 1. is the clearest so I would like to keep the PR as is. I don't think the space separation is an issue because:
- This will be a rare use case
- Space separation is the idiomatic way to provide a list in html, see `class` attribute

Please let me know what you think.

TaylorJ76:
I think the main purpose of this task was so the a _modal_ dialog (opened with `dialog.showModal()`) can behave as per the [WC3 recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/), which means:

- It does not close when the scrim is clicked (light-dismiss)
- ESC key press does close the dialog (I don't think we should disable this in any situation)

You might want to hide the close button if you are providing another button to close the dialog (eg a Cancel button) but I don't see a problem with having both. Or we could provide a separate prop for this purpose (`no-close-button` for eg).

I think I'd rather a `no-light-dismiss` prop which just disables the click on the scrim and an update to the docs outlining the differences between a normal dialog and a modal dialog (from an a11y POV).

rinaok:
> I think the main purpose of this task was so the a _modal_ dialog (opened with `dialog.showModal()`) can behave as per the [WC3 recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/), which means:
> 
> * It does not close when the scrim is clicked (light-dismiss)
> * ESC key press does close the dialog (I don't think we should disable this in any situation)
> 
> You might want to hide the close button if you are providing another button to close the dialog (eg a Cancel button) but I don't see a problem with having both. Or we could provide a separate prop for this purpose (`no-close-button` for eg).
> 
> I think I'd rather a `no-light-dismiss` prop which just disables the click on the scrim and an update to the docs outlining the differences between a normal dialog and a modal dialog (from an a11y POV).

Agree with @TaylorJ76 , 
1. According to the html spec `Modal dialogs can also be closed by pressing the Esc key.` I don't see a reason to disable it.
2. `no-light-dismiss` is a better naming.

Also, the regular modal dialog stopped working: https://vivid-kwej0wcb2jt7.deno.dev/components/dialog#modal-dialog

RichardHelm:
I know that we have a non-dismissible dialog in dashboard. It's used to collect some required data before the user can continue. So we need that functionality.

In that case, close by ESC needs to be disabled as well. 

The APG does not recommend explicit dismiss, it states: 
> in some implementations, attempts to interact with the inert content cause the dialog to close.

So you're proposing the following, which is actually option 2 with different names.
```
<vwc-dialog>
<vwc-dialog no-close-button no-light-dismiss>
<vwc-dialog no-light-dismiss>
```

Should `no-close-button` then also disable ESC and light dismiss?

TaylorJ76:
> I know that we have a non-dismissible dialog in dashboard. It's used to collect some required data before the user can continue. So we need that functionality.
> 
> In that case, close by ESC needs to be disabled as well.
> 
> The APG does not recommend explicit dismiss, it states:
> 
> > in some implementations, attempts to interact with the inert content cause the dialog to close.
> 
> So you're proposing the following, which is actually option 2 with different names.
> 
> ```
> <vwc-dialog>
> <vwc-dialog no-close-button no-light-dismiss>
> <vwc-dialog no-light-dismiss>
> ```
> 
> Should `no-close-button` then also disable ESC and light dismiss?

I think not having any method to close a dialog is an [accessibility concern](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations).

TaylorJ76:
> > I know that we have a non-dismissible dialog in dashboard. It's used to collect some required data before the user can continue. So we need that functionality.
> > In that case, close by ESC needs to be disabled as well.
> > The APG does not recommend explicit dismiss, it states:
> > > in some implementations, attempts to interact with the inert content cause the dialog to close.
> > 
> > 
> > So you're proposing the following, which is actually option 2 with different names.
> > ```
> > <vwc-dialog>
> > <vwc-dialog no-close-button no-light-dismiss>
> > <vwc-dialog no-light-dismiss>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should `no-close-button` then also disable ESC and light dismiss?

No, I think if we were to have a `no-close-button` prop, it would just hide the close button. And I would state in the docs that when set, another method of closing the dialog (a button) should be provided.

I think not having any method to close a dialog is probably an [accessibility issue](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations).

RichardHelm:
> I think not having any method to close a dialog is probably an [accessibility issue](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations).

Sure, but in the case of dashboard there is (by entering the required data). We can highlight that in the docs though.

rachelbt:
> I think the main purpose of this task was so the a _modal_ dialog (opened with `dialog.showModal()`) can behave as per the [WC3 recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/), which means:
> 
> * It does not close when the scrim is clicked (light-dismiss)
> * ESC key press does close the dialog (I don't think we should disable this in any situation)
> 
> You might want to hide the close button if you are providing another button to close the dialog (eg a Cancel button) but I don't see a problem with having both. Or we could provide a separate prop for this purpose (`no-close-button` for eg).
> 
> I think I'd rather a `no-light-dismiss` prop which just disables the click on the scrim and an update to the docs outlining the differences between a normal dialog and a modal dialog (from an a11y POV).

I also agree with @TaylorJ76 , but I don't like the negative properties we are creating (also very long names) :(
I'm aware of the problem we have - that the both `light-dismiss` and the `close-button` are true by default, but I think **we** need to think this over again with the names. 
`<vwc-dialog removable light-dismiss>`
looks much nicer IMO then this:
`<vwc-dialog no-close-button no-light-dismiss>`

RichardHelm:
@rachelbt: Sure it looks nicer, but defaults are also important. Even if we could do this, it would make non-dismissible dialog the default, which is undesirable.

To continue the discussion, it seems we have two options:

1. Current:
```html
<vwc-dialog>
<vwc-dialog non-dismissible>
<vwc-dialog non-dismissible="scrim">
```

2. Seperate (names tbd):
```html
<vwc-dialog>
<vwc-dialog no-x no-y no-z>
<vwc-dialog no-x>
```

Of which all three of you prefer 2. which I am fine with. So let's choose good names for it:

1. close button: `no-close-button`, `no-dismiss-button`
  * the button is referred to as "dismiss button" already (`dismiss-button-aria-label`), so I guess it should be `no-dismiss-button`
2. esc: `no-close-on-esc`, `no-esc-action`
  * I think `no-close-on-esc` is more clear
3. scrim: `no-light-dismiss`, `no-auto-dismiss`, `no-close-on-scrim-click`, `no-close-on-click-outside`
  * I'm happy with `no-light-dismiss`

Please let me know what you think.

RichardHelm:
@TaylorJ76 and I have decided for now to just add the `no-light-dismiss` functionality.

libs/components/src/lib/dialog/README.md
```diff
@@ -90,7 +112,9 @@ The `icon-placement` attribute specifies where the dialog's icon should appear (
 
 ### Open
 
-Sets or returns whether a dialog should be open or not
+Sets or returns whether a dialog should be open or not.
+
+Note: You cannot use the `open` attribute to open a modal dialog. Use `showModal()` instead.
```
rachelbt, line 117: 👍

libs/components/src/lib/dialog/README.md
```diff
@@ -106,6 +130,24 @@ Sets or returns whether a dialog should be open or not
 <vwc-dialog id="dialog" headline="Headline" subtitle="subtitle"></vwc-dialog>
 ```
 
+### No-light-dismiss
```
rachelbt, line 133: I know we had long dissection regarding this, but I can't see any summery about this....
Since the native behaviour is for the dialog **not to be closed** when screen presses, why don't we make this the default behaviour and add - `light-dismiss` if someone needs it. I think this is also what @rinaok had suggessted. I don't find this as a breaking change.
@James226 , @rinaok WDYT?

rachelbt (commented):
left one comment.
Lets finish with this :)

TaylorJ76 (approved):

PR #1460, opened 2023-11-23
chore(selectable-box): use card inside use case (VIV-000)

rachelbt:


libs/components/src/lib/selectable-box/README.md
```diff
@@ -117,33 +117,24 @@ Also, the boxes need a parent element to have a `role` attribute set to `group`.
 
 ## Use cases
 
-### Text based boxes
+### Selectable Card
 
 ```html preview
 <vwc-layout role="group">
   <vwc-selectable-box control-aria-labelledby="box1-heading" style="max-inline-size: 450px">
```
RichardHelm, line 124: Very nice! However it's got double padding now, so I think they should be `tight`. And we can add `clickable` / `clickable-box` to make it nicer

RichardHelm (changes requested):

TaylorJ76 (approved):

PR #1473, opened 2023-11-30
fix(note): adjust accent background color to other connotations (VIV-1397)

rachelbt:


libs/components/src/lib/note/note.scss
```diff
@@ -19,6 +19,10 @@
 	@include connotation.connotation (note);
 	@include appearance.appearance;
 
+	// using field-set as appearance causes accent to look like a field.
+	// if we will be using fieldset to components other then input types - we may consider adding appearance.
+	--vvd-note-accent-backdrop: var(#{constants.$vvd-color-neutral-50});
```
RichardHelm, line 24: Will prevent consumers from overriding the value though.
Maybe something like will do?

```scss
:host {
  --_vvd-note-accent-backdrop: var(--vvd-note-accent-backdrop, #{constants.$vvd-color-neutral-50})
}
```
```suggestion
	--vvd-note-accent-backdrop: var(--_vvd-note-accent-backdrop);
```

RichardHelm (changes requested):

libs/components/src/lib/note/note.scss
rachelbt: Thanks!

rachelbt: > Will prevent consumers from overriding the value though. Maybe something like will do?
> 
> ```css-scss
> :host {
>   --_vvd-note-accent-backdrop: var(--vvd-note-accent-backdrop, #{constants.$vvd-color-neutral-50})
> }
> ```

I added a new appearance after all,
see also this: https://jira.vonage.com/browse/VIV-1441

RichardHelm (approved):

PR #1477, opened 2023-12-04
fix(menu): accessible name with aria-label attribute (VIV-1395)

TaylorJ76:
Adds aria-label to menu.
Also adds Esc key close behaviour.

Adds aria-label to the examples and also the theme switcher menu in docs and fixes it so it can work with keyboard only.


libs/components/src/lib/menu/menu.template.ts
```diff
@@ -21,6 +21,13 @@ const getClasses = ({
 		['hide-body', items && !(items as unknown as HTMLElement[]).length]
 	);
 
+function handleEscapeKey(menu: Menu, event: Event) {
```
rachelbt, line 24: if the menu is open esc should close it?
If yes - it doesn't work :(
we also have auto dismiss, maybe it suppose to be only there?

rachelbt (commented):
all of the examples in menu-item are inside of menu - we should add the aria-label there too

libs/components/src/lib/menu/menu.template.ts
TaylorJ76: You need to be focussed on the menu in order to close the menu with esc.
I will open a story to look at the general keyboard accessibility of menu as there are other issues which as out of the scope of this PR.

apps/docs/assets/scripts/theme-handler.js
```diff
@@ -42,6 +42,11 @@ button.addEventListener('click', toggleMenu);
 const menu = document.querySelector('vwc-menu#dark-mode-menu');
 menu.addEventListener('click', event => setTheme(event.target.text?.toLowerCase()));
 
+menu.addEventListener('keydown', event => {
+    if (event.key === 'Enter' || event.key === 'Space')
+        return setTheme(event.target.text?.toLowerCase());
+});
+
```
RichardHelm, line 49: Should not be necessary to do this
```suggestion
menu.addEventListener('change', event => {
  if (event.target.checked)
    setTheme(event.target.text?.toLowerCase()));
});
```

RichardHelm (changes requested):

apps/docs/assets/scripts/theme-handler.js
```diff
@@ -40,8 +40,10 @@ const button = document.querySelector('vwc-button#dark-mode-toggle');
 button.addEventListener('click', toggleMenu);
 
 const menu = document.querySelector('vwc-menu#dark-mode-menu');
-menu.addEventListener('click', event => setTheme(event.target.text?.toLowerCase()));
-
+menu.addEventListener('change', event => {
+  if (event.target.checked)
+    setTheme(event.target.text?.toLowerCase()));
+});
```
RichardHelm, line 46:
```suggestion
menu.addEventListener('change', event => {
  if (event.target.checked)
    setTheme(event.target.text?.toLowerCase());
});
```

RichardHelm (approved):

PR #1486, opened 2023-12-10
feat(tabs): add gutters attribute (VIV-1422)

rinaok:


rachelbt (commented):
didn't went over the code yet, but isn't this a breaking change? I don't remember what we have decided regarding this one.

rinaok:
> didn't went over the code yet, but isn't this a breaking change? I don't remember what we have decided regarding this one.


@rachelbt We can make a breaking change here since tabs are in alpha.

TaylorJ76:
> > didn't went over the code yet, but isn't this a breaking change? I don't remember what we have decided regarding this one.
> 
> @rachelbt We can make a breaking change here since tabs are in alpha.

@rinaok We can't make this change until the next major release. I've updated the ticket details of what was discussed in backlog grooming

rinaok:
@TaylorJ76 @rachelbt I switched from `tight` to `gutters`. Would it be better to receive "small", "medium", and "large" instead of a boolean attribute, same as in layout https://vivid.deno.dev/components/layout#gutters ? WDYT?

rachelbt:
> @TaylorJ76 @rachelbt I switched from `tight` to `gutters`. Would it be better to receive "small", "medium", and "large" instead of a boolean attribute, same as in layout https://vivid.deno.dev/components/layout#gutters ? WDYT?

can you revert the new png's of the tab? they are not related - and were not suppose to change. 

I like the gutters!
I wonder if we need all the options of: none (default), small, medium, large.
Maybe we should start off with: `none` (default), `small` & `small-block` and if needed we will add more?
@James226 @RichardHelm WDYT?

RichardHelm:
I like the gutters! Maybe we should just have `none` and `medium` for now, we can add more later.

I want to point out 2 things:
* This is a design problem. Why are coming up with this in code and how are designers going to know the right padding to apply to their tabs now?
* We want gutters to become the default in a future release, right? So we should probably tell people that in the docs and should also use gutters=medium in all the examples so that their code becomes "future-proof".

rachelbt:
> I like the gutters! Maybe we should just have `none` and `medium` for now, we can add more later.
> 
> I want to point out 2 things:
> 
> * This is a design problem. Why are coming up with this in code and how are designers going to know the right padding to apply to their tabs now?
> * We want gutters to become the default in a future release, right? So we should probably tell people that in the docs and should also use gutters=medium in all the examples so that their code becomes "future-proof".

why medium? in layout small is 16px;

RichardHelm:
> why medium? in layout small is 16px;

Ok, small then.
I was just thinking medium for the normal amount

rachelbt:
> * This is a design problem. Why are coming up with this in code and how are designers going to know the right padding to apply to their tabs now?

I opened a ticket for it:
https://jira.vonage.com/browse/VIV-1467

rachelbt (commented):
look great!
two little things:
1. can you revert the tab ui tests? it seems like nothing changed there - or safari is missing...
2. though gutters are presented last - I think the tabs look weird without the gutters - maybe we should add it to the code samples? WDYT?

libs/components/src/lib/tabs/README.md
```diff
@@ -10,40 +10,49 @@ The vwc-tabs accepts [vwc-tab](../../components/tab) and `vwc-tab-panel` element
...
-    <vwc-tab label="Desserts" id="desserts"></vwc-tab>
-    <vwc-tab-panel id="appsPanel">
-        <ol>
-            <li>Stuffed artichokes</li>
-            <li>Bruschetta</li>
-            <li>Oven-baked polenta</li>
-            <li>Salami and Fig Crostini with Ricotta</li>
-            <li>Rosemary-Potato Focaccia with Goat Cheese</li>
-        </ol>
+<vwc-tabs>
```
rachelbt, line 13:
```suggestion
<vwc-tabs gutters="small">
```

rinaok: I left the first one without gutters on purpose, without that we don't have an example without gutters..

rachelbt (approved):
LGTM

PR #1488, opened 2023-12-11
fix(menu-item): prevent click event when disabled (VIV-1440)

TaylorJ76:


libs/components/src/lib/menu-item/menu-item.scss
```diff
@@ -24,7 +24,7 @@
 }
 
 :host([disabled]) {
-	cursor: not-allowed;
```
rachelbt, line 27: why did you removed that?
In button we have the not-allowed, I think we should have it here as well

TaylorJ76: Because by adding `pointer-events: none` stops you from being able to control the cursor.

RichardHelm: Can't we stop the event with stop propagation instead?

TaylorJ76: No, I tried that originally, but the event originates on the host so cannot be suppressed from inside the component (if far as I could tell).

RichardHelm (approved):

PR #1496, opened 2023-12-12
fix(date-picker, date-range-picker): fix popup not being anchored (VIV-1439)

RichardHelm:


rachelbt (commented):
wasn't the popup suppose to anchor now to the text-field?
like this:
image
and not like this:
image

RichardHelm:
@rachelbt No, this PR fixes the issue that the popup not actually being anchored at all and not scrolling with the text-field.

I'm not sure if we want to position it like you showed, because it will hide the helper text when open

rachelbt (approved):
LGTM

PR #1498, opened 2023-12-12
fix(text-field): fix input styles getting overridden by css resets (VIV-1433)

RichardHelm:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -52,14 +53,18 @@ $low-ink-color: --_low-ink-color;
...
-	font: var(#{constants.$vvd-typography-base});
-	grid-column: 1 / 4;
-	grid-row: 1;
+	/* As label is in light dom, use !important and initial to prevent values getting overridden by global CSS resets etc. */
+	box-sizing: initial !important;
+	padding: initial !important;
+	margin: initial !important;
+	color: var(#{constants.$vvd-color-canvas-text}) !important;
+	contain: inline-size !important;
+	font: var(#{constants.$vvd-typography-base}) !important;
```
rachelbt, line 62: as we discussed - this may cause an a11y issue:

> From a Web Accessibility perspective, the use of !important is negative because it would [override the end-user defined styling](https://dev.to/alvaromontoro/allow-end-user-styling-overrides-2467). And there are multiple cases in which users –and especially users with disabilities– may want to override some of the CSS properties (e.g. to enlarge the font size, or change the font-family, or even change the distribution and position of elements in a page.)

libs/components/src/lib/text-field/text-field.scss
```diff
@@ -52,14 +53,18 @@ $low-ink-color: --_low-ink-color;
...
+	grid-column: 1 / 4 !important;
+	grid-row: 1 !important;
```
rachelbt, line 64:
```suggestion
	grid-column: 1 / 4;
	grid-row: 1;
```
I don't think this is needed - why would some reset or css variable setting values for grid?

libs/components/src/lib/text-field/text-field.scss
```diff
@@ -52,14 +53,18 @@ $low-ink-color: --_low-ink-color;
...
 
 	.char-count + & {
-		grid-column: 1 / 3;
+		grid-column: 1 / 3 !important;
```
rachelbt, line 67:
```suggestion
		grid-column: 1 / 3;
```
same

libs/components/src/lib/text-field/text-field.scss
```diff
@@ -153,12 +160,12 @@ $low-ink-color: --_low-ink-color;
 
 	&:not(slot):focus-visible,
 	&::slotted(input:focus-visible) {
-		outline: none;
+		outline: none !important;
```
rachelbt, line 163: not sure about this because of a11y.
Lets leave this one out of the important. The worse case - there will be an outline, because something added it - and I guess it was intended to be there.

RichardHelm:
@rachelbt Ok, I removed !important from these

rachelbt (approved):
Don't like the important but I get we have no option :(

PR #1501, opened 2023-12-19
chore(components): use border radius variables (VIV-1461)

rachelbt:
moved all hard-coded values to use border-radius variables.
3 components are left out (VIV-1462)

left border-radius: inherit out of the variables.


RichardHelm (approved):
👍

PR #1503, opened 2023-12-19
fix(banner): fix control being focusable (VIV-1414)

RichardHelm:


rachelbt:
vivid-2 banner (material) is tubbable. 
Do we really don't need the banner to be in the tab keyboard navigation?

TaylorJ76 (approved):

RichardHelm:
@rachelbt Why should a banner be focusable? It's non-interactive

rachelbt:
> @rachelbt Why should a banner be focusable? It's non-interactive

not sure.
pointed out the fact that it was in vivid-2. Not that vivid-2 was accessible but wondered if it was there for a reason

RichardHelm:
It was probably is mistake in Vivid 2 as well. I checked some other sites and their banners are non-focusable as expected

PR #1504, opened 2023-12-19
feat(docs): add edit in github link (VIV-000)

RichardHelm:
Adds a edit link at the bottom of component pages to edit directly in Github:
image

This will make it faster to make tweaks to our docs, and hopefully encourage outside contributions


rachelbt (commented):
why do we need that?
I get that it will help us with editing the docs - but how this will help with contributors?
We are not like MDN that the doc page is the core of our doing.
What do others think? (@James226 @rinaok @YonatanKra )

RichardHelm:
I have added this after seeing it from other design systems like Primer

It will hopefully encourage contributions to the docs, because the link makes it easier and its presence indicates that contributions are welcome

rachelbt:
> I have added this after seeing it from other design systems like Primer
> 
> It will hopefully encourage contributions to the docs, because the link makes it easier and its presence indicates that contributions are welcome

yes, Primer is the only one that I seen doing it.
If all are good with it - then I'm too

rinaok (approved):

PR #1509, opened 2023-12-20
feat(alert): add option for non-fixed alert (VIV-1315)

rachelbt:
Not sure about the property name...


libs/components/src/lib/alert/alert.scss
```diff
@@ -51,6 +51,12 @@ $alert-transition-delay: --transition-delay;
 		margin-inline: 16px;
 	}
 
+	&:where(.in-flow) {
+		position: static;
+		margin: 16px;
```
RichardHelm, line 56: Why does it have a built-in margin?

libs/components/src/lib/alert/README.md
```diff
@@ -125,6 +125,17 @@ Use the `placement` attribute to set the location of the alert.
 <vwc-alert class="small-alert" placement="bottom-end" text="bottom-end" open></vwc-alert>
 ```
 
+### In-flow
```
RichardHelm, line 128: An alternative could be to add a `position` attribute, with "fixed" being the default

rachelbt: I thought about it - was afraid it was confusing with 'placement'...

libs/components/src/lib/alert/alert.scss
rachelbt: 👍  no reason. I'll remove it

libs/components/src/lib/alert/README.md
rinaok: We use an attribute called `strategy` in popup for this, and we use `fixed-dropdown` in select.

RichardHelm: In the future we might want to add more options like `position: absolute`, so maybe `strategy` is good

RichardHelm (approved):

RichardHelm (approved):

PR #1510, opened 2023-12-20
feat(data-grid): add cell-click event (VIV-1443)

RichardHelm:


libs/components/src/lib/data-grid/README.md
```diff
@@ -447,14 +447,44 @@ The default slot, where all the content is rendered.
...
+### Cell-click
+
+The `cell-click` event is fired when a cell is clicked on or when the enter or space key is pressed on a focused cell.
+
+Event details: `{ cell, row, isHeaderCell, columnDataKey }`
+
+```html preview
+
+```html preview
+<vwc-data-grid></vwc-data-grid>
```
rinaok, line 467:
```suggestion
```html preview
<vwc-data-grid></vwc-data-grid>
```

libs/components/src/lib/data-grid/data-grid-cell.ts
```diff
@@ -62,6 +67,48 @@ export class DataGridCell extends FoundationDataGridCell {
 			}
 		};
 	}
+
+	#getColumnDataKey() {
```
rinaok, line 71:
```suggestion
/**
	 * @internal
	 */
	#getColumnDataKey() {
```

libs/components/src/lib/data-grid/data-grid-cell.ts
```diff
@@ -62,6 +67,48 @@ export class DataGridCell extends FoundationDataGridCell {
...
+			});
+		}
+	}
+
+	/**
+	 * @internal
+	 */
+	_handleKeyDown(e: KeyboardEvent) {
+		if (e.target === this && (e.key === keyEnter || e.key === keySpace)) {
+			this._handleClick();
```
rinaok, line 108: Will be better to move it to a separate method instead of calling handeclick

libs/components/src/lib/data-grid/data-grid-cell.ts
RichardHelm: It's not needed with `#`, the methods cannot be called from outside

libs/components/src/lib/data-grid/data-grid-cell.ts
RichardHelm: Sure, done 👍

rinaok (approved):

PR #1514, opened 2023-12-26
fix(tabs): add panel scroll support when height set on tabs (VIV-1468)

rachelbt:


libs/components/src/lib/tabs/README.md
```diff
@@ -137,4 +137,46 @@ Each `tab panel` has an associated [vwc-tab](../../components/tab) element, that
...
+		block-size: 200px;  
+  }
+  .my-panel {block-size: 100%; overflow-y: auto;}
+</style>
+<vwc-tabs class="tabs">
+    <vwc-tab label="Tab one" id="one"></vwc-tab>
+    <vwc-tab label="Tab two" id="two"></vwc-tab>
+    <vwc-tab label="Tab three" id="tree"></vwc-tab>
+    <vwc-tab-panel id="onePanel">
+      <div class="my-panel">
```
RichardHelm, line 160: Won't this be a problem with the new `gutters` attribute?
Perhaps the overflow needs to be on `.tabpanel` of vwc-tabs instead, to include the gutters?

rachelbt: I moved the scroll to the panel, I'm hoping this will not effect users that use inside div with scroll. As they do not use the gutters yey I guess it will be OK - but when we will move gutters as default - the scroll will be inside the panel and padded from the end by 16px.

RichardHelm (approved):

PR #1516, opened 2023-12-28
fix(popup, tooltip, toggletip): fix anchored components not handling being reconnected to DOM (VIV-1412)

RichardHelm:
This PR:

1. Fixes these components breaking when moved in the DOM / reconnected 
2. Also fixes the performance bug of popup by registering autoUpdate only when `open`


rinaok (changes requested):
When you toggle the popup, the text has a glitch: https://vivid-1gdsrnws0ede.deno.dev/components/popup#open

RichardHelm:
@rinaok Yes, I have fixed this now. (Although the issue was not related to my changes, only getting triggered now)

rinaok (approved):

PR #1518, opened 2023-12-28
chore(docs): update main page (VIV-000)

rinaok:


libs/components/README.md
```diff
@@ -1,171 +1,405 @@
...
-- [Dropzone](https://dropzone.dev) - for handling dropped files
-
----
-## Have questions?
-
-Still looking for answers, ask us in [#ask-vivid](https://vonage.slack.com/archives/C013F0YKH99) slack channel.
 
+<vwc-layout column-basis="small" gutters="large">
+ <vwc-card headline="Design guidelines" subtitle="Vivid Design foundations, principles, styles and best practices.">
+<svg slot="media" width="100%" height="394" viewBox="0 0 940 788" fill="none" xmlns="http://www.w3.org/2000/svg">
```
rachelbt, line 103:
```suggestion
<svg slot="media"  viewBox="0 0 940 788" fill="none" xmlns="http://www.w3.org/2000/svg">
```
add css for the SVG with:
inline-size: 100%;
block-size: auto;

rachelbt (commented):
Looks so much better.
added some comments to make it responsive.
We should check if changing the readme has an effect on the preview in git

rinaok:
@rachelbt The github readme is a different one, it's not the same file.

> Looks so much better. added some comments to make it responsive. We should check if changing the readme has an effect on the preview in git

rachelbt (dismissed):
Makes our home page so much nicer!!

libs/components/README.md
```diff
@@ -1,171 +1,405 @@
...
+  }
+  design.onclick = () => {
+    window.location.href = '/designs/guidelines/';
+  }
+  migration.onclick = () => {
+    window.location.href = '/getting-started/vivid-2-migration/';
+  }
+  iconsGallery.onclick = () => {
+    window.location.href = '/icons/icons-gallery/';
+  }
```
RichardHelm, line 404: These should be links. But I understand Vivid doesn't support links as buttons at the moment, so I've created a ticket to investigate: https://jira.vonage.com/browse/VIV-147

RichardHelm (approved):
Looks great! 😍

PR #1520, opened 2024-01-02
chore(components): change 6px to 8px border-radius (VIV-1474)

rachelbt:


libs/components/src/lib/checkbox/ui.test.ts-snapshots/snapshots-checkbox-Desktop-Chromium-linux.png
RichardHelm, line 1: This doesn't seem right, only the focus ring changed
image

RichardHelm (changes requested):

libs/components/src/lib/checkbox/ui.test.ts-snapshots/snapshots-checkbox-Desktop-Chromium-linux.png
rachelbt: the fact that the focus ring changed is wrong. I probably didn't changed it when I changed the checkbox from 6 to 4px.
Now both checkbox + its focus is 4px.

PR #1523, opened 2024-01-07
feat(split-button): icon slot (VIV-1442)

rachelbt:


libs/components/src/lib/split-button/README.md
```diff
@@ -180,6 +180,19 @@ Add the `disabled` attribute to disable the split button.
...
+### Icon
+
+Set the `icon` slot to show an icon.  
+If set, the `icon` attribute is ignored.  
+Note: Icon, by its own, doesn't make a discernible text. An `aria-label` must be provided to ensure that the user can understand the split button's purpose.
+
+
+```html preview
+<vwc-split-button appearance='outlined' label='submit'>
+		<vwc-icon slot="icon" name="check-circle-solid" connotation="success"></vwc-icon>
```
TaylorJ76, line 192: Should we add an `aria-label` here, like we advise in the text above?

TaylorJ76 (changes requested):

libs/components/src/lib/split-button/README.md
rachelbt: yes, you are right! I'll check if we have it in the icon example as well

rachelbt: maybe the text needs to be more accurate - if only an icon is set in the button...

rachelbt: changed the text :)

TaylorJ76 (approved):

PR #1528, opened 2024-01-10
chore(focus, fab): creat focus mixin and use it in fab (VIV-1492)

rachelbt:


libs/components/src/lib/fab/fab.scss
RichardHelm, line 1: I think we can remove `position: relative;` as well, we always needed that for the focus

RichardHelm (changes requested):
Awesome 👍

Issue #1511, opened 2023-12-20
[ALERT]: Alert gets clipped on smaller screens


ChristianSama:
### Description

The alert component does not resize appropriately to address smaller screens. I would be happy to help by opening a PR for this fix.

### Component(s) impacted

Alert component
![Screenshot from 2023-12-20 16-37-02](https://github.com/Vonage/vivid-3/assets/43491324/938bb1ed-62a4-457c-b131-83a9352d2ca6)

### Browser

Chrome

### Vivid Web Components version

v3.43.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

Personal project

### CodeSandbox example

https://codepen.io/christiansama/pen/xxBKQrV

### Steps to reproduce the issue (if applicable)

1. Use the alert component with a long enough text and an action item.
2. Reduce the width of the viewport until the component gets clipped.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/vonage/vivid-3/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/vonage/vivid-3/issues) for duplicate issues

rachelbt:
Hi @ChristianSama , you are more then welcome to do so.
You can fine help here: [CONTRIBUTING.md](https://github.com/Vonage/vivid-3/blob/main/.github/CONTRIBUTING.md)
Thanks you!

ChristianSama:
I noticed the `--alert-min-inline-size` css variable takes care of this. Setting it to `auto` makes the component resize appropriately. Quick question though, would it make sense to make `auto` the default value? Currently the default is `420px`, but by setting it to auto it would responsive by default. WDYT?

rachelbt:
basically it make sense, but there are 2 issues with it:
1. people who already use the component are expecting it to be 420px - changing it may break the design.
2. if there's a long text - the auto causes the alert to push the margin:
![image](https://github.com/Vonage/vivid-3/assets/7512792/8bd13096-c882-4732-99ad-b5cb493c29a0)

I think the fix here needs to be by setting style for mobile using media-query for `max-width: 768px;

rachelbt:
Hi @ChristianSama we are very happy for your contribution!
However - opening a PR like you did is not the proper way.
We kindly ask you to follow our contribution page:
[...]

PR #1529, opened 2024-01-11
fix(alert): add media query to set min-inline-size to auto (VIV-1564)

ChristianSama:
This change is meant to address https://github.com/Vonage/vivid-3/issues/1511 by adding a media query to set the `min-inline-size` to auto when the viewport is smaller than 768px. It still takes the `$alert-min-inline-size` in case the user wants to override this value.
image


rachelbt:
https://github.com/Vonage/vivid-3/issues/1511#issuecomment-1886780120

YonatanKra:
Hi @ChristianSama. 

I am Sorry it has taken so long to get back to you. I've been away for a few months.

Your solution looks solid and I'll be happy to integrate it. Would you like to add the small screen size use-case to our UI tests to make it a wrap for this PR?

rachelbt (approved):
LGTM

TaylorJ76 (approved):

PR #1539, opened 2024-01-16
feat(time-picker): add component (VIV-1321)

RichardHelm:
Alpha docs are now hidden. It can be found at `/components/time-picker`

The component still has 2 problems:
- The picker doesn't implement the design, because the design doesn't make sense to me
- UX problem when using 12h clock in combination with min. Requires the user to choose PM first which closes the picker. I have not dealt with as I'm waiting for the design to be confirmed.


libs/components/src/lib/time-picker/time-picker.scss
```diff
@@ -0,0 +1,123 @@
...
+		display: flex;
+		flex-shrink: 0;
+		align-items: center;
+		justify-content: center;
+		background-color: var(#{appearance.get-appearance-token(fill)});
+		block-size: $item-block-size;
+		border-radius: #{border-radius-variables.$border-radius-semi-condensed};
+		color: var(#{appearance.get-appearance-token(text)});
+		cursor: pointer;
+		font-size: 14px;
```
rachelbt, line 120:
```suggestion
		font: var(#{constants.$vvd-typography-base});
```

libs/components/src/lib/time-picker/README.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+### Clock
+
+The time picker will display the time in 12h or 24h format depending on the configured locale.
+
+Use the `clock` attribute to override this behavior.
+
+- Type: `'12h'` | `'24h'` | `undefined`
+- Default: `undefined`
```
rachelbt, line 163:
```suggestion
- Default: `12h`
```
although it is undefined - it does have a default - and its 12h

libs/components/src/lib/time-picker/time-picker.scss
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	display: flex;
+	overflow: hidden;
+	padding: $item-gap;
+	block-size: space-for-items($num-items-visible);
+	gap: $item-gap;
+
+	.picker {
+		position: relative;
```
rachelbt, line 61: not sure this one is needed, is it for the :after?

libs/components/src/lib/time-picker/time-picker.scss
```diff
@@ -0,0 +1,123 @@
...
+			@include focus.focus;
+		}
+	}
+
+
+	.item {
+		@include connotation.connotation (time-picker);
+		@include appearance.appearance;
+
+		position: relative;
```
rachelbt, line 110: not sure this is needed :)

rachelbt (commented):
Great work! 👍 
went over only the readme and the scss.

about the scroll:
1. I think we shouldn't present the scroll if there isn't one - in the am-pm there's no reason for it to be.
2. I understand the reason to style the scroll - but this will make this component the only one with it - for win users it might seem strange.
3. also - the scroll on-top of the area causes the option to look a bit strange:


It was nice if we could use the option component inside rather then write all from scratch. maybe its worth adding `size` to option so we can use it here?

libs/components/src/lib/time-picker/README.md
RichardHelm: Actually it's dependent on locale. But we can write that

libs/components/src/lib/time-picker/time-picker.scss
RichardHelm: This one is for scroll offset calculation

RichardHelm:
@rachelbt 
> I think we shouldn't present the scroll if there isn't one - in the am-pm there's no reason for it to be.

It's for alignment, it will look weird otherwise.
image
Other libraries have the same behaviour: MUI and Ant

> I understand the reason to style the scroll - but this will make this component the only one with it - for win users it might seem strange.
> also - the scroll on-top of the area causes the option to look a bit strange:

I have removed it.

> It was nice if we could use the option component inside rather then write all from scratch. maybe its worth adding size to option so we can use it here?

Maybe, but that is a question for design. As I'm currently not following the design at all, I will not touch anything until it is resolved

TaylorJ76 (dismissed):

TaylorJ76 (approved):

TaylorJ76 (approved):

PR #1540, opened 2024-01-17
chore(fieldset appearance): remove backdrop usage (VIV-1441)

rachelbt:


libs/components/src/lib/date-range-picker/ui.test.ts-snapshots/snapshots-date-range-picker-Desktop-Chromium-linux.png
RichardHelm, line 1: Why did this change?

rachelbt: because the hover border color slightly changed

RichardHelm (approved):

PR #1541, opened 2024-01-17
fix(select, checkbox): error & success style (VIV-1458)

rachelbt:
what was done?
1. styled select ghost error + success as it is in Figma:
image
2. added style for checkbox error + success - need to be approved by designers
image


libs/components/src/lib/select/select.scss
RichardHelm, line 1: Can we add a screenshot for this?

RichardHelm (changes requested):
Please see comment

libs/components/src/lib/select/select.scss
rachelbt: you mean for the ghost select states?
I guess we need to.

RichardHelm (approved):

rachelbt:
@RichardHelm  - changed  the code for checkbox according to Ayala's request of design

RichardHelm (approved):

PR #1543, opened 2024-01-18
fix(date-picker, date-range-picker): add event for click on clear (VIV-1478)

rachelbt:


RichardHelm (dismissed):

RichardHelm (approved):

libs/wrapper-gen/componentsDef.ts
```diff
@@ -83,6 +83,11 @@ ComponentRegister.registerComponent('date-picker', component => {
     eventName: 'input',
     valueMapping: '(event.target as HTMLInputElement).value',
   });
+	addVueModel(component, {
+		attributeName: 'value',
+		eventName: 'clear-click',
+		valueMapping: '(event.target as HTMLInputElement).value',
+	});
```
RichardHelm, line 90: This doesn't seem right: the addVueModel would only be needed make the component support [v-model](https://vuejs.org/guide/essentials/forms.html#form-input-bindings)

For the new even there should be no need to change anything.

```suggestion
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1545, opened 2024-01-19
feat(components): minify css (VIV-1376)

RichardHelm:
As css is part included in the js bundle, it cannot be optimised later.

This change reduces bundle size of an empty Vue app from 748.78 kB -> 690.13 kB


rachelbt (approved):
👍 NICE :) 
LGTM

PR #1546, opened 2024-01-19
chore(styles): disable source maps (VIV-1476)

RichardHelm:
We have css source maps enabled but do ship the source maps with the library, which can create warnings.

We can simply remove them as they are not needed.


rachelbt (approved):
LGTM

PR #1553, opened 2024-01-25
chore(select, menu-item, nav-disclosure): refactor focus (VIV-1519)

rachelbt:


libs/components/src/lib/menu-item/ui.test.ts-snapshots/snapshots-menu-item-Desktop-Firefox-linux.png
RichardHelm, line 1: What happened to this?

rachelbt: I added transparent to the box-shadow so the focus now is like it should be and not 3px

RichardHelm: But why does it look like this in Firefox?
image

libs/components/src/lib/menu-item/menu-item.scss
```diff
@@ -16,12 +16,12 @@
...
+@use "../../../../shared/src/lib/sass/mixins/focus" as focus-variables;
 
 
-@supports selector(:focus-visible) {
-	:host(:focus-visible) {
-		outline: none;
-	}
+:host(:focus-visible) {
+	@include focus.focus;
 }
```
RichardHelm, line 25: I think this is not right, outline: none turned into focus style

RichardHelm (changes requested):

RichardHelm (approved):

PR #1555, opened 2024-01-29
feat(tabs): add tab scroll panel as property (VIV-1528)

rachelbt:
This change is due to the breaking change that was caused by this issue:
https://github.com/Vonage/vivid-3/pull/1514
The thing is - that this is creating a breaking change as well - not sure what we are suppose to do :( 

The vertical addition - its not a breaking change and the change that was made in 1514 did not change the vertical behaviour:
https://codepen.io/Rachel-Tannenbaum/pen/yLwpEQw


libs/components/src/lib/tabs/tabs.scss
```diff
@@ -101,8 +101,15 @@
 	// this is for making overflow inside grid to work
 	min-block-size: 0;
 	min-inline-size: 0;
-	overflow-x: hidden;
-	overflow-y: auto;
+
+	.base.scroll & {
+		overflow-x: hidden;
```
RichardHelm, line 106: Shouldn't this be auto as well? 🤔

libs/components/src/lib/tabs/tabs.spec.ts
```diff
@@ -263,6 +263,15 @@ describe('vwc-tabs', () => {
 		});
 	});
 
+	describe('scroll', () => {
+		it('should set class "scroll" is panel-scroll is set', async () => {
```
RichardHelm, line 267:
```suggestion
		it('should set class "scroll" if panel-scroll is set', async () => {
```

libs/components/src/lib/tabs/README.md
```diff
@@ -124,32 +124,99 @@ Add an `activeid` attribute of the active tab.
...
+Each `tab panel` has an associated [vwc-tab](../../components/tab) element, that when activated, displays the `tab panel`.
 
-</div>
 
-## Use Case
-Tab panel with scroll.  
-Add `block-size` to `vwc-tabs` (see code bellow).
+### CSS Variable
+This variable only effect `orientation="vertical`.  
+For scroll only in the panel - set a value for the `--tabs-vertical-max-height` variable.
```
RichardHelm, line 213: I don't understand the purpose of this variable.
In the example there is no difference between using height and the variable.


When I make it smaller than the natural height of the tabs, I get this? Why would anyone want this?
image

libs/components/src/lib/tabs/tabs.ts
```diff
@@ -39,6 +39,13 @@ export class Tabs extends FoundationTabs {
 	 */
 	@attr gutters?: Gutters;
 
+	/**
+	 * sets overflow to the tab-panel
+	 *
+	 * @public
+	 */
+	@attr ({mode: 'boolean',attribute: 'panel-scroll', }) panelScroll = false;
```
RichardHelm, line 47: Mhm, maybe there can be a better name for this:
* Because it controls overflow, it could be called `panel-overflow="auto"`?
* Or panel-scrollable

RichardHelm (changes requested):
I think this is the right way to go, even though it should be the default

libs/components/src/lib/tabs/tabs.ts
rachelbt: whar do you think about: `scrollable-panel`

libs/components/src/lib/tabs/README.md
rachelbt: I started writing an example for a use case for it - and as I write it - I come to understand that it is useless. I'll remove it. Thanks

libs/components/src/lib/tabs/tabs.scss
rachelbt: not sure that horizontal & vertical scroll are good us, on the other side - hiding content is not a good UX too....

libs/components/src/lib/tabs/tabs.ts
RichardHelm: @rachelbt Sure, sounds good!

PR #1559, opened 2024-02-01
fix(button): improve button accessibility (VIV-1505)

RichardHelm:
Most buttons are announced as button now:

* `<vwc-button label="Click me"></vwc-button>` -> "Click me, button"
* `<vwc-button icon="icon" label="Click me"></vwc-button>` -> "Click me, button"
* `<vwc-button icon="icon" title="Click me"></vwc-button>` -> "Click me, button"

However, when using icon button with `aria-label` it is still announced as group:

* `<vwc-button icon="icon" aria-label="Click me"></vwc-button>` -> "Click me, group"

I don't think this can be easily fixed due to how our button works.

### DIFF
--- a/libs/components/src/lib/button/button.template.ts
+++ b/libs/components/src/lib/button/button.template.ts
@@ -4,7 +4,7 @@ import type { ElementDefinitionContext, FoundationElementDefinition } from '@mic
 import { classNames } from '@microsoft/fast-web-utilities';
 import { ProgressRing } from '../progress-ring/progress-ring';
 import { Size } from '../enums';
-import { affixIconTemplateFactory, IconWrapper } from '../../shared/patterns/affix';
+import { affixIconTemplateFactory, IconAriaHidden, IconWrapper } from '../../shared/patterns/affix';
 import type { Button, ButtonAppearance, ButtonSize } from './button';
 
 
@@ -44,7 +44,7 @@ function renderIconOrPending(
 
 	} else {
 		const affixIconTemplate = affixIconTemplateFactory(context);
-		return affixIconTemplate(icon, IconWrapper.Slot);
+		return affixIconTemplate(icon, IconWrapper.Slot, IconAriaHidden.Hidden);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


libs/components/src/lib/button/button.template.ts
```diff
@@ -60,39 +60,41 @@ export const buttonTemplate: (
...
+				aria-label="${(x) => x.ariaLabel}"
+				aria-live="${(x) => x.ariaLive}"
+				aria-pressed="${(x) => x.ariaPressed}"
+				aria-relevant="${(x) => x.ariaRelevant}"
+				aria-roledescription="${(x) => x.ariaRoledescription}"
+				title="${x => x.title}"
+				${ref('control')}
+			>
+				${x => renderIconOrPending(context, x.icon, x.pending, x.size)}
+				${when(x => x.label, html`<span class="text" role="presentation">${(x) => x.label}</span>`)}
```
rachelbt, line 96: role="presentation" - why is this needed?
simple span do not have a role anyway.

rachelbt (commented):
nice!
I don't think it is such an accessibility issue the group with when there's only icon.
I like the aria-hidden that you added - I guess we need to add this fix to split vbutton as well and to nav-disclosure too.

libs/components/src/lib/button/button.template.ts
RichardHelm: Without it, it still gets announced as group in VoiceOver. Maybe that is an issue in VoiceOver.

rachelbt: I wonder then why adding the role presentation to the host didn't solved the group issue with the only icon. But I think we can leave it.

rachelbt (approved):
LGTM
👍

libs/components/src/lib/button/button.template.ts
RichardHelm: That one is actually working as intended, unfortunately: https://www.w3.org/WAI/ARIA/apg/practices/hiding-semantics/#presentation_role_ignored

PR #1564, opened 2024-02-05
fix(text-field): fix focus and blur events bubbling (VIV-1538)

RichardHelm:


rachelbt (approved):
LGTM,
also data-grid issue with focus is fine now!

PR #1567, opened 2024-02-06
feat(tooltip, toggletip, menu): add anchor slot (VIV-1539)

RichardHelm:
Sets `display: contents` only when using the slot to maintain backwards compatibility


libs/components/src/lib/menu/README.md
```diff
@@ -8,10 +8,20 @@ Menu, commonly known as a "context menu", is an element that is displayed upon u
 </script>
 ```
 
+## Anchoring
```
rachelbt, line 11: I'm not sure this needs to be on the top.
Isn't it enough as a slot?

libs/components/src/lib/popup/partials/variables.scss
```diff
@@ -1 +1,3 @@
 $popup-width: --_popup-width;
+$popup-display: --popup-display;
+$popup-z-index: --popup-z-index;
```
rachelbt, line 3: these too are mainly for our internal use - right?
if so - I think they need to be with low-dash as well:
$popup-display: --_popup-display;
$popup-z-index: --_popup-z-index;

rachelbt (commented):
NICE!
I don't see a reason for the anchored to be in the top of the doc page.
More over - I think that all that info needs to be under slot- anchor, and maybe we should add about the anchor as a property that this is not the best way to anchor and send people to use the slot (this way we can one day deprecate that).
Not sure about the image that changes - seems not relevant

libs/components/src/lib/menu/README.md
```diff
@@ -115,6 +80,40 @@ Use the `placement` attribute to control the position of the menu relative to it
 </script>
 ```
 
+### Anchor
+
+As an alternative to using the `anchor` slot, you can use the `anchor` member to set the anchor.
```
rachelbt, line 85: its better to use the slot - right?
So why won't we say something like:
It is best to use the anchor slot (link) but you can also....

rachelbt (commented):
left a comment for menu.- but its for all the others as well :)

libs/components/src/shared/patterns/anchored.scss
```diff
@@ -0,0 +1,9 @@
+@use "../../lib/popup/partials/variables" as variables;
+
+:host {
+	&[slotted-anchor] {
+		#{variables.$popup-display}: contents;
+	}
```
rachelbt, line 6: this is not working...
this needs to be : &([slotted-anchor]) {...}
but this throws error.
maybe this is better:
```
:host {
	display: var(#{variables.$popup-display}, inline);
}

:host([slotted-anchor]) {
	#{variables.$popup-display}: contents;
}

```

RichardHelm: Good catch 👍

rachelbt (approved):
NICE!

PR #1569, opened 2024-02-06
chore(docs): code blocks should fit content (VIV-1546)

rinaok:
Instead of this:
Screenshot 2024-02-06 at 17 43 26

Now the code blocks fit content:
Screenshot 2024-02-06 at 17 43 38


apps/docs/assets/styles/main.css
```diff
@@ -174,25 +174,25 @@ article > .cbd-container {
 @media (min-width: 960px) {
```
rachelbt, line 174:
```suggestion
@media (min-width: 1200px) {
```

apps/docs/assets/styles/main.css
```diff
@@ -221,14 +221,20 @@ article.article > h1 {
...
+  text-align: center;
+  background-image: linear-gradient(to left,rgba(255,0,0,0), var(--vvd-color-cta-50), rgba(255,0,0,0));
+}
+
 .article {
   padding-inline: 16px;
   word-break: break-word;
 }
 
 @media (min-width: 960px) {
```
rachelbt, line 234:
```suggestion
@media (min-width: 1200px) {
```

apps/docs/assets/styles/main.css
```diff
@@ -247,11 +253,12 @@ article.article > h1 {
   display: flex;
   flex-direction: column;
   gap: 16px;
+  max-inline-size: 1000px;
 }
 
 @media (min-width: 960px) {
```
rachelbt, line 259:
```suggestion
@media (min-width: 1200px) {
```

apps/docs/assets/styles/main.css
```diff
@@ -274,6 +281,7 @@ article.article > h1 {
   overflow: hidden;
   border: none;
   width: 100%;
+  height: 30px;
```
rachelbt, line 284:
```suggestion
  block-size: 30px;
```

rachelbt (commented):
nice change!! 🦾 

though I think it looks better in light mode then in dark mode, the cta looks a bit off there. 

Can we make a custom heights to out code blocks - like small- 30px, medium- 200px, and large: 460px (maybe we will need more in-between) and remove all the html { /*for demo purposes */}

rinaok:
> nice change!! 🦾
> 
> though I think it looks better in light mode then in dark mode, the cta looks a bit off there.
> 
> Can we make a custom heights to out code blocks - like small- 30px, medium- 200px, and large: 460px (maybe we will need more in-between) and remove all the html { /*for demo purposes */}

It's a good idea, I'll do it.

rinaok:
> > nice change!! 🦾
> > though I think it looks better in light mode then in dark mode, the cta looks a bit off there.
> > Can we make a custom heights to out code blocks - like small- 30px, medium- 200px, and large: 460px (maybe we will need more in-between) and remove all the html { /*for demo purposes */}
> 
> It's a good idea, I'll do it.

@rachelbt Done. Due to too many different sizes, I moved the block-size to the html preview instead of predefined sizes. What do you think? Now the examples look much cleaner.

rachelbt:
> @rachelbt Done. Due to too many different sizes, I moved the block-size to the html preview instead of predefined sizes. What do you think? Now the examples look much cleaner.

Looks so much better!
- when there's no value set in the preview the style should be: `block-size: auto; `right now its is set as `block-size: `
- I still think we can get all the values to be aligned to predefined heights. I went over it we have: 
100, 150, 170, 200, 230, 250, 300, 350, 400, 420, 460, 560.
I think we can narrow it down to:
100, 150, 200, 250, 300, 350, 400, 450, 500, 600
can we set it to match: 1. 1.5, 2, 2.5...?

rinaok:
> > @rachelbt Done. Due to too many different sizes, I moved the block-size to the html preview instead of predefined sizes. What do you think? Now the examples look much cleaner.
> 
> Looks so much better!
> 
> * when there's no value set in the preview the style should be: `block-size: auto; `right now its is set as `block-size: `
> * I still think we can get all the values to be aligned to predefined heights. I went over it we have:
>   100, 150, 170, 200, 230, 250, 300, 350, 400, 420, 460, 560.
>   I think we can narrow it down to:
>   100, 150, 200, 250, 300, 350, 400, 450, 500, 600
>   can we set it to match: 1. 1.5, 2, 2.5...?

Why do you want it to be predefined? It can now be made to fit exactly the size you need. There are too many different sizes, I don't think it's necessary, it will only complicate our work.

Updated the block-size: auto.

AyalaBu:
It looks better now
Thank you

rachelbt (approved):
🦾

PR #1573, opened 2024-02-08
chore(tree-item, tab, tag): refactor focus (VIV-1548)

rachelbt:


libs/components/src/lib/tree-item/tree-item.scss
```diff
@@ -72,6 +66,12 @@
 		pointer-events: none;
 	}
 
+	:host(:focus-visible) & {
+		#{focus-variables.$focus-stroke-gap-color}: transparent;
+
```
RichardHelm, line 71: Selected tree item looks the same with and without focus because of this

RichardHelm (changes requested):

libs/components/src/lib/tree-item/tree-item.scss
rachelbt: tried to fixe the none selected and destroyed the selected :(
Fixed that now!

RichardHelm (approved):

PR #1576, opened 2024-02-11
chore(dialog, data-grid, file-picker, date-picker ): refactor focus (VIV-1555)

rachelbt:


libs/components/src/lib/data-grid/data-grid-cell.scss
```diff
@@ -39,6 +39,12 @@
 	color: var(#{appearance.get-appearance-token(text)});
 	font: var(#{constants.$vvd-typography-base});
 
+	:host(:focus-visible) & {
```
RichardHelm, line 42: This has changed from `:focus` to `:focus-visible`. On purpose?

RichardHelm (changes requested):

libs/components/src/lib/data-grid/data-grid-cell.scss
rachelbt: it was not suppose to be :focus from the first place. :focus not suppose to be styled - only :focus-visible.

RichardHelm (approved):

PR #1579, opened 2024-02-12
chore(wrapper-gen): wrap all exported components (VIV-1553)

RichardHelm:
Generate wrappers for all exported components instead of having to list them in componentsDef


rachelbt (approved):
LGTM

PR #1580, opened 2024-02-13
fix(alert): add icon slot (VIV-1515)

rachelbt:


libs/components/src/lib/alert/alert.template.ts
```diff
@@ -25,7 +25,7 @@ const getControlClasses = ({ open, placement, strategy }: Alert) => classNames(
 function renderIcon(context: ElementDefinitionContext) {
 	const affixIconTemplate = affixIconTemplateFactory(context);
 
-	return html`${x => affixIconTemplate(x.conditionedIcon)}`;
+	return html`${x => affixIconTemplate(x.conditionedIcon || x.icon, IconWrapper.Slot)}`;
```
RichardHelm, line 28: Should already be handled by conditionedIcon getter:
```suggestion
	return html`${x => affixIconTemplate(x.conditionedIcon, IconWrapper.Slot)}`;
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1584, opened 2024-02-16
feat(range-slider): add component (VIV-1488)

RichardHelm:


rinaok (dismissed):
The navigation does not contain the component.

rachelbt:
> The navigation does not contain the component.

this is because we decided to have one or 2 teams checking the component before we add it to the "public" :)

libs/components/src/lib/range-slider/README.md
```diff
@@ -0,0 +1,212 @@
...
+		end.innerText = slider.end;
+	}
+	customElements.whenDefined('vwc-range-slider').then(updateDescription);
+
+	slider.addEventListener('change', updateDescription);
+</script>
+```
+
+### Orientation
+
```
rachelbt, line 100: when using keyboard navigation the up-down arrows are acting strange.
top arrow is down and down is up :stuck_out_tongue_winking_eye:

rachelbt (changes requested):
looks good!
left 2 comments

libs/components/src/lib/range-slider/range-slider.scss
```diff
@@ -0,0 +1,7 @@
+@import "../slider/slider";
+
+.track-start {
+	// Unlike slider, track-start should not be aligned to the start of the track
+	top: unset !important;
+	left: unset !important;
```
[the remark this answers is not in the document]
RichardHelm: Sure, done 👍

libs/components/src/lib/range-slider/README.md
RichardHelm: 🙃

rachelbt:
> > The navigation does not contain the component.
> 
> this is because we decided to have one or 2 teams checking the component before we add it to the "public" :)

but you can see it with adding the component name:
https://vivid-fd8y50sq3b1b.deno.dev/components/range-slider

libs/components/src/lib/slider/slider.scss
```diff
@@ -40,6 +40,8 @@ $track-thickness: 2px;
 
 	#{variables.$thumb-size}: 12px;
 	#{variables.$thumb-interaction-indicator-size}: 36px;
+	#{variables.$track-start-left}: 0;
+	#{variables.$track-start-top}: 0;
```
rachelbt, line 44: sorry for nagging - but maybe if we are allready adding variables lets change it to logical properties of inset-start & end 🙏🏻

TaylorJ76 (changes requested):
Looking good!
Only thing I noticed was that the screen reader (voiceover on mac) was not reading out the minus values correctly.
Eg: "-4" is being read as "4". I can see that the code is correct, but I wonder if we need to investigate why.

RichardHelm:
@TaylorJ76 Well spotted, this seems to be an issue with VoiceOver: https://discussions.apple.com/thread/254773710?sortBy=best

It also affects other components like slider, number-field etc. so I think a possible fix / workaround is out of scope for this PR

TaylorJ76 (approved):

PR #1590, opened 2024-02-20
feat(button): add link option (VIV-1544)

rinaok:


libs/components/src/lib/button/README.md
```diff
@@ -172,6 +172,22 @@ The indicator is not displayed when using the `super-condensed` size.
 <vwc-button appearance="outlined" icon="check-line" label="outlined with icon" pending></vwc-button>
 ```
 
+### Href
+
+Use the `href` attribute to convert the button to a link.
+In addition, all of the native attributes of [`<a>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) are supported, included `target`.
```
RichardHelm, line 178:
```suggestion
In addition, all of the native attributes of [`<a>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) are supported, including `target`.
```

libs/components/src/lib/button/button.template.ts
```diff
@@ -61,40 +130,7 @@ export const buttonTemplate: (
...
-				title="${x => x.title}"
-				${ref('control')}
-			>
-				${x => renderIconOrPending(context, x.icon, x.pending, x.size)}
-				${when(x => x.label, html`<span class="text" role="presentation">${(x) => x.label}</span>`)}
-			</button>
-		</template>
-`;
+			${when(x => !x.href, html<Button>`${renderButtonContent(context)}`)}
+			${when(x => x.href && x.href.length > 0, html<Button>`${renderAnchorContent(context)}`)}
```
RichardHelm, line 134: I think the conditions should be exact opposites so that one is always true
```suggestion
			${when(x => !x.href, html<Button>`${renderButtonContent(context)}`)}
			${when(x => x.href, html<Button>`${renderAnchorContent(context)}`)}
```

libs/components/src/lib/button/button.spec.ts
```diff
@@ -184,6 +184,82 @@ describe('vwc-button', () => {
 
 	});
 
+	describe('link', () => {
```
RichardHelm, line 187: IMO there should be one test for href here, which checks that when href is set, the control turns into an anchor tag

```suggestion
	describe('href', () => {
```

libs/components/src/lib/button/button.spec.ts
```diff
@@ -184,6 +184,82 @@ describe('vwc-button', () => {
...
+			expect(getAnchorElement()?.getAttribute(attribute)).toEqual(text);
+		});
+
+		it('should set the "hreflang" attribute', async function () {			
+			const attribute = 'hreflang';
+			const text = 'stam';
+			await setAttribute(element, attribute, text);
+
+			expect(getAnchorElement()?.getAttribute(attribute)).toEqual(text);
+		});
```
RichardHelm, line 212: Then all the other tests are exactly the same except for data, so a good candidate for parametrised tests (describe.each)

libs/components/src/lib/button/button.spec.ts
```diff
@@ -184,6 +184,82 @@ describe('vwc-button', () => {
...
+		
+		beforeEach(async () => {	
+			element.href = '/somewhere';
+			await elementUpdated(element);
+		});
+
+		it('should set the "href" attribute', async function () {
+			const attribute = 'href';
+			const text = 'link';
+			await setAttribute(element, attribute, text);
```
RichardHelm, line 201: Unrelated, but `setAttribute` does not actually set an attribute so it is terribly named 😱

docs/index.md
```diff
@@ -438,22 +438,4 @@ This project is licensed under the Apache 2.0 License - see the [LICENSE.md](htt
...
-  design.onclick = () => {
-    window.location.href = '/designs/guidelines/';
-  }
-  migration.onclick = () => {
-    window.location.href = '/guides/vivid-2-migration/';
-  }
-  iconsGallery.onclick = () => {
-    window.location.href = '/icons/icons-gallery/';
-  }
-</script>
```
RichardHelm, line 459: 👍

RichardHelm (changes requested):

libs/components/src/lib/button/button.spec.ts
```diff
@@ -184,6 +184,21 @@ describe('vwc-button', () => {
...
 	});
 
+
+	describe.each(['href', 'hreflang', 'download', 'ping', 'referrerpolicy', 'rel', 'target', 'type'])('%s attribute', (attribute) => {
+		beforeEach(async () => {
+			element.href = '/somewhere';
+			await elementUpdated(element);
+		});
+
+		it('should be fired when a user toggles the checkbox', async () => {
```
RichardHelm, line 194:
```suggestion
		it('should be forwarded to the anchor element', async () => {
```

RichardHelm (changes requested):

RichardHelm (approved):

PR #1592, opened 2024-02-20
chore(wrapper-gen): use information from CEM (VIV-1578)

RichardHelm:
* All components are marked with `@component`
* All public components are marked with `@public`
* All v-model are declared via `@vueModel`


libs/components/src/lib/appearance-ui/appearance-ui.ts
```diff
@@ -22,9 +22,7 @@ Appearance.Filled
 
 
 /**
- * Base class for appearance-ui
- *
- * @public
+ * @component appearance-ui
```
rachelbt, line 25: not sure we need this - this is not a component to use :)

libs/components/src/lib/data-grid/data-grid.ts
```diff
@@ -24,9 +24,8 @@ export type ValueOf<T> = T[keyof T];
 export type DataGridSelectionMode = ValueOf<typeof DataGridSelectionMode>;
 
 /**
- * Base class for data-grid
- *
  * @public
+ * @component data-grid
  * @slot - Default slot.
  * @event cell-click - Event that fires when a cell is clicked
```
rachelbt, line 30: Are this is needed for Vue as well?
```
ComponentRegister.addComponentOverride('data-grid', component => {
  component.attributes.push({
    name: 'rowsData',
    description: 'Array of objects representing the rows of the grid.',
    type: [{ text: 'any[]', vuePropType: 'Array' }],
  });
});
```

rachelbt (commented):
are components that are under lying such as popup, elevation, listbox still need the public @component?
You are missing - range slider

libs/components/src/lib/appearance-ui/appearance-ui.ts
RichardHelm: Yes, the non-public components are hidden by not having `@public`

TaylorJ76 (approved):

PR #1601, opened 2024-02-23
feat(docs): speed up navigation through turbolinks (VIV-1585)

RichardHelm:


rachelbt (dismissed):
nice addition to the site 👍

rachelbt (approved):

PR #1604, opened 2024-02-26
fix(menu-item): fix font-size (VIV-1589)

RichardHelm:


libs/components/src/lib/menu-item/menu-item.scss
```diff
@@ -42,15 +42,16 @@
 	box-shadow: inset 0 0 0 1px var(#{appearance.get-appearance-token(outline)});
 	gap: $gap;
 	inline-size: 100%;
-	padding-block: 2px;
```
rachelbt, line 45: it doesn't make sense to me that with two lines has different block-padding. can't they both get either 8 or 12?

rachelbt (commented):
I still think that 12px is rather small :(

rachelbt (commented):
also - not fully related but the FF PNG is a bit off - we need to:
- change the style from `div `to `#wrapper` 
- remove the `div` wrapper of `vwc-menu`
- change the `page.setViewportSize({ width: 250, height: 1800 });`

libs/components/src/lib/menu-item/menu-item.scss
RichardHelm: I've taken this from the Figma file 🤷

rachelbt: @AyalaBu - WDYT?

AyalaBu: It looks fine
Thank you

AyalaBu: The _**right & left padding**_ is supposed to be 12px
Other than that looks great
Thank you

RichardHelm: @AyalaBu Ok, I took another look and tried to align it fully with Figma:

Horizontal padding: 12px

Vertical padding:
  * Basic: 10px
    * In Figma this goes to 8px on multiline, which I consider unintentional
  * Expanded: 12px

Gap:
  * Basic: 12px
  * Expanded: 16px

Secondary text color: `neutral-600` -> `neutral-800`

rachelbt: I think that the basic block padding should be 8px - with 10px we are reaching min-height of 42px, where normally our default height is 40.
I still think the 4px gap between the primary + secondary text is too much - they seem almost unrelated to one another:
image

with 2px:
image

secondary text is 600 in vivid-2. This need to be like it - Figma is not aligned.
it needs to be 800 only when selected .
@AyalaBu  am I correct?

AyalaBu: I agree - the padding-block in basic need so be 8px.
The gap between primary + secondary needs to be 4px as @RichardHelm did.
Thanks

RichardHelm: @AyalaBu, what about the secondary text color?

libs/components/src/lib/menu-item/menu-item.scss
```diff
@@ -131,15 +131,17 @@
...
 	}
 }
 
 .text-secondary {
+	font: var(#{constants.$vvd-typography-base-condensed});
 	-webkit-line-clamp: var(#{menu-item-variables.$text-secondary-line-clamp}, 1);
 	.base.two-lines:not(.disabled) & {
-		color: var(#{constants.$vvd-color-neutral-600});
+		color: var(#{constants.$vvd-color-neutral-800});
 	}
```
rachelbt, line 146:
```suggestion
	}
		.base.two-lines:where(.selected):not(.disabled) & {
		color: var(#{constants.$vvd-color-neutral-800});
	}
```

RichardHelm:
@rachelbt I've addressed the comments, please have another look

rachelbt (approved):
LGTM

rachelbt (approved):

PR #1612, opened 2024-02-28
fix(button): link doesn't work when it's inside the data-grid (VIV-1587)

rinaok:


rachelbt (changes requested):
the issue with the button link not opening is not related to the link button but to the data-grid.
Look at this cpode pen - the button-link is OK outside the data grid but inside the data-grid both the plain link and button link are not opening:
https://codepen.io/Rachel-Tannenbaum/pen/ZEZEOeB

PR #1614, opened 2024-02-29
fix(data-grid): links inside data-grid (VIV-1587)

rinaok:


libs/components/src/lib/data-grid/ui.test.ts
```diff
@@ -276,9 +276,52 @@ async function testSortColumns({ page }: { page: Page }) {
...
+		template,
+	});
+
+	await page.waitForLoadState('networkidle');
+
+	const link = await page.locator('vwc-data-grid-cell a');
+	await link.isVisible();
+	await link.click();
+	await page.pause();
+	expect(await page.url()).toEqual('https:/google.com/');
```
rachelbt, line 318:
```suggestion
	expect(await page.url()).toEqual('https://www.google.com/');
```

rachelbt (commented):
Nice :) 
if there's a new ui test - how come there's no new PNG? :) 
And if it do not test the UI then - maybe this need to go to the unitest?

rinaok:
> Nice :) if there's a new ui test - how come there's no new PNG? :) And if it do not test the UI then - maybe this need to go to the unitest?

We have other tests like that for example in dialog.

rachelbt (approved):

PR #1619, opened 2024-03-01
chore(all): format json, yml and html (VIV-1605)

RichardHelm:
As a first step, code formatting for just these file types


rachelbt (commented):
why are the indentation are so "big"?
image

How will we keep the indentation aligned - we (vivid devs) are using both vs code and WebStorm

RichardHelm:
@rachelbt Files are formatted with tabs (as it's set in our `.editorconfig`). You can configure how wide a tab appears in GitHub: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-user-account-settings/managing-your-tab-size-rendering-preference

The CI pipeline enforces the formatting and will fail if new code is not properly formatted.

Dev's should setup their IDE to auto-format code as they write it, see: https://prettier.io/docs/en/editors

rachelbt:
> Dev's should setup their IDE to auto-format code as they write it, see: https://prettier.io/docs/en/editors

think we need to add this somewhere in the docs (internal, not site)

rachelbt (approved):
LGTM

PR #1620, opened 2024-03-03
chore(text-field, combobox, number-field): refactor focus (VIV-1572)

rachelbt:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -196,16 +192,26 @@ $low-ink-color: --_low-ink-color;
...
-	// only text-field has wrapper.
+		@include focus.focus;
+	}
+}
 
-	.wrapper:not(:focus-visible, :focus-within) > &,
-	.fieldset:not(:focus-visible, :focus-within) > & {
-		display: none;
+@supports not selector(:has(*)) {
+	.wrapper:focus-within::after {
```
RichardHelm, line 206: If this code works fine, why add the one with `:has` as well? It isn't even any simpler

libs/components/src/lib/combobox/combobox.scss
```diff
@@ -50,3 +51,28 @@
...
+@supports selector(:has(*)) {
+.fieldset:has(.control:focus-within)::after {
+	#{focus-variables.$focus-stroke-gap-color}: transparent;
+
+	@include focus.focus;
+}
+}
+
+@supports not selector(:has(*)) {
+.fieldset:focus-within::after {
```
RichardHelm, line 66: We could consider adding wrapper element to combobox as well to simplify the CSS

libs/components/src/shared/patterns/focus.ts
RichardHelm, line 1: 🎉

RichardHelm (changes requested):

libs/components/src/lib/text-field/text-field.scss
rachelbt: bacause I do not like the face that we add an addition pseudo element. Its the right way to use has. I guess in 1/2 a year or so we can remove the fallback

libs/components/src/lib/combobox/combobox.scss
rachelbt: Its a good question wether we should add it.
We don't need an addition HTML wrapper, but on the other hand it helps consistency and cleaner css.
Either way - I wouldn't want to do thin in this PR.

libs/components/src/lib/text-field/text-field.scss
RichardHelm: But the `:has` code is also using a pseudo element 🤔

rachelbt: that we already have :) Not creating a new one for the focus

RichardHelm (approved):

PR #1624, opened 2024-03-05
chore(all): format css/scss (VIV-1619)

RichardHelm:


rachelbt (dismissed):
LGTM
I still think we need to document pre-setting of IDE

rachelbt (approved):

PR #1629, opened 2024-03-07
feat(dial pad): new composed component (VIV-1512)

rinaok:


apps/docs/_data/components.json
```diff
@@ -268,5 +268,10 @@
 	{
 		"title": "Audio Player",
 		"markdown": "./libs/components/src/lib/audio-player/README.md"
+	},
+	{
```
rachelbt, line 272: I wonder if this needs to be under: composed components or ant other title.

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
+.base {
```
rachelbt, line 1:
```suggestion
@use "../../../../../dist/libs/tokens/scss/tokens.constants" as constants;

.base {
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
...
+    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-rows: 1fr 1fr 1fr 1fr;
+    inline-size: 230px;
+}
+
+.phone-field {
+    --vvd-size-density: 2;
+
+    align-self: flex-start;
+    border-bottom: 1px solid var(--vvd-color-neutral-300);
```
rachelbt, line 20:
```suggestion
    border-bottom: 1px solid var(#{constants.$vvd-color-neutral-300});
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
...
+    grid-column: 1 / -1;
+}
+
+.digit-btn {
+    display: flex;
+    flex-direction: column;
+    border: 1px solid var(--vvd-color-neutral-300);
```
rachelbt, line 27:
```suggestion
    border: 1px solid var(#{constants.$vvd-color-neutral-300});;
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
...
+    border-radius: 16px;
```
rachelbt, line 28:
```suggestion
     border-radius: #{border-radius-variables.$border-radius-normal};
```
also regarding the border-radius - nested vwc-buttons have 8px border and with here 16px ic causes design break in hover:
image
maybe vwc-buttons need to be rounded?

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
+.base {
+    display: grid;
+    margin: 16px;
+    gap: 16px;
+    inline-size: 230px;
```
rachelbt, line 5: since this is repeating it will be nice to move it to a sass variable.
Also this is OK for desktop but it can look strange on mobile 230px when there's 460px screen width...

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.digits {
+    display: grid;
+    gap: 16px;
+    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-rows: 1fr 1fr 1fr 1fr;
+    inline-size: 230px;
```
rachelbt, line 13: same - regarding width

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,33 @@
...
+    margin: 16px;
+    gap: 16px;
+    inline-size: 230px;
+}
+
+.digits {
+    display: grid;
+    gap: 16px;
+    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-rows: 1fr 1fr 1fr 1fr;
```
rachelbt, line 12:
```suggestion
    grid-template-columns: repeat(3, 1fr);
    grid-template-rows: repeat(4, 1fr);
```

rachelbt (commented):
Nice!
Didn't went over the unit-tests :)

apps/docs/_data/components.json
rinaok: @rachelbt I know it's written in the documents that it should, but I left it for the end to decide if it's best.

libs/components/src/lib/dial-pad/dial-pad.scss
rinaok: @rachelbt I think it's a question for @AyalaBu

libs/components/src/lib/dial-pad/dial-pad.scss
rinaok: @rachelbt So should we make the inline-size public?

AyalaBu: The designers' request was for a centered text and also it was bigger.
Can we make it happen?
Also I know that it's pretty common for a dialpad in a cell to appear without text,
but with underline it just looks weird.
Can add some text like "enter a number"? (it was also in the designers' original design)

libs/components/src/lib/dial-pad/dial-pad.scss
AyalaBu: 16px radius is border-radius in stacked button in pill shape
This was done a long time ago and I don't think we can make this shape disappear
Also this was an option chosen by designers in this component

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -0,0 +1,200 @@
...
+	Localized
+} from '../../shared/patterns';
+import { TextField } from '../text-field/text-field';
+import { Button } from '../button/button';
+
+/**
+ * Base class for dial-pad
+ *
+ * @public
+ * @event change - Emitted when the text field value changes
```
RichardHelm, line 13: We now annotate component like this for the wrappers:
```suggestion
 * @public
 * @component dial-pad
 * @event change - Emitted when the text field value changes
```

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -0,0 +1,200 @@
...
+ * @event input - Emitted when the text field value changes
+ * @event blur - Emitted when the text field loses focus
+ * @event focus - Emitted when the text field receives focus
+ * @event keypad-click - Emitted when a digit button is clicked
+ * @event dial - Emitted when the call button is clicked
+ * @event end-call - Emitted when the end call button is clicked
+ *
```
RichardHelm, line 20: Can we add a v-model for value as well?

RichardHelm (changes requested):
Looks great! I left some comments

libs/components/src/lib/dial-pad/dial-pad.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+			const spy = jest.fn();
+			element.addEventListener('end-call', spy);
+			element.active = true;
+			await elementUpdated(element);
+			getCallButton().click();
+			expect(spy).toHaveBeenCalledTimes(1);
+		});
+	});
+
+	describe.each(['input', 'change', 'blur', 'focus'])('%s event', (eventName) => {
```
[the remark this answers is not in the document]
rinaok: added

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -0,0 +1,200 @@
...
+	@attr({ attribute: 'digit-three-aria-label' }) digitThreeAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-four-aria-label' }) digitFourAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-five-aria-label' }) digitFiveAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-six-aria-label' }) digitSixAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-seven-aria-label' }) digitSevenAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-eight-aria-label' }) digitEightAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-nine-aria-label' }) digitNineAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-asterisk-aria-label' }) digitAsteriskAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-zero-aria-label' }) digitZeroAriaLabel: string | null = null;
+	@attr({ attribute: 'digit-hashtag-aria-label' }) digitHashtagAriaLabel: string | null = null;
```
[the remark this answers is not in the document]
rinaok: @RichardHelm WDYM? They exists in the locale:


```
dialPad: {
		inputLabel: 'Phone number',
		deleteButtonLabel: 'Delete',
		callButtonLabel: 'Call',
		endCallButtonLabel: 'End call',
		digitOneLabel: '1',
		digitTwoLabel: '2 ABC',
		digitThreeLabel: '3 DEF',
		digitFourLabel: '4 GHI',
		digitFiveLabel: '5 JKL',
		digitSixLabel: '6 MNO',
		digitSevenLabel: '7 PQRS',
		digitEightLabel: '8 TUV',
		digitNineLabel: '9 WXYZ',
		digitAsteriskLabel: '*',
		digitZeroLabel: '0',
		digitHashtagLabel: '#',
	},
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+
+	align-self: flex-start;
+	grid-column: 1 / -1;
+}
+
+.digit-btn {
+	display: flex;
+	flex-direction: column;
+	border: 1px solid var(#{constants.$vvd-color-neutral-300});
+	border-radius: #{border-radius-variables.$border-radius-extra-expanded};
```
rachelbt, line 34:
```suggestion
	border-radius: #{border-radius-variables.$border-radius-expanded};
 overflow: hidden;
 inline-size: 100%;
```
I would go with this one, and remove the shape-pill from the the digits buttons.

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
+@use '../../../../../dist/libs/tokens/scss/tokens.constants' as constants;
+@use '../../../../shared/src/lib/sass/mixins/border-radius' as
+	border-radius-variables;
+
+$inline-size: 230px;
+$gap: 16px;
+
+.base {
+	display: grid;
+	margin: 16px;
```
rachelbt, line 10: I wonder if it will be better on the host and if someone wants to remove it or change it it will be easier

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+	gap: $gap;
```
rachelbt, line 11:
```suggestion
 grid-template-rows: 80px 1fr auto;
```
this way - there will be no layout change when the error text or helper text is added (as long as it is max 2 lines there)

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+
+.digit-btn {
+	display: flex;
+	flex-direction: column;
+	border: 1px solid var(#{constants.$vvd-color-neutral-300});
+	border-radius: #{border-radius-variables.$border-radius-extra-expanded};
+}
+
+.call-btn {
+	margin-top: 12px;
```
rachelbt, line 38:
```suggestion
	margin-top: 24px;
```
I removed the gap in base so this need to be bigger

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+	gap: $gap;
+	inline-size: $inline-size;
+}
+
+.digits {
+	display: grid;
+	gap: $gap;
+	grid-template-columns: repeat(3, 1fr);
+	grid-template-rows: repeat(4, 1fr);
+	inline-size: $inline-size;
```
rachelbt, line 20:
```suggestion
	inline-size: 100%;
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+	border-radius-variables;
+
+$inline-size: 230px;
+$gap: 16px;
+
+.base {
+	display: grid;
+	margin: 16px;
+	gap: $gap;
+	inline-size: $inline-size;
```
rachelbt, line 12:
```suggestion
	inline-size: 100%;
	box-sizing: border-box;
```
I think that the inline-size needs also to be set on host.
I also think that 230px is maybe nice on desktop but maybe will be needed to be smaller or a bit wider. 
Moving the width to host will make the devs life easier

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+	--vvd-size-density: 2;
+
+	align-self: flex-start;
+	grid-column: 1 / -1;
+}
+
+.digit-btn {
+	display: flex;
+	flex-direction: column;
+	border: 1px solid var(#{constants.$vvd-color-neutral-300});
```
rachelbt, line 33:
```suggestion
	box-shadow: inset 0 0 0 1px  var(#{constants.$vvd-color-neutral-300});
```

libs/components/src/lib/dial-pad/dial-pad.scss
```diff
@@ -0,0 +1,40 @@
...
+
+.phone-field {
+	--vvd-size-density: 2;
+
+	align-self: flex-start;
+	grid-column: 1 / -1;
+}
+
+.digit-btn {
+	display: flex;
```
rachelbt, line 31:
```suggestion
--vvd-button-accent-primary: var(#{constants.$vvd-color-neutral-600});

	display: flex;
```
this is for the text of the digits.
but this will also effect the digits - so you'll need to add class for the slotted icon and set it with color:

rachelbt (commented):
left some css comments to make the dial-pad more accurate to design.
will need to go over again after changes are made :)

libs/components/src/lib/dial-pad/dial-pad.scss
rinaok: @AyalaBu approved this shape, I don't think we can change it

libs/components/src/lib/dial-pad/dial-pad.ts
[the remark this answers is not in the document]
RichardHelm: Yes, but the values from the locale are not being used
The proposal states that the aria-labels will come from the locale and it doesn't have any attributes to customise them, so that is what I would expect here

libs/components/src/lib/dial-pad/dial-pad.ts
RichardHelm: There is some documentation in wrapper-gen README.
It probably needs to look like this: `` * @vueModel modelValue value input `(event.target as any).value` ``

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 *
+	 * @internal
+	 */
+	_onDigit = (e: Event) => {
+		if (
+			e.target === undefined ||
+			e.target === null ||
+			!(e.target instanceof Button)
```
RichardHelm, line 111: Click doesn't register when clicking on icon

RichardHelm (changes requested):

RichardHelm (approved):

PR #1640, opened 2024-03-18
feat(text-field): custom for dialpad (VIV-1618)

rachelbt:


libs/components/src/lib/text-field/text-field.ts
```diff
@@ -89,6 +89,9 @@ export class TextField extends FoundationTextfield {
 	@observable actionItemsSlottedContent?: HTMLElement[];
 	@observable leadingActionItemsSlottedContent?: HTMLElement[];
 
+
+
+
```
rachelbt, line 94:
```suggestion
```

rinaok (dismissed):

rinaok (approved):

PR #1641, opened 2024-03-19
feat(data-grid): support dynamic row height (VIV-1602)

RichardHelm:


apps/docs/assets/styles/main.css
```diff
@@ -59,6 +59,10 @@ img {
 	block-size: auto;
 }
 
+vwc-note > p {
```
rachelbt, line 62: why is the slotted text gets wrapped with P element?
Is it eleventy?
If its the component - I think we need to open a ticker for it

libs/components/src/lib/data-grid/README.md
```diff
@@ -480,7 +480,9 @@ Event details: `{ cell, row, isHeaderCell, columnDataKey }`
 </script>
 ```
 
-## CSS Variable
+## CSS Variables
+
+### --data-grid-row-background
```
rachelbt, line 485:
```suggestion
### Row Background
```
it looks strange that only this one gets the actual name as headline

libs/components/src/lib/data-grid/README.md
```diff
@@ -506,6 +508,91 @@ vwc-data-grid {--data-grid-row-background: var(--vvd-color-neutral-50);}
...
+		<vwc-data-grid-cell>
+			<vwc-button label="Action" appearance="outlined"></vwc-button>
+		</vwc-data-grid-cell>
+	</vwc-data-grid-row>
+	<vwc-data-grid-row class="dynamic-height">
+		<vwc-data-grid-cell>
+			Variable height
+		</vwc-data-grid-cell>
+		<vwc-data-grid-cell>
+			<vwc-button label="Action" appearance="outlined"></vwc-button>
```
rachelbt, line 550:
```suggestion
			<vwc-button label="Action" appearance="outlined" size="expanded"></vwc-button>
```
to make the diff bigger :)

libs/components/src/lib/data-grid/data-grid-cell.scss
```diff
@@ -31,11 +33,15 @@
 	display: flex;
 	box-sizing: border-box;
 	align-items: center;
-	padding: 8px 12px;
+	padding: 14px 12px;
```
rachelbt, line 36:
```suggestion
	padding: 12px;
```
14px is not on the 8px scale...

apps/docs/assets/styles/main.css
RichardHelm: No this is because because of markdown, every paragraph in gets wrapped in a `<p>`

libs/components/src/lib/data-grid/data-grid-cell.scss
RichardHelm: The value is taken from the current design in Figma. I've asked Ayala to comment on it.
However, I don't think it really matters, as changing the margin would be a breaking change and is unrelated to this ticket.

rachelbt: I don't think it will be a breaking change as cell has min-height of 48px, so it can be either way.
And only if its bigger then 48px then the padding make a difference.

RichardHelm: Yes, but I think the padding should always be the same

rachelbt: @AyalaBu ?

AyalaBu: I think that top margin and bottom margin should stay as they are right now,
Even the text has multiple lines.
If the text "jumps" while filling the table is a very unpleasant experience
And also it there are a few cells one near another with different amount of content and different top margin is very unpleasant to the eye.

rachelbt (approved):

PR #1651, opened 2024-03-28
fix(docs): center headings (VIV-000)

rachelbt:


RichardHelm:
Why are we changing this back and forth? Could you reach a decision with @TaylorJ76 please

rinaok (changes requested):
image

rachelbt:
@rinaok @RichardHelm @TaylorJ76 @AyalaBu 
I dont mind if its not centered - but the centered gradient looks strange if the text is not centered.
So either all is centered or aligned left and we change the gradient to be different and start from left

PR #1656, opened 2024-04-03
feat(audio-player): skip buttons (VIV-1645)

rachelbt:


libs/consts/src/lib/icons.ts
```diff
@@ -1,2 +1,2 @@
 export const ICONS_BASE_URL = 'https://icon.resources.vonage.com';
-export const ICONS_VERSION = '4.5.3';
+export const ICONS_VERSION = '4.5.5';
```
rachelbt, line 2:
```suggestion
export const ICONS_VERSION = '4.5.6';
```

rinaok (changes requested):
Clicking on the slider is not working anymore.

rinaok (approved):

PR #1657, opened 2024-04-03
fix(icon): fix icon size `0` not working when set as prop (VIV-1705)

RichardHelm:


rachelbt (commented):
I'm not sure I undersatand the change. Now if someone sets `size=""` it behaves as 0, but I'm not sure this is what is expected, also - it can be a breaking change if someone placed size but left it with no size...

RichardHelm:
@rachelbt The bug is that setting size 0 as a property did not work previously

`size=""` turning in 0 is unintended and I have fixed it now

rachelbt:
> The bug is that setting size 0 as a property did not work previously

I'm not sure I understand...
It look OK here:
image

RichardHelm:
@rachelbt In the example it's not being set as a property. If you do it doesn't work

rachelbt (approved):
👍 
BTY - Is this also needed in Video-player for the skip buttons?

RichardHelm:
@rachelbt I don't think so, because it's a string enum there

PR #1662, opened 2024-04-05
fix(a11y): error message announcement in form elements (VIV-1593)

TaylorJ76:
This can be tested by dynamically adding the `error-text` and the screen reader should announce it.

### DIFF
--- a/libs/components/src/shared/patterns/form-elements/form-elements.ts
+++ b/libs/components/src/shared/patterns/form-elements/form-elements.ts
@@ -176,7 +176,7 @@ type FeedbackConfig = {
 		name: string;
 		slottedContentProperty: '_helperTextSlottedContent';
 	};
-	role: 'status' | 'none'
+	role: 'status' | 'none';
 };
 const feedback: Record<string, FeedbackConfig> = {
 	helper: {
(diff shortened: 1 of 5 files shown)


libs/components/src/shared/patterns/form-elements/form-elements.ts
```diff
@@ -249,10 +249,16 @@ function getFeedbackTypeTemplate(
...
 		class="${(x) =>
-			classNames('message', `${config.className}-message`, [
-				'message--visible',
-				shouldShow(x),
-			])}"
+			classNames(
+				'message',
+				`${config.className}-message`,
+				['message--visible', shouldShow(x)],
+				['sr-only', !shouldShow(x)]
```
RichardHelm, line 256: Is this needed? Because when the error is not visually displayed it also shouldn't be accessible to SR, right?
```suggestion
```

libs/components/src/shared/patterns/form-elements/message.scss
```diff
@@ -2,13 +2,27 @@
...
 .message {
 	display: none;
 	contain: inline-size;
 	font: var(#{constants.$vvd-typography-base-condensed});
 	gap: 4px;
 	grid-column: 1 / -1;
 
+	&.error-message {
+		display: flex;
+	}
```
RichardHelm, line 24: Kind of confusing, maybe better to always apply message--visible for error message through code instead
```suggestion
```

libs/components/src/shared/patterns/form-elements/form-elements.ts
TaylorJ76: When it is hidden with display: none the message is not announced.

RichardHelm (approved):

PR #1671, opened 2024-04-16
feat(docs): add toc (VIV-1728)

RichardHelm:


rachelbt (commented):
This is soooo helpful!
Its the first component in our docs that we are not using one o vivid's components :( But I guess its not worth creating it from scratch when we can use the plugin.

I personally like the design but it looks a bit not "vivid". WDYT about this:
image
.article-toc a[aria-active]::before {
    content: "";
    display: block;
    block-size 100%;
    background-color: currentColor;
    inline-size: 2px;
    position: absolute;
    inset-inline-start: -9px;
    z-index: 1;
    inset-block-start: 0;
}
and this to the a:
    border-radius: 8px;
    border: 0;
    margin-inline-start: 8px;
    position: relative;

RichardHelm:
@rachelbt Sure, I updated the design now

rachelbt (commented):
👍 
BTW, is there an option to add classes instead of targeting html elements?

RichardHelm:
@rachelbt No, doesn't seem like it

rachelbt (approved):

PR #1673, opened 2024-04-18
feat(text-field): use helper-text as accessible description (VIV-1592)

RichardHelm:


rachelbt (commented):
looks good 👍 
Is the help text suppose to be read in voice-over?

rachelbt (approved):
LGTM

PR #1675, opened 2024-04-22
chore(docs): Migrate to vite (VIV-1742)

RichardHelm:


apps/docs/assets/scripts/main.ts
```diff
@@ -0,0 +1,10 @@
+import './scroll-handler';
+import './theme-handler';
+import './viewport-handler';
+import './cache-assets';
+import './live-sample';
+import './turbolinks';
+import './toc/toc';
+import './components-filter';
+import './icons-gallery/icons-gallery';
+import './side-drawer';
```
YonatanKra, line 10: Can we add `.js` to each of them to comply with ES imports standard?

libs/components/src/lib/video-player/definition.ts
```diff
@@ -1,6 +1,6 @@
 import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
 import { registerFactory } from '../../shared/design-system';
-import styles from './video-player.scss';
+import styles from './video-player.scss?inline';
```
YonatanKra, line 3: Why the inline? What happens if I remove the `inline` from this import?

tsconfig.base.json
```diff
@@ -22,6 +22,10 @@
 			"@vonage/nx-vivid": ["libs/nx-vivid/src/index.ts"],
 			"@vonage/vivid": ["libs/components/src/index.ts"],
 			"@vonage/vivid-vue": ["libs/vue-wrappers/src/index.ts"],
+			"@vonage/vivid/locales/en-GB": ["libs/components/src/locales/en-GB.ts"],
+			"@vonage/vivid/locales/en-US": ["libs/components/src/locales/en-US.ts"],
+			"@vonage/vivid/locales/ja-JP": ["libs/components/src/locales/ja-JP.ts"],
+			"@vonage/vivid/locales/zh-CN": ["libs/components/src/locales/zh-CN.ts"],
```
YonatanKra, line 28: Seems like a good candidate to extract from the main library. Not for this PR, but as a future dev task. WDYT?

apps/docs/assets/scripts/main.ts
RichardHelm: Sure, done

libs/components/src/lib/video-player/definition.ts
RichardHelm: Otherwise Vite will inject the CSS into the page: https://vitejs.dev/guide/features#disabling-css-injection-into-the-page

apps/docs/_layouts/base.njk
```diff
@@ -23,64 +23,25 @@
...
+			:not(:defined), .page-not-ready {
+				visibility: hidden;
+			}
+		</style>
+		<style id="vivid-theme"></style>
+
+		<script type="text/javascript">
+			window._bufferedLoadedIFrames = [];
+			window.onloadIframe = function(iframe) { window._bufferedLoadedIFrames.push(iframe); };
+		</script>
```
YonatanKra, line 37: What is this script doing? Can it be moved to a file?

tsconfig.base.json
YonatanKra: I know - but they would be separated from the library like the `consts` library.

libs/components/src/lib/video-player/definition.ts
YonatanKra: Oh - it's part of our standard. That's fine, then. I thought video was special :)

apps/docs/_layouts/base.njk
RichardHelm: The iframes use `onload=onloadIframe(this)` but the main.js is now loading async so `onloadIframe` might not be defined on load
So it buffers the events until main.js loads and processes them

PR #1679, opened 2024-04-26
feat(docs): add search (VIV-1748)

RichardHelm:


rachelbt (commented):
NICEEEEEEEEEEEEE!!
Two small comments regarding the design:
Can the focus be the same as vivid focus (in mobile view in dev tools there's a yellow focus as well).
I think the separation between the upper nav elements and the components was helpfull - can we can the border back? (you can add a divider I guess)
I guess the search results need also design but I think that can wait :)

YonatanKra:
Really nice addition.
One major thing that pops up to me. I'm using the current components filter a lot. It the easiest and fastest way to find a component.
I suggest not removing it and keep this PR as its name implies: `add` search :)

YonatanKra (dismissed):
I think we can just reinstate the old components filter and this PR will be good to go.

RichardHelm:
@YonatanKra 
Mhm, I don't agree. I think it would be confusing to have two ways of searching

YonatanKra:
Remarks from @AyalaBu :
1. Enlarge the magnifier icon + "clear" text
2. Change the yellow text background to purple 200

RichardHelm:
@YonatanKra Done

TaylorJ76 (dismissed):

TaylorJ76 (approved):

PR #1681, opened 2024-04-30
fix(audio-player): css for responsive extended audio (VIV-1745)

rachelbt:
This PR is ready for review - I added the playback as a boolean just to make my like easy - this PR is only for the css view of it.

- After code review - will comment out the change, leaving only the css


libs/components/src/lib/audio-player/audio-player.scss
```diff
@@ -1,29 +1,79 @@
 @use '../../../../../dist/libs/tokens/scss/tokens.constants' as constants;
 @use 'partials/variables' as variables;
 
+:host {
+	display: block;
+	inline-size: 350px;
```
TaylorJ76, line 6: This could create a horizontal scroll in small viewports and will fail [this rule](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html) in a11y testing.
Could we change it to `max-inline-size`?

TaylorJ76 (changes requested):

libs/components/src/lib/audio-player/audio-player.scss
rachelbt, line 6:
```suggestion
	max-inline-size: 350px;
```

TaylorJ76 (approved):

libs/components/src/lib/audio-player/definition.ts
```diff
@@ -2,6 +2,7 @@ import type { FoundationElementDefinition } from '@microsoft/fast-foundation';
 import { registerFactory } from '../../shared/design-system';
 import { buttonRegistries } from '../button/definition';
 import { sliderRegistries } from '../slider/definition';
+import { menuRegistries } from '../menu/definition';
```
rachelbt, line 5:
```suggestion

```

libs/components/src/lib/audio-player/definition.ts
```diff
@@ -22,6 +23,7 @@ export const audioPlayerRegistries = [
 	audioPlayerDefinition(),
 	...buttonRegistries,
 	...sliderRegistries,
+	...menuRegistries,
```
rachelbt, line 26:
```suggestion

```

libs/components/src/lib/audio-player/README.md
```diff
@@ -81,7 +81,26 @@ Use the `disabled` attribute to disable the audio-player.
...
+		max-inline-size: initial;
+	}
+</style>
+<vwc-audio-player
+	class="audio-player"
+	src="https://download.samplelib.com/mp3/sample-6s.mp3"
+></vwc-audio-player>
+```
+
+### Minimum inline Size **Deprecated**
```
TaylorJ76, line 101: Can we follow the same format as selectable-box for the deprecated variable?

TaylorJ76 (changes requested):

TaylorJ76 (approved):

PR #1682, opened 2024-04-30
feat(dial-pad): input features and fixes (VIV-1744)

rinaok:
1. Add a no-input attribute to remove the input.
2. Remove the background of the text-field in the custom css in text-field css.
3. Change input's type="tel" instead of "text".


rachelbt (commented):
All looks OK by me.
The only thing that I don't really like is the name of the no-input.
Unfortunately I do not have any alternatives as a positive member (I don't line the no-call either :)).
Not sure - `only-buttons` or `NumericTouch` is better 😵‍💫

RichardHelm:
Well, I don't remember having this idea 😄 
But I think `no-input` is fine and consistent with the `no-call` so I wouldn't change anything

RichardHelm (dismissed):
image
I don't know if it's a new issue but I noticed when clicking on the delete button, it shows the input with a 1px border. I don't think that's intended

rachelbt:
> image I don't know if it's a new issue but I noticed when clicking on the delete button, it shows the input with a 1px border. I don't think that's intended

this is my css that did it.
I think I'll open a ticket for it, it is a bit complicated to solve... :(

PR #1690, opened 2024-05-02
fix(menu): make auto-dismiss dismiss on focusout (VIV-1747)

RichardHelm:
Aligns auto-dismiss behaviour with select, which uses focus to dismiss


libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -145,110 +145,39 @@ describe('vwc-menu', () => {
...
 			element.autoDismiss = true;
-			const parent = element.parentElement!;
-			element.remove();
-			parent.appendChild(element);
 
-			document.body.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+			element.dispatchEvent(new FocusEvent('focusout'));
 
 			expect(element.open).toBe(false);
 		});
```
YonatanKra, line 158: Should set these tests in the (badly named :) ) `when anchor is removed` describe to make sure we remove the listener - so as to prevent memory leak.

YonatanKra (changes requested):
Nice job! Simplifies the whole mechanism!
One little request and it's done.
Thanks!

libs/components/src/lib/menu/menu.spec.ts
RichardHelm: Sure, added a test for it

YonatanKra (approved):

PR #1702, opened 2024-05-09
feat(button): add dropdown-indicator option (VIV-1741)

RichardHelm:


libs/components/src/lib/button/button.scss
```diff
@@ -82,20 +83,6 @@
...
-		padding-inline: 0;
-		place-content: center;
-
-		@supports (aspect-ratio: 1) {
-			aspect-ratio: 1;
-		}
-
-		@supports not (aspect-ratio: 1) {
-			inline-size: var(#{variables.$block-size});
-		}
```
rachelbt, line 96: why did you removed this?
This is better then adding padding

libs/components/src/lib/button/button.scss
```diff
@@ -228,15 +212,32 @@ slot[name='icon'] {
...
+	gap: var(#{variables.$icon-gap});
+
+	.control.stacked & {
+		flex-direction: column;
+	}
+}
+
+.chevron {
+	flex-shrink: 0;
+}
```
rachelbt, line 239:
```suggestion
```
this is not needed - its in the chevron file already

rachelbt (commented):
nice 👍

libs/components/src/lib/button/button.scss
RichardHelm: When icon-only button has chevron, it is no longer square

rachelbt (approved):
⚡

rachelbt (approved):

rachelbt (approved):
I think it looks better now with 16px for the chevron in normal

PR #1716, opened 2024-05-19
fix(appearance): change ghost text-color from 500 to 600 (VIV-1774)

rachelbt:
even thought this change is only color change - it effects the css-variables that are generated.
You can see that the dial-pad ui test are failing because the css-variable: `--vvd-button-accent-primary` is not in use anymore.
Wonder what how we can fix it.


libs/components/src/lib/accordion-item/accordion-item.scss
```diff
@@ -16,10 +16,15 @@
...
+$vvd-accordion-item-accent-firm-wrapper: --_vvd-accordion-item-accent-firm-wrapper;
+
 :host {
 	display: flex;
 	box-sizing: border-box;
 	flex-direction: column;
+	#{$vvd-accordion-item-accent-firm-wrapper}: var(
+		--vvd-accordion-item-accent-firm
+	);
 }
```
RichardHelm, line 29: Maybe the ugliness can be better contained with something like this:

```suggestion
:host {
	display: flex;
	box-sizing: border-box;
	flex-direction: column;
}

@include variables-backwards-compatibility();
```

with 

```
@mixin variables-backwards-compatibility() {
  :host {
    // build vars for the connotations configured in connotation/config
  }

  .control {
    // do the mapping
  }
}
```

Will make it easier to delete once the time comes

libs/shared/src/lib/sass/mixins/connotation/_mixins.scss
```diff
@@ -36,3 +36,22 @@
 		);
 	}
 }
+
+@mixin variables-backwards-compatibility($component, $element-class) {
```
RichardHelm, line 40: Let's add some comment to our crazy code

```suggestion
/**
 * Ghost text color has changed from primary to firm in VIV-1774. For backwards compatibility with code
 * setting primary we need to default firm to primary value if it is provided.
 */
@mixin variables-backwards-compatibility($component, $element-class) {
```

RichardHelm (changes requested):
Looks good, minor comments

libs/components/src/lib/split-button/split-button.scss
```diff
@@ -7,7 +7,7 @@
 @use '../../../../shared/src/lib/sass/mixins/connotation/config' with (
 	$connotations: accent cta success alert,
 	$shades: primary primary-text primary-increment contrast fierce firm soft
-		faint,
+		faint firm,
```
rachelbt, line 10:
```suggestion
		faint,
```

RichardHelm (approved):

PR #1724, opened 2024-05-22
feat(slider, range-slider): add pin feature (VIV-1611)

RichardHelm:


libs/components/src/lib/range-slider/range-slider.template.ts
```diff
@@ -40,7 +43,11 @@ export const RangeSliderTemplate: (
 				</div>
 				<div
 					${ref('_startThumbEl')}
-					class="thumb-container"
+					class="${(x) =>
+						classNames('thumb-container', [
+							'focus-visible',
+							x._visiblyFocusedThumb === 'start',
+						])}"
```
YonatanKra, line 50: Can we set this as a function at the top? `getRangeSliderClasses` should be fitting.

libs/components/src/lib/range-slider/range-slider.template.ts
```diff
@@ -53,9 +60,32 @@ export const RangeSliderTemplate: (
...
 					aria-orientation="${(x) => x.orientation}"
 				></div>
+				${when(
+					(x) => x.pin,
+					html<RangeSlider>`<${popupTag}
+					class="popup"
+					arrow
+					alternate
+					:anchor="${(x) => x._startThumbEl}"
+					:open=${(x) => x._isThumbPopupOpen('start')}
```
YonatanKra, line 70: I truly discourage `_` properties and methods.
We can either use internal events, private methods and props etc.
If the property/method is public, then it should be documented. If it is private - it should be private and not accessible outside the class.
You can look at #1726 for an example of how I privatize public functions. We can set a pair-coding session on this one as well.

libs/components/src/lib/range-slider/range-slider.template.ts
```diff
@@ -53,9 +60,32 @@ export const RangeSliderTemplate: (
...
+				</${popupTag}>`
+				)}
 				<div
 					${ref('_endThumbEl')}
-					class="thumb-container"
+					class="${(x) =>
+						classNames('thumb-container', [
+							'focus-visible',
+							x._visiblyFocusedThumb === 'end',
+						])}"
```
YonatanKra, line 88: `getThumbContainerClasses` function?

libs/components/src/lib/range-slider/range-slider.ts
```diff
@@ -387,7 +397,28 @@ export class RangeSlider extends FormAssociatedRangeSlider {
 		};
 	}
 
-	#draggingThumb: false | 'start' | 'end' = false;
+	/**
+	 * @internal
+	 */
```
YonatanKra, line 402: Adding comments like these hurt my eyes while reading the code (I'm not kidding - it physically hurts). Better to make the property/method truly private than giving some obscure indication it is "internal" in a comment. 
I know FAST is doing it. I know a lot of people are doing it. Right now these comments are not part of JS and hence will not help in runtime (they might be, according to a person I know, but it'll take years from now).
So IMO it is better to make functions private.

libs/components/src/lib/slider/README.md
```diff
@@ -139,5 +156,6 @@ Use `value` to set the value of the slider.
 
 ## Accessibility
 
-- The slider's `role` is `slider` which means it needs an accessible label.
-- It is the consumer's concern to add `aria-label` to the `slider` element.
+The slider has a `role` of `slider`, which needs an accessible label. You must provide it using `aria-label`.
+
+You can set the `valueTextFormatter` member to customize how values will be formatted for the sliders' `aria-valuetext` attribute. The formatting will also be used for the tooltip when `pin` is enabled.
```
YonatanKra, line 161: See my comment in the popup regarding this.
Part of my suggestion was to split the a11y part to sub parts using `###` titles.
Here it should be:
`### Role`
`### aria-label`
`### aria-valuetext`
WDYT?

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+			element.pin = true;
+			element.valueTextFormatter = (value) => `${value} bits`;
+			await elementUpdated(element);
+
+			expect(getPopup()!.textContent!.trim()).toBe('5 bits');
+		});
+
+		it('should display ariaValuetext in the popup if provided', async () => {
+			element.pin = true;
+			element.ariaValuetext = 'value text';
```
YonatanKra, line 233: What happens if I use `valueTextFormatter`? Will it override the `aria-valuetext`?

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+			await elementUpdated(element);
+
+			getControlElement(element).focus();
+			getControlElement(element).blur();
+			await elementUpdated(element);
+
+			expect(getPopup()!.open).toBe(false);
+		});
+
+		it('should not show popup when thumb focus is not visible', async () => {
```
YonatanKra, line 281: Try to use positive terms when describing a test. What it should do and not what it shouldn't do. I guess it should `hide` or `remove` the elements from the DOM.
In addition, how do you emulate `thumb focus is not visible`?

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+			await elementUpdated(element);
+
+			expect(getPopup()!.open).toBe(false);
+		});
+	});
+
+	describe('thumb', () => {
+		beforeEach(() => {
+			// Work around JSDOM not supporting delegatesFocus correctly
+			element.focus = () => getControlElement(element).focus();
```
YonatanKra, line 296: Maybe this can enter our test-utils? I think we do that a lot, no?

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+		});
+
+		it('should have visible focus when control is focused', async () => {
```
YonatanKra, line 299:
```suggestion
		it('should have visible-focus class when control is focused', async () => {
```

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+
+			element.dispatchEvent(new MouseEvent('mousedown'));
+			window.dispatchEvent(new MouseEvent('mouseup'));
+			await elementUpdated(element);
+
+			expect(getPopup()!.open).toBe(false);
+		});
+	});
+
+	describe('thumb', () => {
```
YonatanKra, line 293: Is `thumb` a property?
Because if not, the test is wrong. You should test the property the consumer is using the manipulate the `thumb` and test it and not the `thumb` itself.

libs/components/src/lib/slider/slider.spec.ts
```diff
@@ -190,6 +205,183 @@ describe('vwc-slider', () => {
...
+			beforeEach(async () => {
+				element.value = '0';
+				element.orientation = orientation as Orientation;
+				await elementUpdated(element);
+			});
+
+			it('should update value by dragging thumb', async () => {
+				mouseDown(thumb, 0);
+				mouseMove(300);
+				mouseUp();
```
YonatanKra, line 358: This can be simplified with functions like: "dragVertically", "dragHorizontally" maybe?

libs/components/src/lib/slider/slider.template.ts
```diff
@@ -29,18 +32,17 @@ export const getMarkersTemplate = (
...
-) => ViewTemplate<Slider> = () => {
+	context: ElementDefinitionContext,
+	definition: FoundationElementDefinition
+) => ViewTemplate<Slider> = (context: ElementDefinitionContext) => {
+	const popupTag = context.tagFor(Popup);
+
 	/* eslint-disable @typescript-eslint/indent */
 	return html<Slider>`<template
 		role="${(x) => (x.ariaLabel ? 'presentation' : null)}"
+		@focusin="${(x) => x._onFocusIn()}"
```
YonatanKra, line 44: `_` methods are discouraged :)

YonatanKra (changes requested):
Nice job all in all.
Quite a lot of comments, but quite a lot of changes ;)
 It'll probably take a few rounds to complete this CR.

libs/components/src/lib/popup/README.md
```diff
@@ -150,6 +150,22 @@ Use the `placement` attribute to set the placement of the popup around the ancho
...
+
+### Animation Frame
+
+Whether to update the position of the floating element on every animation frame if required.
+
+See the [Floating UI Documentation](https://floating-ui.com/docs/autoUpdate#animationframe) for more information.
+
+- Type: `boolean`
+- Default: `false`
+
```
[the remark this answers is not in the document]
RichardHelm: Popup is an internal component that cannot be used by consumers, so this is just internal documentation

libs/components/src/lib/popup/popup.ts
```diff
@@ -105,6 +130,29 @@ export class Popup extends FoundationElement {
...
+	 */
+	placementStrategy = PlacementStrategy.Flip;
+
+	/**
+	 * Whether to update the position of the floating element on every animation frame if required.
+	 *
+	 * @public
+	 * HTML Attribute: animation-frame
+	 */
+	@attr({ mode: 'boolean', attribute: 'animation-frame' }) animationFrame =
```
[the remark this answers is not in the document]
RichardHelm: Same as above, popup is internal

libs/components/src/lib/slider/slider.spec.ts
RichardHelm: `valueTextFormatter` has a default value. So no, it will not

libs/components/src/lib/slider/slider.spec.ts
RichardHelm: No thumb refers to the actual thumb of the slider

libs/components/src/lib/slider/slider.spec.ts
RichardHelm: I don't really care what class it has, only that focus is not visible

libs/components/src/lib/range-slider/README.md
```diff
@@ -193,7 +214,7 @@ The upper value of the range.
 Both thumbs have a `role` of `slider`, which needs an accessible label. By default, they use a localized version of "min" and "max".
 You can change the labels by setting the `aria-start-label` and `aria-end-label` attributes.
 
-You can set the `valueTextFormatter` member to customize how values will be formatted for the thumbs' `aria-valuetext` attribute.
+You can set the `valueTextFormatter` member to customize how values will be formatted for the thumbs' `aria-valuetext` attribute. The formatting will also be used for the tooltip when `pin` is enabled.
```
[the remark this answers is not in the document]
RichardHelm: Sure but I don't want to start adding sections in this PR. It is not how we do it elsewhere so changing it is out of scope of the PR

libs/components/src/lib/range-slider/range-slider.spec.ts
```diff
@@ -405,6 +525,17 @@ describe('vwc-range-slider', () => {
 				expect(element.end).toBe('7');
 			});
 
+			it('should move non-visible focus to the corresponding thumb', async () => {
```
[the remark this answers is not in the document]
RichardHelm: Updated it

libs/components/src/lib/slider/README.md
RichardHelm: As above, I don't want to make these changes in this PR

libs/components/src/lib/range-slider/README.md
```diff
@@ -142,6 +142,27 @@ Toggles markers display.
 <vwc-range-slider connotation="cta"></vwc-range-slider>
 ```
 
+### Pin
+
+Toggles display the start and end values with a tooltip. Use the `valueTextFormatter` member to customize the format of the values.
```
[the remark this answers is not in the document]
YonatanKra: Looking at it now it makes even less sense. Why `start` and `end` values? It just shows the current value, doesn't it?

libs/components/src/lib/range-slider/README.md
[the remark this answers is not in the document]
YonatanKra: So we don't do something right because we did it wrong until now? ;)
Let's start a change - this section is supposed to inform people regarding a11y features and IMO my suggestion sets a11y considerations at the forefront.

RichardHelm:
Splitting this PR into smaller ones

PR #1731, opened 2024-05-30
fix(select): design alignment (VIV-1755)

rachelbt:


RichardHelm (approved):
Code looks good.

I don't like the aesthetic of the flip / animation, but up to design

PR #1732, opened 2024-05-30
fix(slider): new marker design (VIV-1766)

rachelbt:
- [x] markers disabled


libs/components/src/lib/slider/ui.test.ts
RichardHelm, line 1: 👍

RichardHelm (approved):
Thanks, matches design now

There is an problem with rendering though, probably because of subpixel placement:
image

Approved anyway, as it is not a new issue

RichardHelm (approved):

PR #1734, opened 2024-05-31
chore(docs): clean dist folder (VIV-000)

RichardHelm:


YonatanKra (approved):
LGTM. Mind just adding some context - what does this solve?

RichardHelm:
@YonatanKra The watch mode could create lots of code snippet html files, which then would never be cleared and slow down the build, so this PR fixes both

PR #1736, opened 2024-06-03
feat(slider, range-slider): add pin feature (VIV-1611)

RichardHelm:


libs/components/src/lib/popup/popup.ts
```diff
@@ -61,7 +93,7 @@ export class Popup extends FoundationElement {
 	open = false;
 	openChanged(_: boolean, newValue: boolean): void {
 		newValue ? this.$emit('vwc-popup:open') : this.$emit('vwc-popup:close');
-		this.#updateAutoUpdate();
+		DOM.queueUpdate(() => this.#updateAutoUpdate());
```
YonatanKra, line 96: Now I see that the reason we have 1 `rAF` in the case `animationFrame` is false. 
I gather a few things from this:
1. If we delete the `DOM.queueupdate` and leave just the `autoUpdate` all tests are passing. It is either irrelevant or it should be tested in the `open` tests. `should update position in the next animation frame` or something like that. 
2. This explains why when we set `animationFrame` to false, we get one `rAF` call. So I'd might change the `animationFrame` property's name to something more meaningful like `softPositionAnimation` or `smoothPositionAnimation` or `gradualPositionAnimation`. WDYT?

libs/components/src/lib/popup/popup.ts
```diff
@@ -105,6 +137,29 @@ export class Popup extends FoundationElement {
...
+	 * HTML Attribute: animation-frame
+	 */
+	@attr({ mode: 'boolean', attribute: 'animation-frame' }) animationFrame =
+		false;
+
+	/**
+	 * @internal
+	 */
+	animationFrameChanged() {
+		this.#updateAutoUpdate();
```
YonatanKra, line 160: `updateAutoUpdate` is not a descriptive name ;)
Wouldn't `updatePopupParmeters` be more descriptive?
I'm open for other names as well :)

YonatanKra (changes requested):
Nice!
Thanks for this split and work well done.
Because now I understand that `popup` is our "facade" for `floatingUI` it makes much more sense.
3 comments and we're done here :)

libs/components/src/lib/popup/README.md
```diff
@@ -150,6 +150,22 @@ Use the `placement` attribute to set the placement of the popup around the ancho
 </script>
 ```
 
+### Placement Strategy
+
+Controls the placement strategy of the popup. The `Flip` strategy will flip the popup to the opposite side if there is not enough space. The `AutoPlacementHorizontal` and `AutoPlacementVertical` strategies will place the popup on the side with the most space in the respective axis.
+
+- Type: `PlacementStrategy.Flip` | `PlacementStrategy.AutoPlacementHorizontal` | `PlacementStrategy.AutoPlacementVertical`
```
rachelbt, line 157: why no use the same convention like placement with value?
`placement-strategy = "flip" | "auto-placement-horizontal" | "auto-placement-vertical"`

rachelbt (commented):
I haven't assigned myself and saw that @YonatanKra reviewed. Adding only one comment

libs/components/src/lib/popup/popup.ts
RichardHelm: No, this change has nothing to do with animationFrame or its tests
If you delete it, the ui tests will fail. The reason is that it is wrong to call autoUpdate before open has changed in the the DOM, because floatingUI requires the floating element to be measurable. Which it isn't if open is false (`display: none`)

libs/components/src/lib/popup/popup.ts
RichardHelm: Mhm, I don't think the name is that bad. Anyway, I would say that is out of scope of this PR

libs/components/src/lib/popup/README.md
RichardHelm: Well, the actual reason why it is that way is that originally `PlacementStrategy.Flip` would be the actual middleware and not a string. But I had to change it to support the UI tests.

I don't think anything is improved by changing it now

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -254,6 +256,64 @@ describe('vwc-popup', () => {
...
+			element.anchor = anchor;
+			await elementUpdated(element);
+			jest.spyOn(window, 'requestAnimationFrame');
+
+			element.open = true;
+
+			expect(window.requestAnimationFrame).toHaveBeenCalledTimes(
+				RAF_CALLS_FROM_SETTING_ATTRIBUTE
+			);
+		});
```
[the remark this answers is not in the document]
YonatanKra: I do not believe so. 
We do care about the amount of times rAF is called.
Floating UI is an implementation detail. You can implement this on your own or switch library - the tests should still pass :)

libs/components/src/lib/popup/popup.ts
YonatanKra: Then why not add a small unit test for it?
It should autoUpdate after `open` changed (e.g. after `elementUpdated`)

libs/components/src/lib/popup/popup.ts
YonatanKra: Well... it made me read the function in order to just understand what it does, which makes it a bad name (always take the weak link as your test case - and I'm the weak link here ;) ).

RichardHelm: Sure, maybe there is a better name but this review is already getting very big so I want to focus only the things actually in the PR

libs/components/src/lib/popup/popup.spec.ts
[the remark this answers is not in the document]
YonatanKra: We are not testing dependencies for correctness. The fact you are using Floating UI is irrelevant. It is an implementation detail. We are testing the outcome of this - so if we removed floating UI or the corresponding code that implements this API, the test should fail.

libs/components/src/lib/popup/popup.ts
YonatanKra: This review has only 4 comments ;)

libs/components/src/lib/popup/popup.spec.ts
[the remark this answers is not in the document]
RichardHelm: I've updated the test now to test that it continuously calls rAF. However the test is still bad and doesn't really test anything at all

If I change the test further to test that it actually updates position I end up testing floatingUI through popup...

Testing the outcome is not always what you want to do. E.g. when you have some code that sends an email, you don't want to test that it sends an actual email. Just that it calls some other code that is responsible for sending emails.

I'd argue that it's the same here. It's not popups job to calculate positions and keep them up to date, so we just verify that calls the code that is responsible for that

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -156,10 +156,48 @@ describe('vwc-popup', () => {
...
+			expect(floatingUI.autoUpdate).not.toHaveBeenCalled();
+		});
+
+		it('should begin to auto update after DOM is updated', async function () {
+			element.anchor = anchor;
+			element.open = true;
+			await elementUpdated(element);
+
+			expect(floatingUI.autoUpdate).toHaveBeenCalled();
+		});
```
YonatanKra, line 198: These can be combined to a single test:
```suggestion
		it('should begin to auto update after DOM is updated', async function () {
			element.anchor = anchor;
			element.open = true;
			const updateCallsBeforeDOMUpdate = floatingUI.autoUpdate.mock.calls.length;
			await elementUpdated(element);
      expect(updateCallsBeforeDOMUpdate).toBe(0);
			expect(floatingUI.autoUpdate).toHaveBeenCalledTimes(1);
		});
```

libs/components/src/lib/popup/popup.spec.ts
```diff
@@ -254,6 +292,137 @@ describe('vwc-popup', () => {
...
+			expect(floatingUI.computePosition).toHaveBeenLastCalledWith(
+				expect.anything(),
+				expect.anything(),
+				expect.objectContaining({
+					middleware: expect.arrayContaining([
+						expect.objectContaining({
+							name: 'autoPlacement',
+						}),
+					]),
+				})
```
YonatanKra, line 319: If we are going for this testing strategy, we should at least make sure the function is called with the correct parameters (anchor, popup element etc.)

YonatanKra (changes requested):

libs/components/src/lib/popup/popup.spec.ts
YonatanKra: Ok.
I deleted `this.anchor` from the call to `autoUpdate` and tests still passed.
Can you share where this is tested?

RichardHelm: Apparently nowhere then ;) But that is unrelated to my change

YonatanKra (approved):

TaylorJ76 (approved):

PR #1742, opened 2024-06-03
feat(eslint-plugin): add no-slot-attribute rule (VIV-1700)

RichardHelm:


libs/eslint-plugin/src/rules/no-slot-attribute.ts
```diff
@@ -0,0 +1,84 @@
+import * as utils from 'eslint-plugin-vue/lib/utils';
+import type { Rule } from 'eslint';
+import { ComponentMetadata } from '../utils/ComponentMetadata';
+import { normalizeTag } from '../utils/components';
+
+const componentsWithNamedSlots = new ComponentMetadata<null>();
```
rachelbt, line 6: can't we do this automatically?
I'm afraid that we will miss component - like - banner and dial-pad are missing,

RichardHelm: You're right, we should get them automatically. But currently we have no way of doing it

I've checked the list again and updated it. Dial-pad doesn't have named slots so it's not included

rachelbt: can't see button (icon) + banner (icon + Action Items)

RichardHelm: You're right 🤦‍♂️ Updated

rachelbt: we need to find a better way for that, or maybe add all the component all together, as you are not specifying slots name. WDYT?

RichardHelm: I see your point but we currently don't have a way to get any metadata, not even a list of all components

rachelbt (approved):

rachelbt (approved):

PR #1749, opened 2024-06-06
feat(combobox): adds ghost appearance to combobox (VIV-1784)

TaylorJ76:


libs/components/src/lib/combobox/README.md
```diff
@@ -104,6 +104,26 @@ Use the `placement` member to set the combobox's placement in accordance to its
...
 ```
 
+### Appearance
+
+Set the `appearance` attribute to change the combobox's appearance.
+
+- Type: `'fieldset'` | `'ghost'`
+- Default: `'fieldset'`
+
+(`'ghost'` is typically used within a composition such as action group / toolbar).
```
rachelbt, line 114:
```suggestion
(`'ghost'` is typically used within a composition such as action-group / toolbar or slotted inside text-field).
```

libs/components/src/lib/combobox/combobox.spec.ts
```diff
@@ -104,6 +105,18 @@ describe('vwc-combobox', () => {
 		});
 	});
 
+	describe('appearance', function () {
+		it('should set the shape class on the root', async function () {
```
rachelbt, line 109:
```suggestion
		it('should set the appearance class on the `.control`', async function () {
```

libs/components/src/lib/combobox/ui.test.ts
```diff
@@ -24,6 +24,7 @@ test('should show the component', async ({ page }: { page: Page }) => {
 				}
 			</style>
 			<vwc-combobox></vwc-combobox>
+			<vwc-combobox appearance="ghost"></vwc-combobox>
```
rachelbt, line 27: this is a little strange, maybe add a placeholder?

rachelbt (commented):
left some comments.
While you are in the combobox - mind adding "pinter-events: none" to the `@supports not selector(:has(*))` - to prevent the same bug in VIV-1788 🙏

rachelbt (approved):
👍 

LGTM

PR #1752, opened 2024-06-07
feat(eslint-plugin): add no-anchor-attribute rule (VIV-1701)

RichardHelm:


libs/eslint-plugin/src/rules/no-anchor-attribute.md
```diff
@@ -0,0 +1,19 @@
...
+```html
+<!-- ❌ BAD -->
+<VButton id="anchor" label="Save"></VButton>
+<VTooltip anchor="anchor" text="Tooltip text"></VTooltip>
+
+<!-- ✅ GOOD -->
+<VTooltip text="Tooltip text">
+	<template #anchor><VButton label="Save"></VButton></template>
+</VTooltip>
+```
```
YonatanKra, line 19: So, from what I figure, this rule is valid only for Vue. Right?

RichardHelm: Yes, all rules are Vue only for now

YonatanKra (approved):

TaylorJ76 (approved):

PR #1756, opened 2024-06-10
fix(tokens): updates dark theme elevation shadow color (VIV-1765)

TaylorJ76:


rachelbt (changes requested):
although the tokens are in the new version - the old tokens are there and not the fix :(
this is what is now - with the old token:
image
needs to be:
image

rachelbt (approved):
🎉

PR #1775, opened 2024-06-14
chore(deps): moves angular eslint to dev deps (VIV-000)

TaylorJ76:


libs/eslint-plugin/.eslintrc.json
```diff
@@ -2,7 +2,8 @@
 	"extends": ["../../.eslintrc.json", "prettier"],
 	"ignorePatterns": ["!**/*"],
 	"rules": {
-		"@typescript-eslint/no-non-null-assertion": "off"
+		"@typescript-eslint/no-non-null-assertion": "off",
+		"@typescript-eslint/no-attribute-default-value": "off"
```
YonatanKra, line 6: What's that for?

YonatanKra (changes requested):

libs/eslint-plugin/.eslintrc.json
TaylorJ76: Not needed. Removed.

YonatanKra (approved):

PR #1779, opened 2024-06-18
fix(number-field): disabled buttons in number field (VIV-1817)

TaylorJ76:


rachelbt (commented):
looks OK.
I just wonder if we should hide the buttons in readonly & disable as it is in native.
WDYT?

libs/components/src/lib/number-field/number-field.spec.ts
```diff
@@ -618,6 +618,20 @@ describe('vwc-number-field', () => {
 			expect(inertWhenDisabled).toEqual(true);
 		});
 
+		it('should set the disabled in readOnly', async function () {
```
RichardHelm, line 621:
```suggestion
		it('should set disabled in readOnly', async function () {
```

libs/components/src/lib/number-field/number-field.spec.ts
```diff
@@ -618,6 +618,20 @@ describe('vwc-number-field', () => {
...
+			element.readOnly = true;
+			await elementUpdated(element);
+			expect(addButton.hasAttribute('disabled')).toBe(true);
+			expect(subtractButton.hasAttribute('disabled')).toBe(true);
+		});
+
+		it('should set the disabled in disabled', async function () {
```
RichardHelm, line 628:
```suggestion
		it('should set disabled in disabled', async function () {
```

RichardHelm (approved):

PR #1785, opened 2024-06-19
feat(eslint-plugin): add no value attribute rules (VIV-1827)

RichardHelm:


libs/eslint-plugin/src/rules/no-current-value-attribute.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+		convertAnnotatedSourceToFailureCase({
+			annotatedSource: `
+				<template><VTextField :current-value="value" /></template>
+				                       ~~~~~~~~~~~~~
+			`,
+			fixedSource: `
+				<template><VTextField :model-value="value" /></template>
+			`,
+			message:
+				'Use `modelValue` instead of `currentValue` to set the current value.',
```
TaylorJ76, line 30: Should this say `current-value` instead of `currentValue`?
Same applies to the other ones below which have kebab case in the `annotatedSource`?

libs/eslint-plugin/src/rules/no-current-value-attribute.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+			`,
+			fixedSource: `
+				<template><VTextField model-value="value" /></template>
+			`,
+			message:
+				'Use `modelValue` instead of `currentValue` to set the current value.',
+		}),
+		convertAnnotatedSourceToFailureCase({
+			annotatedSource: `
+				<template><VCheckbox :current-checked="value" /></template>
```
TaylorJ76, line 56: in `VTextField` you checked for `:current-value`, `:currentValue` and `current-value`. 
Why is that not needed for the other components?

TaylorJ76 (changes requested):

libs/eslint-plugin/src/rules/no-current-value-attribute.spec.ts
RichardHelm: All attributes are handled the same, so I think it's enough to test once that all the variations are supported

libs/eslint-plugin/src/rules/no-current-value-attribute.spec.ts
RichardHelm: In my mind, `currentValue` is the name of the attribute. Otherwise it would also need to say `model-value`.
What do you think?

TaylorJ76 (approved):

PR #1792, opened 2024-06-24
fix(audio-player): keep paused state while dragging (VIV-1787)

YonatanKra:
- [x] Remove multiple event handlers and replace them with a single `change` handler on the slider
- [x] Update the slider's value from the class (single line of code instead of multiple + single source of truth)
- [x] Keep paused state while dragging (according to native Audio player)


RichardHelm (changes requested):
I found two issues when testing this:

1. The slider doesn't drag smoothly anymore

https://github.com/Vonage/vivid-3/assets/86777660/eead0399-3274-4c0e-b1e0-1032ea1fe9e3



2. Audio player should pause while dragging

YonatanKra:
> I found two issues when testing this:
> 
> 1. The slider doesn't drag smoothly anymore
> 
>  Screen.Recording.2024-06-24.at.16.18.10.mov 
> 2. Audio player should pause while dragging

Thanks.
1. This is also happening in the slider with no relation to the audio player (I did send an update that prevents changing the slider from two different locations).
2. This is the main reason for the fix. The native HTML audio keeps playing while dragging.

RichardHelm:
@YonatanKra Thanks, first issue is solved.

For the second, it depends on the implementation but most audio players (e.g. native `<audio>` in Chrome) pause the audio while being dragged and resume it afterwards. Otherwise you get choppy sound as it is dragged around

RichardHelm (approved):

PR #1804, opened 2024-06-26
fix(tabs)!: change gutters default to small (VIV-1796)

rachelbt:


libs/components/src/lib/tabs/tabs.template.ts
```diff
@@ -13,6 +13,7 @@ const getClasses = ({
 		[`connotation-${connotation}`, Boolean(connotation)],
 		[`orientation-${orientation}`, Boolean(orientation)],
 		[`gutters-${gutters}`, Boolean(gutters)],
+		['gutters-small', !gutters],
```
RichardHelm, line 16: Nit, but it will set the class twice.
```suggestion
		`gutters-${gutters ?? 'small'}`,
```

libs/components/src/lib/tabs/README.md
```diff
@@ -22,14 +22,14 @@ The vwc-tabs accepts [vwc-tab](/components/tab/) and `vwc-tab-panel` elements as
...
 ### Gutters
 
-Use the `gutters` attribute to add a margin to the component.
+Use the `gutters` attribute to control the tabs panel padding.
 
-- Type: `'small'`
+- Type: `'small'` | `'none'`
 
-- Default: `none`
+- Default: `small`
```
RichardHelm, line 29: Why does this not require updating any tests? Shouldn't it break screenshots

RichardHelm (changes requested):

libs/components/src/lib/tabs/tabs.template.ts
rachelbt: thanks!!!
Was wondering how to do that and didn't want to bother anyone :)

libs/components/src/lib/tabs/README.md
rachelbt: you are right - the first code should failed the ui tests. (in all the other we added in the code `gutters="small"`).
Updating

RichardHelm (approved):

PR #1813, opened 2024-06-30
chore(docs)!: refine docs for v-4 (VIV-000)

rachelbt:


libs/components/src/lib/tooltip/README.md
```diff
@@ -154,12 +154,6 @@ Use the `--tooltip-max-inline-size` variable to set the tooltip's max inline siz
 </vwc-tooltip>
 ```
 
-### Deprecated CSS Variable: Inline Size
-
-The `--tooltip-inline-size` CSS variable is deprecated (as of 05/24). Use `--tooltip-max-inline-size` instead to limit the inline size.
-`--tooltip-inline-size` is still functional in the component but will be removed in a future major release.
-This will be communicated when it's removal becomes a release candidate at the end of the support period.
-
```
YonatanKra, line 162: Why is this removed?

rachelbt: because it doesn't exists anymore.
In none of the other changes we didn't leave any deprecation notes.

YonatanKra (approved):

PR #1849, opened 2024-07-16
chore(docs): browser support page (VIV-1736)

TaylorJ76:


rachelbt (commented):
looks good.
One question - aren't we supporting 2 last versions in each browser?

TaylorJ76:
> looks good. One question - aren't we supporting 2 last versions in each browser?

Yes, I think that is probably best seeing as users don't always update their browsers immediately.

rachelbt (approved):
👍

PR #1853, opened 2024-07-17
feat(dialog): improve api with modal attribute (VIV-1657)

RichardHelm:
* Adds `modal` attribute
* Adds `open` event
* Adds `v-model` for open


libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -64,18 +55,12 @@ function renderDismissButton(buttonTag: string) {
 }
 
 function handleEscapeKey(dialog: Dialog, event: Event) {
-	if ((event as KeyboardEvent).key === 'Escape' && dialog.modal) {
+	if ((event as KeyboardEvent).key === 'Escape' && dialog._openedAsModal) {
 		dialog.open = false;
```
TaylorJ76, line 59: Should we stop the propagation of the event here to stop it affecting other components that are listening for an escape key event?
For example, if the modal markup was inside a side drawer component for some reason, escape would also close the side drawer as well as the modal.

TaylorJ76 (changes requested):

libs/components/src/lib/dialog/dialog.template.ts
RichardHelm: @TaylorJ76 Yes, but that would not really be related to the `modal` feature.
Also, we need to do it for all other components.
I created a new ticket for it: https://jira.vonage.com/browse/VIV-1950

TaylorJ76 (approved):

PR #1857, opened 2024-07-21
fix(file-picker): add background color to uploaded items in list (VIV-1823)

rachelbt:


libs/components/src/lib/file-picker/README.md
```diff
@@ -141,6 +141,13 @@ The `helper-text` slot allows you to use rich content as the file picker's helpe
 </vwc-file-picker>
 ```
 
+## CSS Variables
+
+### file-picker-list-item-background-color
+
+Each added file in the preview list has default background-color of `vvd-color-canvas` (`#fff` in light theme, `#000` in dark).  
+If needed the background if the item can be changed using the `--file-picker-list-item-background-color` css-variable.
```
RichardHelm, line 149:
```suggestion
Each added file in the preview list has a default background-color of `vvd-color-canvas` (`#fff` in light theme, `#000` in dark).  
If needed, the background of the item can be changed using the `--file-picker-list-item-background-color` CSS variable.
```

RichardHelm (approved):

PR #1863, opened 2024-07-24
feat(empty-state): adding connotation and design variation (VIV-1378)

rachelbt:


libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -21,21 +37,54 @@
 	align-items: center;
 	justify-content: center;
 	border-radius: #{border-radius-variables.$border-radius-full};
-	background-color: var(#{constants.$vvd-color-neutral-100});
+	background-color: color-mix(
+		in srgb,
+		var(#{appearance.get-appearance-token(outline)}),
+		var(#{constants.$vvd-color-canvas}) 80%
```
rachelbt, line 43:
```suggestion
		var(#{constants.$vvd-color-canvas}) 70%
```

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -21,21 +37,54 @@
...
+				);
+			border-radius: inherit;
+		}
+	}
 }
 
 .content {
 	display: flex;
 	flex-direction: column;
+	color: var(#{constants.$vvd-color-canvas});
```
rachelbt, line 79:
```suggestion
	color: var(#{constants.$vvd-color-canvas-text});
```

libs/components/src/lib/empty-state/empty-state.scss
```diff
@@ -1,12 +1,28 @@
...
+);
+@use '../../../../shared/src/lib/sass/mixins/appearance' as appearance;
 
 :host {
 	display: block;
 }
 
 .base {
+	@include connotation.connotation(badge);
+	@include appearance.appearance;
```
RichardHelm, line 24: IMO, it is not right to use appearance mixin here. Design didn't use appearance so why should the code? The mixed colors end up different from the design 

```suggestion
	@include connotation.connotation(empty-state);
```

libs/components/src/lib/empty-state/README.md
```diff
@@ -42,6 +42,112 @@ Use the `icon` attribute to set the icon of the empty state.
...
+	</vwc-empty-state>
+</div>
+```
+
+### Icon-design
+
+Use icon-design to change the design of the circle wrapping the icon
+
+- Type: `boolean`
+- Default: `undefined`
```
RichardHelm, line 101: It seems weird to have a boolean `icon-design`.
Maybe it should be an enum instead, like `graphic-appearance/design=outlined/filled`
Or rename to something like `outlined-icon`. However, I think enum is better to leave room for more options in the future

RichardHelm (changes requested):

libs/components/src/lib/empty-state/empty-state.scss
rachelbt: the badge is a huge mistake of mine and I'll fix it.

But we are always using connotation and appearance together - see alert & progress.
On the other hand we have icons that do not use it.
I wonder what is the right way to do that.
Do we want all components that has one appearance, no states to use connotation directly or use it via appearances...
If I'm not using connotation directly - I can use whatever shades I want, is it good for the design-system or should we align to appearances?
@AyalaBu @TaylorJ76 
WDYT?

rachelbt: I removed the appearance and created a ticket to align other component to the same convention: https://jira.vonage.com/browse/VIV-1960

libs/components/src/lib/empty-state/README.md
```diff
@@ -42,6 +42,120 @@ Use the `icon` attribute to set the icon of the empty state.
...
+		connotation="warning"
+	>
+		No results
+	</vwc-empty-state>
+</div>
+```
+
+### Icon-decoration
+
+Use graphic-design to change the design of the icon circle
```
RichardHelm, line 100:
```suggestion
Use icon-decoration to change the design of the icon circle.
```

libs/components/src/lib/empty-state/empty-state.ts
```diff
@@ -23,6 +57,15 @@ export class EmptyState extends FoundationElement {
 	 */
 	@attr icon?: string;
 
+	/**
+	 * option to a new design for the icon circle
+	 *
+	 * @public
+	 * HTML Attribute: icon-design
```
RichardHelm, line 64:
```suggestion
	 * HTML Attribute: icon-decoration
```

RichardHelm (changes requested):
Looks good! Just these outdated names

RichardHelm (approved):

PR #1869, opened 2024-08-01
feat(dialog, side-drawer): add cancel event (VIV-1899)

RichardHelm:


TaylorJ76 (dismissed):

libs/components/src/lib/dialog/README.md
```diff
@@ -402,10 +402,11 @@ The dialog has a default `--dialog-max-block-size`. If the content is larger, th
...
+| Name     | Type                     | Bubbles | Composed | Description                                                                                                                                                   |
+| -------- | ------------------------ | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `open`   | `CustomEvent<undefined>` | No      | Yes      | The `open` event fires when the dialog opens.                                                                                                                 |
+| `close`  | `CustomEvent<string>`    | No      | Yes      | The `close` event fires when the dialog closes (either via user interaction or via the API). It returns the return value inside the event's details property. |
+| `cancel` | `CustomEvent<undefined>` | No      | Yes      | The `cancel` event fires when the user requests to close the dialog. You can prevent the dialog from closing by calling `.preventDefault()` on the event.     |
```
YonatanKra, line 409:
```suggestion
| `cancel` | `CustomEvent<undefined>` | No      | Yes      | The `cancel` event fires when the user requests to close the dialog, before the dialog closes. You can prevent the dialog from closing by calling `.preventDefault()` on the event.     |
```

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -50,17 +50,10 @@ function renderDismissButton(buttonTag: string) {
 		size="condensed"
 		class="dismiss-button"
 		icon="close-line"
-		@click="${(x) => (x.open = false)}"
+		@click="${(x) => x._handleCloseRequest()}"
```
YonatanKra, line 53: `_handleCloseRequest` does not belong on the class.
Create a local function in the template file and use it instead of polluting the class itself.

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -72,7 +65,7 @@ export const DialogTemplate: (
 	return html<Dialog>`
 	<${elevationTag} dp="8">
 		<dialog class="${getClasses}"
-				@keydown="${(x, c) => handleEscapeKey(x, c.event)}"
+				@keydown="${(x, c) => x._onKeyDown(c.event as KeyboardEvent)}"
 				@cancel="${(_, c) => c.event.preventDefault()}"
 				aria-label="${(x) => x.ariaLabel}"
 				?aria-modal="${(x) => x._openedAsModal}"
```
YonatanKra, line 71: Why do we need `_openAsModal` if we already have an exposed API `modal`?

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -72,7 +65,7 @@ export const DialogTemplate: (
 	return html<Dialog>`
 	<${elevationTag} dp="8">
 		<dialog class="${getClasses}"
-				@keydown="${(x, c) => handleEscapeKey(x, c.event)}"
+				@keydown="${(x, c) => x._onKeyDown(c.event as KeyboardEvent)}"
```
YonatanKra, line 68: Please remove `_onKeyDown` from the class and create a local template function.

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -157,6 +160,36 @@ export class Dialog extends FoundationElement {
...
+	 * @internal
+	 */
+	_onKeyDown(event: KeyboardEvent) {
+		if ((event as KeyboardEvent).key === 'Escape' && this._openedAsModal) {
+			this._handleCloseRequest();
+
+			// Return false to .preventDefault() which will prevent the <dialog>'s cancel event from being fired.
+			// Otherwise, pressing ESC twice would close the <dialog> without the ability to prevent it.
+			// This is because subsequent close requests without "intervening user action" between them are not cancelable.
+			// See: https://html.spec.whatwg.org/multipage/interaction.html#close-watcher-infrastructure
```
YonatanKra, line 173: Can this be tested?

libs/components/src/lib/dialog/dialog.ts
```diff
@@ -157,6 +160,36 @@ export class Dialog extends FoundationElement {
...
+			return false;
+		}
+		return true;
+	}
+
+	/**
+	 * @internal
+	 */
+	_handleCloseRequest() {
```
YonatanKra, line 182: See my comment regarding the removal of this method.

libs/components/src/lib/side-drawer/README.md
```diff
@@ -134,10 +134,11 @@ The `app-content` slot sets assigned nodes to the main application content, the
...
-| `close` | `CustomEvent<undefined>` | No      | Yes      | Fired when the menu is closed |
+| Name     | Type                     | Bubbles | Composed | Description                                                                                                                                      |
+| -------- | ------------------------ | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `open`   | `CustomEvent<undefined>` | No      | Yes      | Fired when the side drawer is opened.                                                                                                            |
+| `close`  | `CustomEvent<undefined>` | No      | Yes      | Fired when the side drawer is closed.                                                                                                            |
+| `cancel` | `CustomEvent<undefined>` | No      | Yes      | Fired when the user requests to close the side-drawer. You can prevent the side drawer from closing by calling `.preventDefault()` on the event. |
```
YonatanKra, line 141:
```suggestion
| `cancel` | `CustomEvent<undefined>` | No      | Yes      | Fired when the user requests to close the side-drawer, before it is closed. You can prevent the side drawer from closing by calling `.preventDefault()` on the event. |
```

libs/components/src/lib/side-drawer/side-drawer.template.ts
```diff
@@ -29,7 +29,7 @@ export const sideDrawerTemplate: FoundationElementTemplate<
 		class="${getClasses}"
 		?inert="${(x) => !x.open}"
 		part="base ${(x) => (x.alternate ? 'vvd-theme-alternate' : '')}"
-		@keydown="${(x, c) => handleKeydown(x, c.event as KeyboardEvent)}"
+		@keydown="${(x, c) => x._onKeydown(c.event as KeyboardEvent)}"
```
YonatanKra, line 32: See my remarks in the dialog

libs/components/src/lib/side-drawer/side-drawer.template.ts
```diff
@@ -42,18 +42,7 @@ export const sideDrawerTemplate: FoundationElementTemplate<
 		(x) => x.modal,
 		html<SideDrawer>`<div
 			class="${getScrimClasses}"
-			@click="${(x) => (x.open = false)}"
+			@click="${(x) => x._handleCloseRequest()}"
```
YonatanKra, line 45: See my remarks in the dialog

YonatanKra (changes requested):
Thanks for that. 
Some minor changes

libs/components/src/lib/dialog/README.md
RichardHelm: I don't think this is relevant or adds anything

libs/components/src/lib/dialog/dialog.ts
RichardHelm: I have added a UI test for correct behaviour now
However, it passes anyway because this behaviour is a recent addition to the spec. It is not yet implemented in the browser versions that we test with

YonatanKra: I'm unable to reproduce this behavior even when I remove this line. Here are the steps I've taken:
1. Create a modal dialog
2. Add event listener to the `cancel` event and prevent default
3. Open the modal
4. Hit escape multiple times
5. Dialog remains open
Is this the flow that's supposed to reproduce this use case?

libs/components/src/lib/side-drawer/side-drawer.ts
```diff
@@ -70,4 +71,33 @@ export class SideDrawer extends FoundationElement {
...
+	 * @internal
+	 */
+	_onKeydown({ key }: KeyboardEvent) {
+		if (key === 'Escape') {
+			this._handleCloseRequest();
+			return undefined;
+		} else {
+			// after this event handler is executed,
+			// preventDefault() will be called on the event object by default.
+			// we need to return true from our handler to opt - out of this behavior.
```
YonatanKra, line 85: How is this helping? What scenario does this intend to fix?

YonatanKra (changes requested):

libs/components/src/lib/dialog/dialog.ts
RichardHelm: Yes, that's right. It will close the dialog in step 5 if you use a browser that implements the spec, e.g. Chrome

libs/components/src/lib/side-drawer/side-drawer.ts
RichardHelm: I have only moved this code, but it will not call preventDefault on all keydown events in side drawer

libs/components/src/lib/dialog/dialog.template.ts
YonatanKra: How would it break backwards compatibility?
When was the `modal` added?

YonatanKra (dismissed):
Mostly the tests comments.

libs/components/src/lib/dialog/dialog.template.ts
RichardHelm: If `showModal()` would set `modal` attribute, behaviour of code like this would change:
```
x.showModal()
x.close()
x.open = true; // will be modal now but wasn't before
```

libs/components/src/lib/dialog/dialog.template.ts
RichardHelm: As with the others, I think all logic should go into the class

libs/components/src/lib/dialog/dialog.template.ts
YonatanKra: We talked about this privately. Let's move to `#` instead of `_` to prevent such cases from slipping to production.

libs/components/src/lib/dialog/dialog.template.ts
YonatanKra: After we saw how things can go wrong in the Dialog, I think we can safely revert to our own coding standards and remove the `_`. WDYT?

TaylorJ76 (approved):

PR #1871, opened 2024-08-02
feat(dialog): add `no-dismiss-on-esc`, `no-dismiss-button` and `non-dismissible` members (VIV-1907)

RichardHelm:


libs/components/src/lib/dialog/README.md
```diff
@@ -149,6 +149,51 @@ Use the `no-light-dismiss` attribute to prevent a modal dialog from being dismis
...
+
+```html preview 230px
+<vwc-button
+	label="Open modal dialog"
+	onclick="document.querySelector('vwc-dialog').open = true"
+></vwc-button>
+<vwc-dialog no-dismiss-on-esc headline="Headline" modal></vwc-dialog>
+```
+
+### No-dismiss-button
```
YonatanKra, line 167: Is this valid a11y-wise? Shouldn't we add a comment saying consumer must add some manner of closing the button?

libs/components/src/lib/dialog/README.md
```diff
@@ -149,6 +149,51 @@ Use the `no-light-dismiss` attribute to prevent a modal dialog from being dismis
 <vwc-dialog no-light-dismiss headline="Headline" modal></vwc-dialog>
 ```
 
+### No-dismiss-on-esc
```
YonatanKra, line 152: Is this valid a11y-wise? How would a keyboard-user close the dialog? Maybe we can add a comment about a keyboard-accessible close button must be present?

libs/components/src/lib/dialog/README.md
```diff
@@ -149,6 +149,51 @@ Use the `no-light-dismiss` attribute to prevent a modal dialog from being dismis
...
+
+```html preview 230px
+<vwc-button
+	label="Open modal dialog"
+	onclick="document.querySelector('vwc-dialog').open = true"
+></vwc-button>
+<vwc-dialog no-dismiss-button headline="Headline" modal></vwc-dialog>
+```
+
+### Non-dismissible
```
YonatanKra, line 182: As with the others, we'd might want to add instructions about this in the a11y section of the dialog component and add a reference to it with each such property.

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -149,6 +148,24 @@ describe('vwc-dialog', () => {
 		});
 	});
 
+	describe('no-dismiss-button', function () {
+		it('should render a dismiss button when no-dismiss-button is not set', async function () {
```
YonatanKra, line 152:
```suggestion
		it('should render a dismiss button when no-dismiss-button falsy', async function () {
```

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -149,6 +148,24 @@ describe('vwc-dialog', () => {
...
+			expect(getDismissButton()).not.toBeNull();
+		});
+
+		it('should not render dismiss button when no-dismiss-button is set', async function () {
```
YonatanKra, line 156:
```suggestion
		it('should remove dismiss button when no-dismiss-button is true', async function () {
```

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -149,6 +148,24 @@ describe('vwc-dialog', () => {
...
+			element.noDismissButton = true;
+			await elementUpdated(element);
+			expect(getDismissButton()).toBeNull();
+		});
+
+		it('should not render dismiss button when no-dismiss-button is implicitly set via non-dismissible', async function () {
```
YonatanKra, line 162:
```suggestion
		it('should remove dismiss button when non-dismissible is true', async function () {
```

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -430,6 +447,16 @@ describe('vwc-dialog', () => {
 			expect(element.open).toEqual(true);
 		});
 
+		it('should leave the dialog open on scrim click when no light dismiss is implicitly set via non-dismissible', async function () {
```
YonatanKra, line 450:
```suggestion
		it('should leave the dialog open on scrim click when non-dismissible is true', async function () {
```

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -492,7 +519,7 @@ describe('vwc-dialog', () => {
 		element.addEventListener('close', spy);
 		await showDialog();
 
-		clickDismissButton();
+		getDismissButton().click();
```
YonatanKra, line 522: Why not just change `clickDismissButton` method instead of replacing it in multiple places?

libs/components/src/lib/dialog/dialog.spec.ts
```diff
@@ -602,6 +629,20 @@ describe('vwc-dialog', () => {
...
 		});
 
+		it('should stay open on escape key press when no-dismiss-on-esc is set', async function () {
+			element.noDismissOnEsc = true;
+			await showModalDialog();
+			await triggerEscapeKey();
+			expect(element.open).toEqual(true);
+		});
+
+		it('should stay open on escape key press when no-dismiss-on-esc is implicitly set via non-dismissible', async function () {
```
YonatanKra, line 639:
```suggestion
		it('should stay open on escape key press when non-dismissible is true', async function () {
```

libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -78,7 +78,7 @@ export const DialogTemplate: (
 							</slot>
 							${when((x) => x.headline, headline())}
 							${when((x) => x.subtitle, subtitle())}
-							${renderDismissButton(buttonTag)}
+							${when((x) => x._showDismissButton, renderDismissButton(buttonTag))}
```
YonatanKra, line 81: Please replace the `_showDismissButton` getter with a template local function that does the same logic of 2 public API's.

libs/components/src/lib/dialog/README.md
RichardHelm: Yes, they can close it in the same way a mouse user would

libs/components/src/lib/dialog/dialog.spec.ts
RichardHelm: It's too simple to be it's own function

libs/components/src/lib/dialog/dialog.template.ts
RichardHelm: As discussed, I don't see a problem with this

libs/components/src/lib/dialog/dialog.spec.ts
RichardHelm: I don't understand why you want to remove the word "not" from all the messages. I think it is clearer this way

libs/components/src/lib/dialog/dialog.spec.ts
RichardHelm: I have phrased it in this way because it is otherwise illogical to be in the `no-dismiss-button` section

TaylorJ76 (approved):

PR #1873, opened 2024-08-05
chore(docs): move content to docs (VIV-000)

RichardHelm:
Use eleventy more idiomatically and remove unnecessary complexity from the docs

- Move non-component docs into the docs app
- JSON data files are no longer needed for them and are removed
- Eleventy navigation plugin is used to build the navigation


rachelbt (commented):
Nice! So much nicer now.
2 things:
1. the resources page changed its place. It used to be under the components but now they are before the icon gallery. Not sure its a big deal if its an issue to fix.
2. the appearance ui are failing

RichardHelm:
@rachelbt 
1. Yes, I moved it on purpose as it doesn't really make sense at the bottom
2. Fixed it

rachelbt (approved):

PR #1875, opened 2024-08-07
feat(components): support tree shaking (VIV-1672)

RichardHelm:
Adds `sideEffects` property to package.json which allows bundlers to tree-shake unused components.
Example bundle with just button being used goes from `1.35 MiB` to `95 KiB`.

Vivid component code is itself is considered by bundlers to have side-effects, which is why it is not tree-shakable out of the box. This property allows bundlers to skip going into the component files in the first place, so they don't get included anymore.


YonatanKra (approved):
Nice work!

PR #1882, opened 2024-08-14
feat(searchable-select): add component (VIV-1793)

RichardHelm:


TaylorJ76 (commented):
Margin between label and control is smaller than other form elements.

RichardHelm:
@TaylorJ76 You're right, fixed

TaylorJ76 (changes requested):
When using a keyboard to navigate the listbox options, when I key down to an item that is outside of the visible scrollable area, it doesn't scroll into view.
As a result you loose track of which item you are focussed on.

libs/components/src/lib/searchable-select/option-tag.scss
```diff
@@ -0,0 +1,89 @@
...
+	border-radius-variables;
+@use '../../../../shared/src/lib/sass/mixins/connotation/config' with (
+	$connotations: accent,
+	$shades: contrast soft pale fierce firm-all faint dim,
+	$default: accent
+);
+@use '../../../../shared/src/lib/sass/mixins/connotation' as connotation;
+@use '../../../../shared/src/lib/sass/mixins/appearance/config' as
+	appearance-config with (
+	$appearances: subtle-light,
```
rachelbt, line 14: I think this is not the right place to create a new appearance.
This needs to be added properly - with all the states and not only the ones that are needed here.
I suggest we use suble and once we have the one with opacity we will use it.

libs/components/src/lib/searchable-select/option-tag.template.ts
```diff
@@ -0,0 +1,52 @@
+import { html, when } from '@microsoft/fast-element';
```
rachelbt, line 1: why can't we use vwc-tag?

TaylorJ76 (commented):
It would also be good if the popup with menu didn't ever exceed the available visible viewport.
So, it would adapt it's height if, when populated, it exceed the viewport.

libs/components/src/lib/searchable-select/option-tag.template.ts
RichardHelm: It has different semantics. Tag is an option while the option-tag is text + remove button. Also, the whole tag appears interactive unlike option-tag where only remove button is clickable. It also does not have `light` appearance.

libs/components/src/lib/searchable-select/option-tag.scss
RichardHelm: Sure, I've removed the appearance and created a ticket for it: https://jira.vonage.com/browse/VIV-2161

RichardHelm:
@TaylorJ76 
> When using a keyboard to navigate the listbox options, when I key down to an item that is outside of the visible scrollable area, it doesn't scroll into view.

Makes sense. I have created a new ticket for this as well to keep the PR focussed: https://jira.vonage.com/browse/VIV-2162

> It would also be good if the popup with menu didn't ever exceed the available visible viewport.
It makes sense but the current behaviour is the same as `vwc-select` etc. So it should be fixed for all of them. I have created a ticket for it: https://jira.vonage.com/browse/VIV-2163

libs/components/src/lib/searchable-select/searchable-select.template.ts
```diff
@@ -0,0 +1,239 @@
...
+							(x) => x._inputValue !== '' && x._filteredOptions.length === 0,
+							html<SearchableSelect>`<div class="empty-message">
+								<slot name="no-matches"
+									>${(x) => x.locale.searchableSelect.noMatchesMessage}</slot
+								>
+							</div>`
+						)}
+						${when(
+							(x) => x._inputValue === '' && x._slottedOptions.length === 0,
+							html<SearchableSelect>`<div class="empty-message">
```
TaylorJ76, line 213: Is there a way this can be optimised by removing the duplication with the *no matches* code block above?

TaylorJ76 (changes requested):

TaylorJ76:
Great stuff @RichardHelm!

PR #1883, opened 2024-08-19
fix(dialog): stop escape press propagation (VIV-1950)

YonatanKra:


apps/docs/assets/scripts/icons-gallery/icons-gallery.js
```diff
@@ -125,10 +125,6 @@ function filterIcons() {
 		)
 	);
 
-	iconsArray = Array.from(
-		new Map(iconsArray.map((item) => [item.id, item])).values()
-	);
-
```
TaylorJ76, line 131: Wasn't this added recently?

YonatanKra: I have no idea how this got there... only shows not having tests on the docs has become a problem :)

libs/components/src/shared/date-picker/date-picker-base.ts
```diff
@@ -315,7 +316,7 @@ export abstract class DatePickerBase extends FormAssociatedDatePickerBase {
 	 */
 	_onBaseKeyDown(event: KeyboardEvent) {
 		// Close dialog on Escape
-		if (event.key === 'Escape') {
+		if (handleEscapeKeyAndStopPropogation(event)) {
```
TaylorJ76, line 319: This needs to only stop propogation when the popup is open.

TaylorJ76 (changes requested):

TaylorJ76 (approved):

PR #1891, opened 2024-09-02
chore(docs): style refresh (VIV-000)

TaylorJ76:
- Removes gradients from headings (looks less stripey and is easier to read)
- Set a max width on paragraphs and lists inside of the main article area (makes them easier to read)
- Increases the font-size of the first paragraph that directly follows the H1 in article (acts like a quick introduction to the page)
- Takes the gutter off tabs (V4 upgrade task)
- Aligns main content to the left (centering makes it look disconnected from navigation on large screensizes)
- Changes border-radius from 6px to 8px
- Remove unnecessary bullet points
- Homepage: Features now on one line of larger screensizes
- Removes elevation from roadmap table


YonatanKra:
Can you add some explanation as to what is the reason for this change and what changed?

apps/docs/assets/styles/main.scss
```diff
@@ -258,33 +278,13 @@ vwc-note ~ vwc-note {
...
-	margin-block-end: 0;
-}
-
 /* #endregion article */
 
 /* #startregion table */
 
 .table-wrapper {
 	display: inline-flex;
 	box-sizing: border-box;
```
rachelbt, line 287: I would keep the block-padding because the table is too close now to the border:
image

rachelbt: also not related to this but can you change this line (293):
max-width: calc(100vw - 32px);
to 
max-width: calc(100vw - 64px);

apps/docs/assets/styles/main.scss
```diff
@@ -227,11 +249,8 @@ article.article > h1 {
 }
 
 .heading-wrapper > h2 {
-	background-image: linear-gradient(
-		to right,
-		var(--vvd-color-cta-50),
-		rgb(255 0 0 / 0%)
-	);
+	border-block-end: 1px solid var(--vvd-color-neutral-200);
+	margin-block-end: 0;
```
rachelbt, line 253:
```suggestion
	margin-block-end: 0;
	margin-block-start: 72px;
```
for me it feels like there are too many lines, I added more space:
image
@AyalaBu WDYT?

rachelbt (commented):
Nice.
Left some small comments :)

rachelbt (approved):
LGTM

PR #1899, opened 2024-09-04
feat(text-field): add scale attribute with condensed (VIV-1813)

YonatanKra:


libs/components/src/lib/text-field/text-field.scss
```diff
@@ -44,6 +44,25 @@ $low-ink-color: --_low-ink-color;
...
+	&.size-condensed {
+		#{variables.$text-field-gutter-start}: calc(
+			#{size.$vvd-size-condensed} / 2.5
+		); // default of 16px
+		#{variables.$text-field-gutter-end}: calc(
+			#{size.$vvd-size-condensed} / 2.5
+		); // default of 16px
+		#{variables.$text-field-icon-size}: calc(
+			#{size.$vvd-size-condensed} / 2
+		); // default of 20px
```
rachelbt, line 64: the values that are in the comments are wrong.
 They actually defaults to 12.8px which is bad. I change the calculation to give the value of 12px

for the last one - the font size is 16px. Is this what needed?
```suggestion
		#{variables.$text-field-gutter-start}: calc(
			#{size.$vvd-size-super-condensed} / 2
		); // default of 12px
		#{variables.$text-field-gutter-end}: calc(
			#{size.$vvd-size-super-condensed} / 2
		); // default of 12px
		#{variables.$text-field-icon-size}: calc(
			#{size.$vvd-size-condensed} / 2
		); // default of 16px
```

libs/components/src/lib/text-field/text-field.scss
```diff
@@ -44,6 +44,25 @@ $low-ink-color: --_low-ink-color;
 		#{variables.$text-field-icon-size}: calc(
 			#{size.$vvd-size-normal} / 2
 		); // default of 20px
+		#{variables.$text-field-block-size}: #{size.$vvd-size-normal};
+
+		display: inline-grid;
+		width: 100%;
+		gap: 4px;
+		grid-template-columns: min-content 1fr max-content;
```
rachelbt, line 52: my comments is not for this lines - but somehow - these are marked as added.
under` &.condensed `you need to remove all these 4 lines (lines 67-70)

libs/components/src/lib/text-field/text-field.scss
rachelbt: we actually keep the convention of using :not(XXX) and &.XXX
so you actually need to add the code line this:
```
& {
		display: inline-grid;
		width: 100%;
		gap: 4px;
		grid-template-columns: min-content 1fr max-content;
	}

	&:not(.size-condensed ) {
		#{variables.$text-field-gutter-start}: calc(
			#{size.$vvd-size-normal} / 2.5
		); // default of 16px
		#{variables.$text-field-gutter-end}: calc(
			#{size.$vvd-size-normal} / 2.5
		); // default of 16px
		#{variables.$text-field-icon-size}: calc(
			#{size.$vvd-size-normal} / 2
		); // default of 20px
	}

	&.size-condensed  {
		#{variables.$text-field-gutter-start}: calc(
			#{size.$vvd-size-super-condensed} / 2
		); // default of 12px
		#{variables.$text-field-gutter-end}: calc(
			#{size.$vvd-size-super-condensed} / 2
		); // default of 12px
		#{variables.$text-field-icon-size}: calc(
			#{size.$vvd-size-condensed} / 2
		); // default of 16px
	}
```

libs/components/src/lib/text-field/README.md
```diff
@@ -122,6 +122,17 @@ Use the `icon` attribute to add an icon.
 <vwc-text-field icon="search-line" label="Search..."></vwc-text-field>
 ```
 
+### Scale
+
+Use the `scale` attribute to change the text field's size.
```
rachelbt, line 127: I wonder if we need to write something about not using the size property. WDYT?

rachelbt (commented):
added some comments.

1. I'm not sure about the font size of the icon in condensed. In my suggested changes it renders as 16px.
2. in other components - when the block size is 32px the border-radius is 4 px and not 8px. @AyalaBu is it true? 
@YonatanKra if this need to be changed - we can do this together :) 
other then that all is OK 👍

AyalaBu:
_32px the border-radius is 4 px_  - Correct

rachelbt (approved):
LGTM

PR #1901, opened 2024-09-05
feat(select): add scale attribute with condensed (VIV-1836)

YonatanKra:


libs/components/src/lib/select/partials/_variables.scss
```diff
@@ -1,3 +1,4 @@
 $select-height: --select-height;
 $select-control-border-radius: --_select-control-border-radius;
 $select-fixed-width: --_select-fixed-width;
+$select-icon-size: --_select-icon-size;
```
rachelbt, line 4: lets add css-variable like we have in text-field:
-block-size
- gutters (I think we don't need start+end but they are the same here in select) - for `inline-padding`
- icons-size

libs/components/src/lib/select/select.scss
```diff
@@ -65,6 +67,16 @@ $low-ink-color: --_low-ink-color;
 		transition: box-shadow 0.2s, background-color 0.2s;
 	}
```
rachelbt, line 69: please add the code same as we did in text-field, change the block-size & padding to use css-var and also:
```
&:not(.size-condensed) {all the relevant vars}
&.size-condensed {all the relevant vars}
```

libs/components/src/lib/select/select.scss
```diff
@@ -65,6 +67,16 @@ $low-ink-color: --_low-ink-color;
...
 
+	&.size-condensed {
+		#{variables.$select-icon-size}: 16px;
+
+		block-size: #{size.$vvd-size-condensed};
+
+		& .chevron {
+			font-size: 12px;
```
rachelbt, line 76: I think is way too small. We can keep it 16px. @AyalaBu ?

libs/components/src/lib/select/select.scss
```diff
@@ -124,8 +136,8 @@ $low-ink-color: --_low-ink-color;
 	}
 
 	slot[name='icon'] {
-		flex: 0 0 20px;
-		font-size: 20px;
+		flex: 0 0 var(#{variables.$select-icon-size});
+		font-size: var(#{variables.$select-icon-size});
```
rachelbt, line 140: 👍

libs/components/src/lib/select/select.scss
```diff
@@ -65,6 +67,16 @@ $low-ink-color: --_low-ink-color;
 		transition: box-shadow 0.2s, background-color 0.2s;
 	}
 
+	&.size-condensed {
+		#{variables.$select-icon-size}: 16px;
```
rachelbt, line 71:
```suggestion
		#{variables.$select-icon-size}: calc(
			#{size.$vvd-size-normal} / 2.5
		); // default of 16px
```

rachelbt: and please set the normal one the same way but divide it into 2 to get 20px.

rachelbt (commented):
added some comments to align the css to be aligned to what we have in text-field :)

libs/components/src/lib/select/select.scss
YonatanKra: Again, I raise the issue here that if someone's going to use a different density (let's say -1) it will be:
32px / 2.5 = 12.8

rachelbt: you are right - but we have density, and we use it.
Fixing the density or removing that option is a totally different issue.
I'm suggesting that if text-field is using the calculation for semi-dynamic scale then I think we need to use them here too.

libs/components/src/lib/select/select.scss
rachelbt: what about this part?

YonatanKra: what part?

libs/components/src/lib/select/select.scss
```diff
@@ -47,24 +47,37 @@ $low-ink-color: --_low-ink-color;
 }
 
 .control {
+	border-radius: var(#{variables.$select-control-border-radius});
+	background-color: var(#{appearance.get-appearance-token(fill)});
+	block-size: var(#{variables.$select-block-size});
+	padding-inline: var(#{variables.$select-padding-inline});
+
```
rachelbt, line 54: this needs to be inside the first &{}.
We prefer the @includes to be the first of the class.

rachelbt (approved):
👍 
I'm approving this - but I think that we might need to add condensed to option as well
image
@AyalaBu ?

AyalaBu:
I absolutely agree about the condensed option
It is needed to be added

PR #1907, opened 2024-09-17
fix(selectable box): screen-reader accessibility when box is clickable (VIV-000)

TaylorJ76:
When `clickable-box` is enabled, screen readers should announcement the control as a "toggle button".
This only happened when the box is checked. The reason is that the control had `aria-pressed="true"` set when it is checked but it is set to null when it is not checked, it should have been set to `false`.


rachelbt (approved):
👍

PR #1911, opened 2024-09-18
fix(dial-pad): fix input, blur, change and focus events leak (VIV-2065)

YonatanKra:
Events now fire only once.
Delete button clicks fire the change and input events.
Known issues (VIV-2125):
- When focused on delete button, typing doesnt work
- When focused on a dial pad button, typing doesnt work
- Consider changing the keypad-click event's detail to give the value and not the button
- Deleting the last value using the delete button fires the Focus of type blur event
- Focus event should not fire for clicking on the call button


libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -164,40 +164,6 @@ export class DialPad extends FoundationElement {
 	_deleteLastCharacter = () => {
 		this.value = this.value.slice(0, -1);
 	};
-
-	/**
-	 *
-	 * @internal
-	 */
-	_handleInput = () => {
-		this.value = this._textFieldEl.value;
```
RichardHelm, line 173: This is no longer happening

libs/components/src/lib/dial-pad/dial-pad.template.ts
```diff
@@ -48,10 +56,10 @@ function renderTextField(textFieldTag: string, buttonTag: string) {
...
             aria-label="${(x) => x.locale.dialPad.inputLabel}"
             @keydown="${(x, c) => handleKeyDown(x, c.event as KeyboardEvent)}"
-            @input="${(x) => x._handleInput()}" @change="${(x) =>
-		x._handleChange()}"
-            @blur="${(x) => x._handleBlur()}" @focus="${(x) =>
-		x._handleFocus()}">
+            @input="${eventHandlerFactory('input')}" 
+			@change="${eventHandlerFactory('change')}"
+            @blur="${eventHandlerFactory('blur')}"
+			@focus="${eventHandlerFactory('focus')}">
```
RichardHelm, line 62: Blur and focus events are still emitted twice
I think `.stopImmediatePropagation()` has no effect as the events do not bubble in first the place. There is actually no need to do anything as the component emits `focus` and `blur` automatically if a child receives focus.
This is however incorrectly documented as `Emitted when the text field receives focus`

libs/components/src/lib/dial-pad/dial-pad.template.ts
```diff
@@ -48,10 +56,10 @@ function renderTextField(textFieldTag: string, buttonTag: string) {
 		x.helperText}" pattern="${(x) => x.pattern}"
             aria-label="${(x) => x.locale.dialPad.inputLabel}"
             @keydown="${(x, c) => handleKeyDown(x, c.event as KeyboardEvent)}"
-            @input="${(x) => x._handleInput()}" @change="${(x) =>
-		x._handleChange()}"
-            @blur="${(x) => x._handleBlur()}" @focus="${(x) =>
-		x._handleFocus()}">
+            @input="${eventHandlerFactory('input')}" 
+			@change="${eventHandlerFactory('change')}"
```
RichardHelm, line 60: The events are not emitted when the delete button is clicked
The change event is still emitted twice for some reason

RichardHelm (changes requested):

libs/components/src/lib/dial-pad/dial-pad.ts
YonatanKra: Nice catch. It was not tested :(
It is now :)

libs/components/src/lib/dial-pad/dial-pad.template.ts
YonatanKra: I can see the problem now.
It stems from the fact we have a hidden knowledge in a `_onDigit` that fires the `change` event also on click...
I saw a lot of opporunities for a refactor here and missing tests.

libs/components/src/lib/dial-pad/dial-pad.template.ts
YonatanKra: Regardig change:
It stems from the fact we have a hidden knowledge in a `_onDigit` that fires the `change` event also on click in certain areas.
Not tested (hence not documented and not in the template file so far away from the actual event listene)... now it will be. Thanks.

YonatanKra: The delete button now works as expected

libs/components/src/lib/dial-pad/dial-pad.template.ts
YonatanKra: Yes, you are correct. Every click anywhere in the dialog emits a focus.
The problem is that our `text-field` emits an event that bubbles despit it being set as `bubbles: false`
Will research further in the morning

YonatanKra: I actually think the author of the component meant for the focus and blur to be fired only on the `text-field`. Hence this is a bug. We need to stop all other focus events and send an event only from `text-field`. WDYT?

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -163,41 +163,8 @@ export class DialPad extends FoundationElement {
...
-		this.$emit('input');
-	};
-
-	/**
-	 *
-	 * @internal
-	 */
-	_handleChange = () => {
-		this.value = this._textFieldEl.value;
 		this.$emit('change');
```
RichardHelm, line 166: It should emit 'input' as well right?

libs/components/src/lib/dial-pad/dial-pad.template.ts
```diff
@@ -39,6 +39,20 @@ function handleKeyDown(x: DialPad, e: KeyboardEvent) {
...
+		x.value = x._textFieldEl.value;
+		e?.stopImmediatePropagation();
+		x.$emit(eventName);
+		return false;
+	};
+}
+
+function stopPropagation(_: DialPad, { event: e }: ExecutionContext) {
+	e.stopImmediatePropagation();
+	_.$emit('Blah');
```
RichardHelm, line 48:
```suggestion
```

libs/components/src/lib/dial-pad/dial-pad.template.ts
RichardHelm: Well, it may be intended that way but the dial-pad becomes the focused element in the main document when any child receives focus. When it becomes focused it should emit a `focus` event like all elements
Maybe the text-field events should be forwarded as `text-field-focus/blur`?

RichardHelm (changes requested):

libs/components/src/lib/dial-pad/dial-pad.ts
YonatanKra: Handled in a different PR.

libs/components/src/lib/dial-pad/dial-pad.template.ts
YonatanKra: I don't know if there's a difference. We need to decide if we want to the text-field to be the focus here. The buttons serve as a different kind of text-field, and eventually, affect it much like typing does.
Moreover, tabbing works just as expected - originating from the focused dial pad button and moving to the next/previous one.
Do we want a dedicated event for text-field? It's doable. Just need to decide. Anyway, it's not part of this PR.

RichardHelm (approved):

PR #1912, opened 2024-09-19
fix(dial-pad): prevent focus loss after removal of delete button (VIV-2126)

YonatanKra:
1. Add blur listener to the dial pad
2. Type 1 number
3. Delete number using the backspace button in the text-field
4. See that focus remains inside the dial pad


libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -163,40 +163,26 @@ export class DialPad extends FoundationElement {
...
-	 */
-	_handleChange = () => {
-		this.value = this._textFieldEl.value;
 		this.$emit('change');
+		if (this.value === '') {
+			document.addEventListener(
+				'blur',
+				this.#blurHandlerAfterDeleteButtonRemoved,
+				true
+			);
```
RichardHelm, line 173: Why would this be needed?
If there is actually a reason for it needs to be explained in a comment

```suggestion
			this._textFieldEl.focus();
```

RichardHelm (changes requested):

libs/components/src/lib/dial-pad/dial-pad.ts
YonatanKra: 1. We need to prevent the focus event from firing
2. This is tested
3. I've wrapped the logic in a function with a descriptive name.

RichardHelm: I don't understand. The component will not emit focus (or blur) when focus moves within, so this should not be necessary.

YonatanKra: See the related issue.
When pressing the delete button (in the UI) to delete the last character in the input, it fires a blur event and the dial pad loses focus. This makes sure it remains that way. 
See the descriptions:
image

YonatanKra: This solves the problem of focus leaving the component after deleting the last character. Focus does not remain inside the component hence the event being fired.

RichardHelm: Yes I understand that. But how does the suggested change not resolve the issue? Focus moves to text-field, therefore no blur (or focus) will be emitted on the dial-pad

YonatanKra: Oh, I see what you mean. 
The focus will happen before the event dispatches.
Nice one, thanks!

RichardHelm (approved):

PR #1915, opened 2024-09-24
feat(tab): adds closable attribute and close event (VIV-2030)

TaylorJ76:


libs/components/src/lib/tab/tab.template.ts
```diff
@@ -34,15 +50,20 @@ const getClasses = ({
  */
 export function TabTemplate<T extends Tab>(context: ElementDefinitionContext) {
 	const affixIconTemplate = affixIconTemplateFactory(context);
+	const buttonTag = context.tagFor(Button);
```
RichardHelm, line 53: Button registries need to be added to `tabRegistries` in `definition.ts` so that Button will registered as well when Tab is registered

libs/components/src/lib/tab/tab.template.ts
```diff
@@ -23,9 +25,23 @@ const getClasses = ({
...
 		['selected', ariaSelected === 'true'],
-		['icon-trailing', iconTrailing]
+		['icon-trailing', iconTrailing],
+		['removable', removable]
 	);
 
+function renderDismissButton(buttonTag: string) {
+	return html<Tab>`
+		<${buttonTag}
+			aria-label=""
```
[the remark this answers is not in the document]
rachelbt: also - the tab now is announced as group and referring the button as - one more item but no way to know what it is :(

rachelbt (changes requested):
didn't went over all the code yet, tested keyboad navigation and screen-reader and its not 100%.
When tab is removed (also - there's no way to know it is removable), the focus disappears, shouldn't it go the the latest tab?
Also - if all tabs are deleted it creates a strange use case:
image
maybe we need to add that if only one is left it can't be removed?

libs/components/src/lib/tab/tab.scss
```diff
@@ -120,3 +128,10 @@ slot[name='icon'] {
 		}
 	}
 }
+
+.close {
+	position: absolute;
+	inset-block-start: 50%;
+	inset-inline-end: 4px;
+	margin-block-start: -12px;
```
rachelbt, line 136:
```suggestion
	transform: translateY(-50%);
```

TaylorJ76:
> didn't went over all the code yet, tested keyboad navigation and screen-reader and its not 100%. When tab is removed (also - there's no way to know it is removable), the focus disappears, shouldn't it go the the latest tab? Also - if all tabs are deleted it creates a strange use case: image maybe we need to add that if only one is left it can't be removed?

As the removal of the tab is up to the consumer, I think handling this use case is also the consumers responsibility. I have update the examples so that when the last tab is removed, the whole tab component is also removed. And I have updated the warning note on the tab page.

rachelbt (commented):
👍 just ine comment about the icon

rachelbt (approved):
looks great

PR #1918, opened 2024-09-24
feat(menu): add position-strategy to allow positioning menu absolute rather then fixed (VIV-2135)

rachelbt:
Not entirely sure about the attribute name 🤔


libs/components/src/lib/menu/README.md
```diff
@@ -175,16 +203,18 @@ While any DOM content is permissible as a child of the menu, only `vwc-menu-item
...
+		<vwc-button
+			slot="anchor"
+			icon="more-vertical-line"
+			aria-label="Open menu"
+			appearance="outlined"
+		></vwc-button>
+		<vwc-menu-item text="Menu item 1"></vwc-menu-item>
+		<vwc-menu-item text="Menu item 2"></vwc-menu-item>
+	</vwc-menu>
+</div>
```
rachelbt, line 217:
```suggestion
<vwc-menu open aria-label="Menu example" placement="bottom-end">
	<vwc-button
		slot="anchor"
		icon="more-vertical-line"
		aria-label="Open menu"
		appearance="outlined"
	></vwc-button>
	<vwc-menu-item text="Menu item 1"></vwc-menu-item>
	<vwc-menu-item text="Menu item 2"></vwc-menu-item>
</vwc-menu>
```

libs/components/src/lib/menu/menu.ts
```diff
@@ -44,6 +44,15 @@ export class Menu extends FastMenu {
 	 */
 	@attr({ mode: 'boolean', attribute: 'auto-dismiss' }) autoDismiss = false;
 
+	/**
+	 * The strategy-absolute attribute.
+	 *
+	 * @public
+	 * HTML Attribute: strategy
+	 */
+	@attr({ mode: 'boolean', attribute: 'strategy-absolute' }) strategyAbsolute =
```
TaylorJ76, line 53: Suggestion:

```
import type { Strategy as Position } from '@floating-ui/dom';

...

@attr position?: Position = 'fixed';
 ```

The reason I'm suggestion `position` is that consumers don't know strategy is because we are using it under the hood in popup.
Position feels a bit more obvious what its function is.

TaylorJ76 (changes requested):

libs/components/src/lib/menu/menu.ts
rachelbt: I agree that strategy is not ideal, the thing is that we also have `placement` - that can be confusing with `position`.

TaylorJ76: How about `placement-strategy`? 
It gives a bit more context to what it does and shows it is related to `placement`.

TaylorJ76 (dismissed):

rachelbt (commented):
@TaylorJ76 , I changed the property to position-strategy, since placement-strategy is something else that we have in popup...

TaylorJ76 (dismissed):

TaylorJ76 (approved):

PR #1927, opened 2024-10-03
feat(searchable-select): component accessibility (VIV-2127)

RichardHelm:


libs/components/src/lib/searchable-select/README.md
```diff
@@ -901,19 +901,17 @@ Use `--searchable-select-height` to set the max-height of the dropdown.
...
-## Use Cases
+If `label` is not set, you must provide an accessible name with the `aria-label` attribute.
 
-### Disabled Options
-
-If an option is disabled, it cannot be selected or unselected.
-
-```html preview 230px
+````html preview 230px
+## Use Cases ### Disabled Options If an option is disabled, it cannot be
```
TaylorJ76, line 907: This seems to have gone a bit wrong.

TaylorJ76 (changes requested):

PR #1934, opened 2024-10-03
chore(button): fix docs (VIV-000)

RichardHelm:


libs/components/src/lib/button/USE-CASES.md
```diff
@@ -1,6 +1,6 @@
 ## Toggle
 
-Buttons may be toggled. The `active` prop represents the change in the state along with the icon and label.
+Buttons may be toggled. Change icon and label to indicate the state of the button.
```
RichardHelm, line 3: Active was not intended for this

TaylorJ76 (approved):

PR #1944, opened 2024-10-10
chore(docs): refactor icons-gallery into component (VIV-000)

RichardHelm:
Fixes some bugs with the old gallery where icons would be displayed multiple times


rachelbt (approved):
LGTM :)

PR #1947, opened 2024-10-15
fix(text-field): reflects inputmode onto input element (VIV-000)

TaylorJ76:


libs/components/src/lib/text-field/text-field.spec.ts
```diff
@@ -194,6 +194,14 @@ describe('vwc-text-field', () => {
 		});
 	});
 
+	describe('inputmode', function () {
+		it('should set inputmode on the input element', async function () {
+			element.inputMode = 'tel';
+			await elementUpdated(element);
+			expect(getInput()?.hasAttribute('inputmode')).toEqual('tel');
```
RichardHelm, line 201:
```suggestion
			expect(getInput()?.getAttribute('inputmode')).toEqual('tel');
```

PR #1948, opened 2024-10-15
chore(docs): update text-field docs to new format (VIV-2111)

TaylorJ76:


libs/components/src/lib/text-field/VARIATIONS.md
```diff
@@ -0,0 +1,188 @@
...
+```html preview
+<vwc-text-field disabled label="Username"></vwc-text-field>
+```
+
+## Read Only
+
+The `readonly` attribute prevents the user from changing the Text Field input element value.
+
+```html preview
+<vwc-text-field disabled label="Username"></vwc-text-field>
```
rachelbt, line 187:
```suggestion
<vwc-text-field readonly label="Username" value="readonly text"></vwc-text-field>
```

libs/components/src/lib/text-field/VARIATIONS.md
```diff
@@ -0,0 +1,188 @@
...
+
+<style>
+	div {
+		display: flex;
+		gap: 16px;
+	}
+</style>
+```
+
+## Shape
```
rachelbt, line 127: in buttons it is shorter and cleaner:
image
same goes to other properties here (size, appearance...).
I think we need to keep all components with the same text.
I personally find it convinient to see the options and default in bullets and not in a paragraph:
image
and not this:
image

libs/components/src/lib/text-field/VARIATIONS.md
```diff
@@ -0,0 +1,188 @@
...
+
+```html preview
+<vwc-text-field
+	icon="search-line"
+	label="Search"
+	type="search"
+></vwc-text-field>
+```
+
+## Scale
```
rachelbt, line 107: maybe we can add here inside a _note_ the comment that we have about size = scale in form elements.

libs/components/src/lib/text-field/VARIATIONS.md
TaylorJ76: I agree. But I actually don't think we need to outline all the options here as they are self explaining in the coded examples, and we have them in the code ref section.
I have removed the extra text.

libs/components/src/lib/text-field/README.md
```diff
@@ -243,166 +146,80 @@ You can add action items elements using the `leading-action-items` slot.
...
-		<vwc-select aria-label="Options Selector" appearance="ghost">
-			<vwc-option value="1" text="ALL" selected></vwc-option>
-		</vwc-select>
-		<vwc-divider orientation="vertical"></vwc-divider>
-	</div>
-</vwc-text-field>
 ```
 
-### Helper-Text
+### Helper Text
```
rachelbt, line 151:
```suggestion
### Helper Text Slot
```

rachelbt (approved):
Great 👍 
left one minor change.

PR #1956, opened 2024-10-21
feat(checkbox): aria-checked now properly conveys and controls checkbox state (VIV-1598)

YonatanKra:
feat(checkbox): aria-checked now controls indeterminate state (VIV-1598)

- [x] aria-checked reflects current checkbox state
- [x] setting aria-checked is reflected in checkbox state

The relevant aria-checked state is the one on the element with `role="checkbox"` - in our case the base element.
FastFoundation also adds `aria-checked` on the host which is currently redundent as it has no role (we should change the base `div` to `template` when we port away from `FoundationCheckbox`.


libs/components/src/lib/checkbox/README.md
```diff
@@ -54,6 +53,30 @@ Use the `indeterminate` member to indicate that the checkbox's is neither on nor
 </script>
 ```
 
+### AriaChecked
```
rachelbt, line 56:
```suggestion
### Aria Checked
```

rachelbt (approved):

PR #1960, opened 2024-10-22
feat(tabs): support action-items slot (VIV-2031)

RichardHelm:
This PR also fixes a bug with the active indicator when tabs are added, and it's general glitchiness


libs/components/src/lib/tab/tab.scss
```diff
@@ -102,33 +102,6 @@ slot[name='icon'] {
 	}
 }
 
-:host(:not(.vertical)) {
```
RichardHelm, line 105: Remove active-indicator from tab and instead keep showing the one from tabs.
Not sure why it was ever added here

libs/components/src/lib/tabs/tabs.ts
```diff
@@ -166,14 +165,68 @@ export class Tabs extends FoundationTabs {
 
 		this.#tabListWrapper.scrollTo({ top, left, behavior: 'smooth' });
 	}
-	// adapted FAST fix https://github.com/microsoft/fast/pull/6606
```
RichardHelm, line 169: This fix has been adopted in our version of FAST so this is no longer needed

libs/components/src/lib/tab/tab.scss
rachelbt: If I remember correctly it was part of the indicator animation. Maybe the change in Fast made this not needed anymore ⚡

rachelbt (commented):
👍

rachelbt (commented):
the code is all OK, but the ui png are not presenting the slotted button - maybe you need to add "button" to the ui tests import

RichardHelm:
@rachelbt You're right, I've updated it

TaylorJ76 (approved):
Nice!

PR #1965, opened 2024-10-24
chore(docs): select page new format (VIV-2124)

TaylorJ76:


libs/components/src/lib/select/VARIATIONS.md
```diff
@@ -0,0 +1,320 @@
...
+		display: flex;
+		gap: 16px;
+	}
+</style>
+```
+
+## Appearance
+
+The `appearance` attribute controls the style of the select element.<br />
+Use `ghost` in combination with a containing element which provides a border, for example [Action Group](/components/action-group/).
```
rachelbt, line 253: I think a better use case can be slotted select inside text-field

libs/components/src/lib/select/GUIDELINES.md
```diff
@@ -0,0 +1,180 @@
...
+</div>
+
+<style>
+	.conainer {
+		display: block;
+		padding: 8px 0;
+	}
+
+	.controls {
+		width: 5.75rem;
```
rachelbt, line 168: why not px?
In general REM are better - but we use px all over....

libs/components/src/lib/select/VARIATIONS.md
```diff
@@ -0,0 +1,320 @@
...
+>
+	<vwc-option value="1" text="+1" icon="flag-united-states"></vwc-option>
+	<vwc-option value="44" text="+44" icon="flag-united-kingdom"></vwc-option>
+	<vwc-option value="49" text="+49" icon="flag-germany"></vwc-option>
+	<vwc-option value="355" text="+355" icon="flag-albania"></vwc-option>
+</vwc-select>
+
+<style>
+	.country-code {
+		width: 120px;
```
rachelbt, line 171:
```suggestion
		inline-size: 120px;
```

libs/components/src/lib/select/README.md
```diff
@@ -1,296 +1,188 @@
...
-</vwc-select>
-```
+## Fixed Dropdown
 
-### Success text
+The `fixed-dropdown` attribute is useful for cases in which the dropdown is obstructed by other elements.
 
-Add the `success-text` to add some success text below the select.
-If provided, `success-text` will take precedence over errors.
+In the example below, if `fixed-dropdown` was not set on the Select, the dropdown menu would be cut off where the Dialog ends.
```
rachelbt, line 81:
```suggestion
In the example below, if `fixed-dropdown` was not set on the Select, the select dropdown would be cut off where the Dialog ends.
```

libs/components/src/lib/select/README.md
```diff
@@ -1,296 +1,188 @@
...
+## Slots
 
-`open` property from _popup_ propagate through _select_ and sets its open state.
+### Default Slot
 
-- Type: `boolean`
-- Default: `false`
+This is where you place the [Option](/components/option/) components to provide the option list.
 
 ```html preview 230px
```
rachelbt, line 157: the options are being cut, 270 would be better :)

libs/components/src/lib/select/README.md
```diff
@@ -376,55 +248,61 @@ Use `--select-height` to set the max-height of the dropdown.
...
+| Name            | Description                                                        |
+| --------------- | ------------------------------------------------------------------ |
+| **default**     | Default slot meant for Option component to provide the option list |
+| **helper-text** | Add HTML content for helper text                                   |
+| **icon**        | Add an icon to the component.                                      |
+| **meta**        | Show meta information after the selected option label              |
+
+</div>
 
 ## Accessibility
```
rachelbt, line 307: from here and to the end of the file - need to be removed

libs/components/src/lib/select/USE-CASES.md
```diff
@@ -0,0 +1,144 @@
+## Validation
+
+### In a Form
+
+When a Select is placed inside a `form` element and validation logic is set on the Text Field, the Text Field is validated when the `form` is submitted (as per a native `select` element).
```
rachelbt, line 5: everything here is about text-field :) :) :)

libs/components/src/lib/select/USE-CASES.md
```diff
@@ -0,0 +1,144 @@
...
+	}
+	form {
+		max-inline-size: 300px;
+	}
+</style>
+```
+
+### Single Field
+
+The Text Field can be validated like the native `input` element.
```
rachelbt, line 40: same.....

libs/components/src/lib/select/USE-CASES.md
```diff
@@ -0,0 +1,144 @@
...
+	};
+</script>
+```
+
+## International Phone Number
+
+The Select component is used in the example below as the dialing code part of an international phone number input field.
+
+```html preview 240px
+<vwc-action-group slot="body">
```
rachelbt, line 64: maybe its better to add a use case using text-field with slotted select?

libs/components/src/lib/select/ACCESSIBILITY.md
```diff
@@ -0,0 +1,4 @@
+## Implementation
```
rachelbt, line 1: the link to the manual test is missing

libs/components/src/lib/select/README.md
```diff
@@ -376,55 +248,61 @@ Use `--select-height` to set the max-height of the dropdown.
...
 <div class="table-wrapper">
 
-| Name              | Type       | Default     | Description                                                      |
-| ----------------- | ---------- | ----------- | ---------------------------------------------------------------- |
-| `options`         | `Option[]` | `[]`        | A read-only list of options.                                     |
-| `selectedOptions` | `Option[]` | `[]`        | A read-only collection of the selected options.                  |
-| `selectedIndex`   | `number`   | `undefined` | The index of the selected option or -1 if no option is selected. |
+| Name               | Type                            | Description                                                                                                                 |
+| ------------------ | ------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| **appearance**     | `fieldset` (default), `ghost`   | Sets the elements's appearance                                                                                              |
```
rachelbt, line 267:
```suggestion
| **appearance**     | `fieldset` (default), `ghost`   | Sets the element's appearance                                                                                              |
```

libs/components/src/lib/select/GUIDELINES.md
```diff
@@ -0,0 +1,180 @@
...
+		placeholder="Select an option"
+		class="marketing"
+	>
+		<vwc-option value="friend" text="A friend"></vwc-option>
+		<vwc-option value="net" text="Internet search"></vwc-option>
+		<vwc-option value="online-ad" text="Online advert"></vwc-option>
+		<vwc-option value="radio-ad" text="Radio advert"></vwc-option>
+		<vwc-option value="other" text="Other"></vwc-option>
+	</vwc-select>
+	<vwc-toggletip placement="bottom">
```
rachelbt, line 76: I'm not so sure this is a good example, we kind of making this hack that we do not suggest devs to do - and showing them how to do it. I know its under don't use" but still, something feels a bit off there

rachelbt (commented):
👍 
left a small comment about guidelines

libs/components/src/lib/select/GUIDELINES.md
TaylorJ76: I don't think it's a problem. We are not showing the controls to view the code and I think it is preferable to using the image.
The trouble is designers are making using this pattern already so I hope this will help to discourage them.

PR #1974, opened 2024-10-30
feat(file-picker): add single file mode (VIV-1988)

YonatanKra:


libs/components/src/lib/file-picker/README.md
```diff
@@ -114,6 +114,19 @@ Use the `size` attribute to set the file picker's to one of the predefined block
 </vwc-file-picker>
 ```
 
+### Single File
```
TaylorJ76, line 117: Can we move this example up the page a bit to be above the Max File Size example?

TaylorJ76 (changes requested):
None of these examples work unless they are wrapped in a form (like in the Form use case). Can we wrap all then examples in a form so that users can see how it functions?

Also, can we change the Max Files example to be more than 1, and put a note to say to use the `single-file` attribute if you only want to upload one file?

libs/components/src/lib/file-picker/README.md
YonatanKra: I moved it just below `label`.
Regarding the form change - can you please elaborate what doesn't work outside a form?
We never put form elements inside a form in the examples, because we'd like people to be able to just copy and paste the code (K.I.S.).

TaylorJ76: Sorry, my bad. It does work. It's just that the file list gets cut off in the examples.
Can we give the examples more height so that at least one select file will appear inside the viewing area?

YonatanKra: Done

TaylorJ76 (approved):

PR #1978, opened 2024-11-04
chore(docs): banner page in new format (VIV-2084) 

rachelbt:


libs/components/src/lib/banner/README.md
```diff
@@ -231,7 +209,7 @@ If set, the `icon` attribute is ignored.
 
 </div>
```
TaylorJ76, line 211: The slot table is missing. I know they are listed above, but I think we should repeat it in tablular form here.

libs/components/src/lib/banner/GUIDELINES.md
```diff
@@ -0,0 +1,77 @@
+## Text
+
+<docs-do-dont no-gutters headline="avoid long sentences in the Banner">
```
TaylorJ76, line 3:
```suggestion
<docs-do-dont no-gutters headline="Avoid long sentences in the Banner">
```

TaylorJ76 (changes requested):
Just a couple of small changes

TaylorJ76 (approved):

PR #1980, opened 2024-11-05
feat(file-picker): adjust styles for better user interaction (VIV-1997)

TaylorJ76:
Adds border around selected files for more definition
Adds hover styles over clickable / drag and drop element for better interaction


libs/components/src/lib/file-picker/file-picker.scss
```diff
@@ -47,6 +47,10 @@ $file-picker-list-item-background-color: --file-picker-list-item-background-colo
 
 		@include focus.focus;
 	}
+	&:hover {
```
rachelbt, line 50: you can add to .control  `transition: all 0.3s ease;` it will make the change nicer

rachelbt (approved):
I think this component is a but strange with its color. 
But we don't have any design feed back or needs so we keep it as it is :)

PR #1983, opened 2024-11-06
feat(file-picker): support clearing files (VIV-1995)

RichardHelm:


libs/components/src/lib/file-picker/file-picker.ts
```diff
@@ -50,22 +50,25 @@ const generateFilePreviewTemplate = (
 @errorText
 @formElements
 export class FilePicker extends FormAssociatedFilePicker {
-	#dropzone?: Dropzone;
+	/**
+	 * @internal
+	 */
+	_dropzone?: Dropzone;
```
YonatanKra, line 56: Why do we need to expose dropzone?

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -73,6 +73,7 @@ describe('vwc-file-picker', () => {
 			unmountedElement.maxFileSize = 256;
 			unmountedElement.maxFiles = 1;
 			unmountedElement.accept = '.jpg';
+			unmountedElement.removeAllFiles();
```
YonatanKra, line 76: Isn't `removeAllFiles` tested in its own `it`?

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -111,6 +112,17 @@ describe('vwc-file-picker', () => {
 			addFiles([await generateFile('london.png', 1)]);
 			expect(element.value).toBe('C:\\fakepath\\london.png');
 		});
+
+		it('should remove all files when setting value to an empty string', async function () {
```
YonatanKra, line 116: Nice addition :)

YonatanKra (changes requested):
Nice and very well needed feature. Thanks :)

libs/components/src/lib/file-picker/file-picker.spec.ts
RichardHelm: Yes, this is just to cover the optional chain.
To ensure it doesn't throw when component is unmounted

libs/components/src/lib/file-picker/file-picker.ts
RichardHelm: Otherwise there is an error when accessing `#dropzone` from `valueChanged`
I believe this is because FAST will call `valueChanged` before the constructor is completed and the private member is not ready

YonatanKra: I pushed a change that reverts it. Seems to work.
The problem is that `valuechanged` is sent to an event listener, which fires outside the context of the element.
By changing the `valueChanged` to be an arrow function, it retains the context wherever it is called.

libs/components/src/lib/file-picker/file-picker.spec.ts
YonatanKra: Looking at the full test (for some reason it was hidden in the diff), it's a very general test and an unclear spec.
1. Would `removeAllFiles` cause an error in unmounted state at any point?
2. Did removing and appending the component to the DOM (connecting/disconnecting in WC terms?) cause an error before?
In other words, `what piece of code do I delete for the expectation to fail?`

libs/components/src/lib/file-picker/file-picker.ts
RichardHelm: That's not what's happening though. With your fix `valueChange` will be initialised at a later point and miss the initial call that causes the error. But I'm happy to accept it like this

YonatanKra (approved):

PR #1991, opened 2024-11-07
chore(workspace): update playwright to 1.48.0 (VIV-2181)

rachelbt:


apps/docs/playwright.config.ts
```diff
@@ -1,4 +1,9 @@
 import type { PlaywrightTestConfig } from '@playwright/test';
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+	snapshotPathTemplate: '{testDir}/__screenshots__/{testFilePath}/{arg}{ext}',
+});
```
RichardHelm, line 6: I don't think this actually does anything? After all we don't have `__screenshots__` in our paths so I think it can be removed
```suggestion
```

RichardHelm (changes requested):
Looks good!

RichardHelm (approved):

PR #1994, opened 2024-11-11
chore(docs): adds label example to button (VIV-000)

TaylorJ76:


rachelbt (commented):
can you please add it also to the properties in code?

TaylorJ76:
> can you please add it also to the properties in code?

It is already there. Above href

rachelbt (approved):

PR #2000, opened 2024-11-13
chore(switch): remove foundation dependency (VIV-2026)

TaylorJ76:


libs/components/src/lib/switch/switch.ts
```diff
@@ -31,4 +32,78 @@ export class Switch extends FoundationSwitch {
...
+	/**
+	 * When true, the control will be immutable by user interaction. See {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly | readonly HTML attribute} for more information.
+	 * @public
+	 * @remarks
+	 * HTML Attribute: readonly
+	 */
+	// @ts-expect-error Type is incorrectly non-optional
+	@attr({ attribute: 'readonly', mode: 'boolean' }) readOnly: boolean; // Map to proxy element
+	// @ts-expect-error Function is declared but never read
+	private readOnlyChanged(): void {
```
RichardHelm, line 45: For consistency with our codebase where we do:
```suggestion
	/**
	 * @internal
	 */
	readOnlyChanged(): void {
```

libs/components/src/lib/switch/switch.ts
```diff
@@ -31,4 +32,78 @@ export class Switch extends FoundationSwitch {
...
+	 *
+	 * @internal
+	 */
+	override initialValue = 'on';
+
+	/**
+	 * @internal
+	 */
+	// @ts-expect-error Type is incorrectly non-optional
+	@observable defaultSlottedNodes: Node[];
```
RichardHelm, line 67: I think for internals we should just do:
```suggestion
	@observable defaultSlottedNodes!: Node[];
```

RichardHelm (dismissed):
Looks good, I have some minor suggestions

RichardHelm (approved):

PR #2006, opened 2024-11-14
feat(elevation): New shadows colors poc (VIV-000)

rachelbt:


rachelbt:
was done properly here: https://github.com/Vonage/vivid-3/pull/2062

PR #2017, opened 2024-11-19
feat(vue-wrappers): type event targets (VIV-000)

RichardHelm:
Until https://jira.vonage.com/browse/VIV-1956 is resolved, these types do not actually work and `target` will be `any` instead


rachelbt (commented):
looks OK.
Only the code coverage issue

rachelbt (approved):
👍

PR #2018, opened 2024-11-19
fix(popup): add min-width to the popup-wrapper (VIV-000)

rachelbt:
To fix issues with max-width calculation when the popup is off-screen and needs to be flipped the other way


rachelbt:
didn't solve the issue

PR #2020, opened 2024-11-20
Create fr-FR

alexistb2904:
Add the french language


TaylorJ76:
@alexistb2904 Thanks for your contribution, but we have no immediate business requirements to add french to our support localized language set.

PR #2027, opened 2024-11-22
chore(docs): alert page restructure (VIV-2080)

RichardHelm:


libs/components/src/lib/alert/VARIATIONS.md
```diff
@@ -0,0 +1,125 @@
...
+	text="Some important information for you"
+	open
+></vwc-alert>
+```
+
+### Connotation
+
+Use the `connotation` attribute to change the Alert's icon and icon color.
+
+Note that each connotation comes with a default icon (that you can override with the `icon` attribute).
```
rachelbt, line 25: this can actually be inside a note :)

libs/components/src/lib/alert/VARIATIONS.md
```diff
@@ -0,0 +1,125 @@
...
+	document.querySelector('vwc-select').addEventListener('change', (e) => {
+		document.querySelector('vwc-alert').connotation = e.target.value;
+	});
+</script>
+```
+
+### Icon
+
+Use the `icon` attribute to add an icon to the Alert. Note that setting this attribute takes precedence
+over the connotation's icon, if any.
```
rachelbt, line 53: we usually also reffer to the slot icon to add icon that is not in the gallery

libs/components/src/lib/alert/README.md
```diff
@@ -1,42 +1,54 @@
...
 
-### Text
-
-Use the `text` attribute to set the alert's main text.
+```js
+import { registerAlert } from '@vonage/vivid';
 
-- Type: `string`
-- Default: `undefined`
+registerAlert('your-prefix');
```
rachelbt, line 16: this is the only component that we have then in 2 rows:
image
I like it better - but it doesn't look like the rest of the components

rachelbt (commented):
started reviewing before I noticed that its not marked as ready for review.
Left some comments and stoped.

libs/components/src/lib/alert/README.md
RichardHelm: I do not understand, this is the same as Button etc. right?

rachelbt: yes you are right, I don't know why I remembered it was placed on the same line. Sorry.

libs/components/src/lib/alert/README.md
```diff
@@ -1,42 +1,54 @@
...
-```
+## Opening and Closing
 
 ### Open
 
-Use the `open` attribute to toggle the alert open state.
+Use the `open` attribute to toggle the Alert open state.
 
 - Type: `boolean`
 - Default: `false`
```
rachelbt, line 54: see comment about position

libs/components/src/lib/alert/ACCESSIBILITY.md
```diff
@@ -0,0 +1,7 @@
+## Implementation
+
+- Use the `timeoutms` attribute with care. Users with disabilities may need more time to read the content. If the Alert contains interactive elements, such as links or buttons, you should avoid using `timeoutms`. See also [WCAG SC 2.2.3 No Timing](https://www.w3.org/WAI/WCAG22/Understanding/no-timing).
```
rachelbt, line 3:
```suggestion
Use the `timeoutms` attribute with care. Users with disabilities may need more time to read the content. If the Alert contains interactive elements, such as links or buttons, you should avoid using `timeoutms`. See also [WCAG SC 2.2.3 No Timing](https://www.w3.org/WAI/WCAG22/Understanding/no-timing).
```

libs/components/src/lib/alert/ACCESSIBILITY.md
```diff
@@ -0,0 +1,7 @@
...
+
+## Resources
+
+- [ARIA APG Alert Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alert/)
```
rachelbt, line 7: you are missing the accessibiliy test:
https://docs.google.com/spreadsheets/d/1Jo0Vn2V-A2K_NKSBwE78-Pn-P_l7VDzQwZ0rD1CNo-0/edit?gid=1175911860#gid=1175911860

libs/components/src/lib/alert/ui.test.ts
```diff
@@ -1,31 +1,28 @@
...
-			path.join(new URL('.', import.meta.url).pathname, 'README.md')
-		).reduce(
-			(htmlString: string, block: string) =>
-				`${htmlString} <div>${block}</div>`,
-			''
-		);
-
-	page.setViewportSize({ width: 1300, height: 2750 });
+	const template = `
+		<div style="padding: 5px; display: flex; flex-direction: column; gap: 8px;">
```
rachelbt, line 12: we are missing the placement test

rachelbt (commented):
dont we have a use case for alert?
Maybe we can add use case for using the strategy?

libs/components/src/lib/alert/README.md
```diff
@@ -200,48 +93,38 @@ Use the `removable` attribute to add a close button to the alert.
...
+## Positioning
 
-Use the `timeoutms` attribute to set the time after which the alert will automatically close.
+### Strategy
+
+Controls the `position` of the Alert. The default is `fixed`, which will position the Alert relative to the viewport.
 
-- Type: `number` (in milliseconds)
-- Default: `0` (stays open indefinitely)
+When set to `static`, placement will have no effect, and the Alert will behave as an element in page flow.
```
rachelbt, line 102: do you think this will be better inside a note?

rachelbt (commented):
all is OK.
Left one more comment.
I think that we should add a use case with the Alert.

apps/docs/content/_data/components.json
```diff
@@ -26,8 +26,12 @@
 	},
 	{
 		"title": "Alert",
-		"page": "legacy",
-		"markdown": "./libs/components/src/lib/alert/README.md"
+		"description": "Alerts display short-lived important information to the user, usually at the top or bottom of the screen.\n\n",
```
rachelbt, line 29:
```suggestion
		"description": "Alerts display short-lived important information to the user, usually at the top or bottom of the screen.",
```

libs/components/src/lib/alert/README.md
RichardHelm: I don't think so, it just makes it harder to read

rachelbt:
> @rachelbt Sure, I've added a use case

nice one!

rachelbt (approved):

PR #2028, opened 2024-11-24
feat(components): add styled scrollbar (VIV-2214)

rachelbt:
text-area width changed in chrome - due to scrollbar design - this is why the ui test has updated - the text-area is smaller in 2px


TaylorJ76 (approved):
Why do we have 1 updated snapshot? I can't see what has changed.
Otherwise all good.

TaylorJ76 (changes requested):
How will the consumers add the scrollbar styles to their applications?
Will this be in a separate PR?

TaylorJ76 (approved):

PR #2029, opened 2024-11-24
feat(file-picker): add styled scrollbar & fix files-list height (VIV-2236)

rachelbt:


libs/components/src/lib/file-picker/README.md
```diff
@@ -170,6 +170,10 @@ The `helper-text` slot allows you to use rich content as the file picker's helpe
 Each added file in the preview list has a default background-color of `vvd-color-canvas` (`#fff` in light theme, `#000` in dark).  
 If needed, the background of the item can be changed using the `--file-picker-list-item-background-color` CSS variable.
 
+### file-picker-list-block-size
+
+Use `--file-picker-list-block-size` the set the max-block-size of list of the added files, other than `408px`.
```
RichardHelm, line 175: Why do we need a default value?

And wouldn't it make more sense to build the layout in a way where the list of files becomes scrollable if there isn't enough space for the component.
So when setting `max-height: 600px` on the host and the list of files becomes too big it will automatically scroll without any CSS variables.

RichardHelm (changes requested):

libs/components/src/lib/file-picker/file-picker.scss
```diff
@@ -48,8 +51,9 @@ $file-picker-list-item-background-color: --file-picker-list-item-background-colo
 
 		@include focus.focus;
 	}
-	&:hover {
-		border-style: solid;
```
rachelbt, line 52: was removed as requested here: https://jira.vonage.com/browse/VIV-2227 (in the comments)

RichardHelm (approved):

PR #2030, opened 2024-11-25
feat(file-picker): custom internal error messages (VIV-1998)

TaylorJ76:


libs/components/src/lib/file-picker/README.md
```diff
@@ -76,6 +62,12 @@ If the user tries to upload a file that exceeds the maximum file size, the file
 </vwc-file-picker>
 ```
 
+#### File Too Big Error
+
+Use the `file-too-big-error` attribute to override the default localized message.
```
RichardHelm, line 67: This needs some more information / example on how to use with placeholders

RichardHelm (changes requested):

RichardHelm (approved):

PR #2032, opened 2024-11-26
chore(option): decoupled fast foundation (VIV-2024)

TaylorJ76:


libs/components/src/lib/option/option.ts
```diff
@@ -1,16 +1,177 @@
...
 /**
+ * An Option Custom HTML Element.
+ * Implements {@link https://www.w3.org/TR/wai-aria-1.1/#option | ARIA option }.
+ *
+ * @slot start - Content which can be provided before the listbox option content
+ * @slot end - Content which can be provided after the listbox option content
+ * @slot - The default slot for listbox option content
+ * @csspart content - Wraps the listbox option content
+ *
  * @public
```
RichardHelm, line 15:
```suggestion
 * An Option Custom HTML Element.
 * Implements {@link https://www.w3.org/TR/wai-aria-1.1/#option | ARIA option }.
 *
 * @public
 * @component option
 * @slot icon - Slot to add an icon to the option.
```

libs/components/src/lib/option/option.spec.ts
```diff
@@ -114,6 +108,43 @@ describe('vwc-option', () => {
...
+		});
+	});
+
+	describe('form', () => {
+		it('should return null if not in a form', async () => {
+			expect(element.form).toBe(null);
+		});
+
+		// Does not work:
+		xit('should return the parent form', async () => {
```
RichardHelm, line 139: So this never worked in the original code either. I've created a ticket for it: https://jira.vonage.com/browse/VIV-2243

PR #2033, opened 2024-11-27
feat(menu): move focus into menu on open (VIV-2196)

RichardHelm:


libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -730,19 +761,6 @@ describe('vwc-menu', () => {
 
 			expect(item1.expanded).toBe(false);
 		});
-
```
rachelbt, line 733: why is this test not needed any more?

rachelbt (approved):

libs/components/src/lib/menu/menu.spec.ts
RichardHelm: Submenus now close because focus moves out of them. So the code in the component to close other submenus isn't needed anymore and this corresponding test doesn't make sense anymore

PR #2038, opened 2024-11-29
chore: tabs docs page

TaylorJ76:


libs/components/src/lib/tabs/VARIATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+</vwc-tabs>
+```
+
+## Shape
+
+The `shape` attribute on the **Tab** component controls the style of the background in hover state. It can be `rounded` or `shape`.
+
+```html preview full
+<vwc-tabs>
+	<vwc-tab shape="rounded" label="Rounded"></vwc-tab>
```
rachelbt, line 114: I'm not so sure about having them together in the same tabs wrapper. Its a bad practice that we do not want them to use. Having said that- it is useless to have 2 sets of tabs.
Maybe we can use a drop down that changes the shape?
Or a code sample with no preview and an image to illustrate the difference.

libs/components/src/lib/tabs/VARIATIONS.md
```diff
@@ -0,0 +1,262 @@
...
+	<vwc-tab label="Tab three" id="tab-3"></vwc-tab>
+	<vwc-tab-panel>Tab Panel one</vwc-tab-panel>
+	<vwc-tab-panel>Tab Panel two</vwc-tab-panel>
+	<vwc-tab-panel>Tab Panel three</vwc-tab-panel>
+</vwc-tabs>
+```
+
+## Gutters
+
+Use the `gutters` attribute on the **Tabs** component to control the spacing inside the Tab Panels. Below it is set to `none`.
```
rachelbt, line 168: gutters can have either `none` or `small` (that are the default). I think that this is not clear enough here

libs/components/src/lib/tabs/README.md
```diff
@@ -1,194 +1,105 @@
...
+## Usage
 
-Represents a tabs custom element.
-The vwc-tabs accepts [vwc-tab](/components/tab/) and `vwc-tab-panel` elements as children. Read more about `tabs` [here](https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/).
+<vwc-tabs>
+<vwc-tab label="Web component"></vwc-tab>
+<vwc-tab-panel>
 
 ```js
-<script type="module">import '@vonage/vivid/tabs';</script>
```
rachelbt, line 7: theres an unneeded space here:
image

libs/components/src/lib/tabs/README.md
```diff
@@ -212,87 +123,110 @@ You can use the `action-items` slot to add action items at the end of the tabs b
...
-| -------- | -------------------------- | ------- | -------- | --------------------------------------------------------------------------------- |
-| `change` | `CustomEvent<HTMLElement>` | Yes     | Yes      | Fires a custom 'change' event when a tab is clicked or during keyboard navigation |
+| Name                 | Type                               | Description                                                                            |
+| -------------------- | ---------------------------------- | -------------------------------------------------------------------------------------- |
+| **activeid**         | `string`                           | Match with an `id` set on a Tab to mark it as active on initial load                   |
+| **connotation**      | `accent` (default), `cta`          | Sets the connotation color of the active tab                                           |
+| **gutters**          | `none` (default), `small`          | Sets the spacing inside the Tab Panels                                                 |
```
rachelbt, line 159:
```suggestion
| **gutters**          | `none`, `small` (default)        | Sets the spacing inside the Tab Panels                                                 |
```

libs/components/src/lib/tabs/README.md
```diff
@@ -212,87 +123,110 @@ You can use the `action-items` slot to add action items at the end of the tabs b
...
+### Tab
+
+#### Properties
+
+<div class="table-wrapper">
+
+| Name              | Type                         | Description                          |
+| ----------------- | ---------------------------- | ------------------------------------ |
+| **disabled**      | `boolean`                    | Sets the disabled state              |
+| **icon-trailing** | `boolean`                    | Places the icon after the label text |
```
rachelbt, line 195: `icon` is missing

rachelbt (changes requested):
wasn't sure that it will be OK to unify tabs and tab together but it looks great!
Do you think it would be nice to add a use case where there are a lot of tabs and there is a scroll shadow?

libs/components/src/lib/tabs/README.md
TaylorJ76: Doesn't seem to be much we can do about this. Happens in other places too.

libs/components/src/lib/tabs/VARIATIONS.md
TaylorJ76: I've added a note after the example to say how they shouldn't be used together.

PR #2058, opened 2024-12-12
chore(components): decoupling loose ends (VIV-000)

TaylorJ76:
Adds missing types and a helper functions.
Changes many components to use the new local classes / functions.


libs/components/src/lib/menu-item/menu-item.ts
```diff
@@ -1,10 +1,5 @@
 import { attr, DOM, observable } from '@microsoft/fast-element';
-import {
-	AnchoredRegion,
-	applyMixins,
-	MenuItemRole as FastMenuItemRole,
-	getDirection,
-} from '@microsoft/fast-foundation';
+import { AnchoredRegion } from '@microsoft/fast-foundation';
```
YonatanKra, line 2: That's still reliant on `@microsoft/fast-foundation`. Shouldn't this be removed as well?

libs/components/src/lib/number-field/number-field.ts
```diff
@@ -5,8 +5,8 @@ import {
 	observable,
 } from '@microsoft/fast-element';
 import { keyArrowDown, keyArrowUp } from '@microsoft/fast-web-utilities';
-import { DelegatesARIATextbox } from '@microsoft/fast-foundation';
 import type { Appearance, Shape, Size } from '../enums';
+import { DelegatesARIATextbox } from '../text-field/text-field';
```
YonatanKra, line 9: Maybe this should be part of our shared foundation for form elements instead of in the `text-field`?

YonatanKra (changes requested):
Two comments. Not bad for such a big PR :)

libs/components/src/lib/menu-item/menu-item.ts
TaylorJ76: I have created a new task to deal with this.

libs/components/src/lib/number-field/number-field.ts
YonatanKra: I've added this to our dev tasks list.

YonatanKra (approved):

PR #2065, opened 2024-12-19
fix(layout): column basis medium and large in mobile (VIV-2049)

rachelbt:
In both switch and action-group there's width of 380px and 250px - which makes the page behave as mobile - this is why the images updated


YonatanKra (approved):
LGTM
Thanks :)

PR #2069, opened 2024-12-25
fix(dialog): prevent scroll to top when opened as Modal (VIV-2248)

rachelbt:
Also we had a but that the dialog that is not modal was positioned relative and not absolute.
This is also fixed now.

### DIFF
--- a/libs/components/src/lib/dialog/dialog.scss
+++ b/libs/components/src/lib/dialog/dialog.scss
@@ -40,8 +40,6 @@ $dialog-space-size: 24px;
 	}
 
 	&.modal {
-		position: fixed;
-
 		// reason for box-shadow & hardcoded color - PR-1098
 		box-shadow: 0 4px 20px rgb(0 0 0 / 35%);
 		&::backdrop {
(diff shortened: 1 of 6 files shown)


YonatanKra (commented):
Looks good.
2 questions:
1. How does "no shadow" help with the scroll?
2. The name `no-position` is a not as descriptive as I'd expect. Can we make this more descriptive? I guess `not-relative` would be more descriptive but meaningless for most users. Can we set the name to something that's related to consumer's experience?

rachelbt:
> Looks good. 2 questions:
> 
> 1. How does "no shadow" help with the scroll?
> 2. The name `no-position` is a not as descriptive as I'd expect. Can we make this more descriptive? I guess `not-relative` would be more descriptive but meaningless for most users. Can we set the name to something that's related to consumer's experience?

no-shadow had no effect but was missing a test so I added one 🦾 
I agree about the name - I also thought so - and I mentioned it in the comment of this PR.
I will change the name!

YonatanKra (approved):

PR #2070, opened 2025-01-01
fix(radio): radio required validation sync with spec (VIV-2265)

YonatanKra:
The spec says the following:
- Radio with a `name` property should be validated for required
- Radio with a `name` in a group should be validated for required with all siblings of the same `name` such that if one of them is checked, required is satisfied and then `validity.valueMissing` should be false.


rachelbt (commented):
👍 
Mind reverting the `package-lock.json`?

YonatanKra:
> 👍 Mind reverting the `package-lock.json`?

No. I installed a new package. It had to change.

rachelbt (approved):

PR #2074, opened 2025-01-06
fix(video-player): control panel design fixes (VIV-2246, VIV-2275)

rachelbt:


libs/components/src/lib/video-player/ui.test.ts
```diff
@@ -11,7 +11,7 @@ test('should show the component', async ({ page }: { page: Page }) => {
 	const template = `
 		<style>
 			.layout {
-				max-width: 800px;
+				max-width: 760px;
```
rachelbt, line 14: I changed the width so the new pngs will be generated

TaylorJ76 (approved):

PR #2078, opened 2025-01-09
fix(menu): add default value to inline-size in small screens (VIV-2006)

rachelbt:


libs/components/src/lib/menu/README.md
```diff
@@ -269,6 +269,11 @@ Use the `--menu-max-inline-size` variable to set the menu's inline size.
 
 - Default: `max-content`
 
+<vwc-note connotation="information" icon="info-solid">
+<p>When setting a value to the <code>max-inline-size</code> - make sure the menu is OK in small resolutions as well. </p>
+<p>In mobile, the <code>max-inline-size</code> is <code>280px</code> by default, but can be changed with the css-variable. </p>
```
TaylorJ76, line 274: Isn't the value `300px` in mobile?

TaylorJ76 (changes requested):

TaylorJ76 (approved):

PR #2080, opened 2025-01-13
chore(docs): Badge docs update (VIV-2083)

rachelbt:


libs/components/src/lib/badge/VARIATIONS.md
```diff
@@ -0,0 +1,162 @@
...
+<vwc-badge
+	icon-trailing
+	icon="message-sent-line"
+	appearance="filled"
+	text="icon trailing"
+	shape="pill"
+></vwc-badge>
+```
+
+### Icon Only
```
TaylorJ76, line 146: I think we should remove this example. I don't think it's something we want to encourage.

TaylorJ76 (changes requested):

libs/components/src/lib/badge/VARIATIONS.md
rachelbt: we have a use case of icon-only badge here:
https://vivid.deno.dev/components/menu-item/#meta 
image

TaylorJ76: Ah ok. Fair enough.

TaylorJ76 (approved):

PR #2081, opened 2025-01-14
chore(docs): Avatar docs update (VIV-2082)

rachelbt:


libs/components/src/lib/avatar/README.md
```diff
@@ -253,9 +199,29 @@ Assign nodes to the `graphic` slot to set a graphic media of any kind (e.g. imag
...
+<div class="table-wrapper">
+
+| Name          | Type                                                      | Description                      |
+| ------------- | --------------------------------------------------------- | -------------------------------- |
+| `appearance`  | Enum: `filled` (default), `duotone`, `outlined`, `subtle` | Sets the element's appearance    |
+| `connotation` | Enum: `accent` (default), `cta`                           | Sets the element's connotation   |
+| `icon`        | `string`                                                  | Sets the element's icon          |
+| `initials`    | `string`                                                  | Sets the initials on the Avatar  |
+| `shape`       | Enum: `rounded` (default), `pill`                         | Sets the element's border-radius |
+| `size`        | Enum: `condensed`, `normal` (default), `expanded`         | Sets the element's text          |
```
TaylorJ76, line 215:
```suggestion
| `size`        | Enum: `condensed`, `normal` (default), `expanded`         | Sets the element's size         |
```

TaylorJ76 (changes requested):
I think it would be good to show a Use Case of an Avatar wrapped in a button with a menu attached.

rachelbt:
> I think it would be good to show a Use Case of an Avatar wrapped in a button with a menu attached.

I added 2 usecases :)

TaylorJ76 (approved):

PR #2082, opened 2025-01-16
chore(docs): Breadcrumb docs update (VIV-2085)

rachelbt:


rachelbt:
> Missing the API Reference section on the Code page.

I added it it the breadcrumb-item and not breadcrumb....

PR #2084, opened 2025-01-16
chore(docs): audio player docs update (VIV-2081) 

TaylorJ76:


libs/components/src/lib/audio-player/VARIATIONS.md
```diff
@@ -0,0 +1,73 @@
...
+
+<vwc-audio-player
+	skip-by="30"
+	src="https://freetestdata.com/wp-content/uploads/2021/09/Free_Test_Data_2MB_MP3.mp3"
+></vwc-audio-player>
+```
+
+## Playback Rates
+
+Playback rates can be modified by passing a comma separated string of numbers to the playback-rates attribute. The playback rates option can be removed by passing an empty string or not including the attribute at all.
```
rachelbt, line 43: I wonder if this should go into a "note":
The playback rates option can be removed by passing an empty string or not including the attribute at all.

rachelbt (commented):
left one small comment

PR #2087, opened 2025-01-20
chore(listbox): remove unused internal listbox component (VIV-2247)

RichardHelm:


YonatanKra (approved):
Looks legit, thanks.
Just check the single comment I added plz.

PR #2088, opened 2025-01-20
fix(vue-wrappers): fix type definitions (VIV-2274)

RichardHelm:


YonatanKra (changes requested):
1. In the issue there's a mention of the VTabPanel type which looks awkward. When I trested this branch locally it didn't fix it.
2. The title in button gets this remark:
```
    /**
     * Determines if the element should receive document focus on page load.
     */
    title: {
        type: PropType<string>;
        default: undefined;
    };
 ```
3. This one can be done in a different issue as a tech debt: I suggest adding some kind of test to make sure we don't get such a production mishap again.
Two suggestions:
- Make elaborate tests making sure everything's passing.
- Compare the JSON to the one before and alert if changed and require if manually approved.

YonatanKra (approved):

PR #2089, opened 2025-01-21
chore(video-player): use dark-light function (VIV-000)

rachelbt:


libs/components/src/lib/video-player/partials/variables.scss
```diff
@@ -1,8 +1,17 @@
 // setting variables with hard-coded color, since design required both dark & light be the same
```
RichardHelm, line 1:
```suggestion
```

rachelbt: this is not needed to all - but is needed to some of the variable - maybe I should change the text position

libs/components/src/lib/video-player/partials/variables.scss
```diff
@@ -1,8 +1,17 @@
...
-$video-player-canvas-text: #fff;
+$video-player-neutral-800: light-dark(
+	var(--vvd-color-neutral-800),
+	var(--vvd-color-neutral-100)
+);
+$video-player-canvas-text: light-dark(
+	var(--vvd-color-canvas),
+	var(--vvd-color-canvas-text)
+);
 $video-player-canvas: #0d0d0d;
```
rachelbt, line 15:
```suggestion
// setting variables with hard-coded color, since design required both dark & light be the same, and #0d0d0d is not set in both themes.
$video-player-canvas: #0d0d0d;
```

rachelbt:
Closing this PR because unfortunately its not effecting correctly when the class `vvd-theme-alternate` is added :(

PR #2094, opened 2025-01-23
fix(data-grid): fix error on createElement (VIV-2289)

RichardHelm:


YonatanKra (changes requested):
Thanks for solving this one.
What use case does this solve?
Can you please add a test that fails without this fix?

RichardHelm:
@YonatanKra 

It solves `document.createElement('vwc-data-grid')` not working.

This isn't tested for any of our components, so it doesn't make sense to add a test for data-grid only. It would need to be added to all components.

YonatanKra:
> @YonatanKra
> 
> It solves `document.createElement('vwc-data-grid')` not working.
> 
> This isn't tested for any of our components, so it doesn't make sense to add a test for data-grid only. It would need to be added to all components.

I does make sense to test something that broke in production to make sure it doesn't break again.
Tests are there to document the code as well as keep us from regressions.
A test for this will answer both criteria.

RichardHelm:
@YonatanKra

Sure, but it doesn't belong into this PR. As I said, it would logically need to be tested for all components. We should consider adding some common tests that run for all components, so that we can easily do so.

RichardHelm:
@YonatanKra Sure, I've added a test and created a ticket to add it to all components here: https://jira.vonage.com/browse/VIV-2293

YonatanKra (approved):
LGTM
Thanks!

PR #2097, opened 2025-01-24
chore(docs): divider docs update (VIV-2094)

TaylorJ76:


rachelbt (approved):
👍

rachelbt (changes requested):
there's something off in the ui-test - was off before as well - there's no divider here:
image

maybe card / dialog needs to be added to the ui-test :)

PR #2098, opened 2025-01-24
fix(divider): orientation attribute issue (VIV-2290)

TaylorJ76:


libs/components/src/lib/divider/divider.spec.ts
```diff
@@ -28,6 +28,26 @@ describe('vwc-divider', () => {
 
 			expect(base?.classList.contains(`${orientation}`)).toBeTruthy();
 		});
+
+		it('should set the aria-orientation class if role is separator', async () => {
```
rachelbt, line 32:
```suggestion
		it('should set the aria-orientation attribute if role is separator', async () => {
```

libs/components/src/lib/divider/divider.spec.ts
```diff
@@ -28,6 +28,26 @@ describe('vwc-divider', () => {
...
+			const base = element.shadowRoot?.querySelector('.base');
+			const orientation = 'vertical';
+			element.orientation = orientation;
+			element.role = 'separator';
+			await elementUpdated(element);
+
+			expect(base?.getAttribute('aria-orientation')).toBe('vertical');
+		});
+
+		it('should NOT set the aria-orientation class if role is presentation', async () => {
```
rachelbt, line 42:
```suggestion
		it('should NOT set the aria-orientation attribute if role is presentation', async () => {
```

PR #2100, opened 2025-01-27
chore(docs): Searchable Select docs update (VIV-2264)

RichardHelm:


libs/components/src/lib/searchable-select/USE-CASES.md
```diff
@@ -0,0 +1,330 @@
+## Country Select
+
+```html preview 550px
```
YonatanKra, line 3: Maybe we can make this example simpler (and more realistic) by saving the data of the countries in a JSON and looping over them to generate the options?

YonatanKra (approved):
A small suggetion in the comments. If you choose to implement go ahead. Approved anyway cause it's good to go.
Nice work :)

PR #2101, opened 2025-01-27
fix(switch): aria-label is correctly handled (VIV-2292)

TaylorJ76:


libs/components/src/lib/switch/switch.spec.ts
```diff
@@ -297,12 +297,31 @@ describe('vwc-switch', () => {
...
+		});
+
+		it('should pass html a11y test with a standard label', async () => {
 			element.label = 'Label';
 			element.checked = true;
 			element.value = 'test';
 			await elementUpdated(element);
 
+			expect(await axe(element)).toHaveNoViolations();
+			expect(element.getAttribute('role')).toBe(null);
```
YonatanKra, line 316: The description says it should pass the a11y (`axe`) test.
What does that mean?

libs/components/src/lib/switch/switch.template.ts
```diff
@@ -16,22 +16,25 @@ const getClasses = (_: Switch) =>
...
-		aria-checked="${(x) => x.checked}"
-		aria-disabled="${(x) => x.disabled}"
-		aria-readonly="${(x) => x.readOnly}"
-		tabindex="${(x) => (x.disabled ? null : 0)}"
-		@keypress="${(x, c) => x.keypressHandler(c.event as KeyboardEvent)}"
-		@click="${(x) => x.clickHandler()}"
-	>
-		<div class="switch">
-			<span class="checked-indicator"></span>
+	<template role="${(x) => (x.ariaLabel ? 'presentation' : null)}">
```
YonatanKra, line 19: The template tag is definitely the way to go! 
image

YonatanKra (changes requested):
Two small comments

YonatanKra (approved):

PR #2108, opened 2025-01-29
chore: updates api reference docs

TaylorJ76:
- Removes padding from "Usage" tabs
- Aligns all API Reference table's first column formatting


rachelbt (approved):
👍

PR #2110, opened 2025-01-29
chore(time-picker): add inline-time-picker component (VIV-2294)

RichardHelm:


libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -0,0 +1,703 @@
...
+				pressKey('ArrowUp');
+				element.value = '23:00:00';
+				await elementUpdated(element);
+				pressKey('ArrowDown');
+				await elementUpdated(element);
+
+				expect(isScrolledIntoView(getPickerItem('hours', '00'))).toBe(true);
+			});
+		});
+	});
```
YonatanKra, line 653: Aren't these a11y traits? WDYT?

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.ts
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	@observable minutesStep = 1;
+	@observable secondsStep?: number;
+	@observable clock: '12h' | '24h' = '24h';
+	@observable min?: string;
+	@observable max?: string;
+
+	scrollSelectedOptionsToTop() {
```
YonatanKra, line 33: Why is that in the class?

YonatanKra (approved):
Nice one! Thanks a lot!

PR #2113, opened 2025-01-30
chore(docs): split-button docs update (VIV-2106)

TaylorJ76:


libs/components/src/lib/split-button/VARIATIONS.md
```diff
@@ -0,0 +1,152 @@
+## Labelling
+
+### Label
+
+The `label` attribute adds the label text for the default action.
+
```
rachelbt, line 6: its a bit strange that the code sample is only after the secondary action label. It make sense but also seems like we forgot to add the code :)

libs/components/src/lib/split-button/VARIATIONS.md
```diff
@@ -0,0 +1,152 @@
...
+The `appearance` sets the Split Button's appearance.
+
+```html preview
+<vwc-split-button label="ghost" appearance="ghost"></vwc-split-button>
+<vwc-split-button label="filled" appearance="filled"></vwc-split-button>
+<vwc-split-button label="outlined" appearance="outlined"></vwc-split-button>
+```
+
+## Connotation
+
```
rachelbt, line 61: its better to wrap the buttons with flex container and gap - in small screens the buttons are tight at each other:
image

same for appearance, size, and disabled

libs/components/src/lib/split-button/USE-CASES.md
```diff
@@ -0,0 +1,51 @@
...
+		splitButton.addEventListener('indicator-click', () => {
+			menu.open = !menu.open;
+		});
+	};
+</script>
+```
+
+## Tooltip
+
+```html preview 100px
```
rachelbt, line 30:
```suggestion
```html preview 140px
```

rachelbt (changes requested):

PR #2114, opened 2025-01-31
chore(components): remove remaining uses and mentions of FAST Foundation (VIV-2295)

RichardHelm:


libs/components/src/lib/menu-item/menu-item.ts
```diff
@@ -106,14 +105,6 @@ export class MenuItem extends VividElement {
 		}
 	}
 
-	/**
-	 * reference to the anchored region
-	 *
-	 * @internal
-	 */
-	@observable
-	submenuRegion!: AnchoredRegion;
```
YonatanKra, line 115: If this one isn't needed, why was it here at all? Didn't it have any purpose?

RichardHelm: No 🤷‍♂️
Inherited from FAST probably

TaylorJ76 (approved):

PR #2124, opened 2025-02-06
chore(components): separate a11y tests (VIV-2312)

TaylorJ76:
- Moves component axe test into a separate *.a11y.spec.ts file
- Adds a new command to run the a11y tests `npm run test:a11y` or `npx nx run components:test:a11y`
- Updates the component generator files with the new a11y test file


libs/components/src/lib/banner/banner.spec.ts
```diff
@@ -311,35 +311,12 @@ describe('vwc-banner', () => {
...
-				expect(await axe(element)).toHaveNoViolations();
-			});
-
-			it('should change role when role attribute is set', async function () {
-				element.setAttribute('role', 'alert');
-				await elementUpdated(element);
-				const role = getBannerMessageAttribute(element, 'role');
-				expect(role).toEqual('alert');
-				expect(await axe(element)).toHaveNoViolations();
-			});
```
YonatanKra, line 342: Were these moved someplace? It seems like they were deleted, even though the functionality is still there (hence the coverage is reduced).

YonatanKra (changes requested):
There's one issue I found with the `banner` in which the tests were removed and not moved, and it caused the coverage to drop (it caught a true positive :) ).
I believe all the places where there were non-axe tests should be checked and make sure they were not removed.
In addition, there's now an issue with `listbox.length`. Because we don't have tests for our `foundation`, we need to at least test this API through its extenders (our components), so we need to find who's extending `foundation/listbox` and make sure there's a test for `length`. 
We are already supposed to take care of foundation during our tech debt tasks, so no need for a overhaul (the PR is already huge as is).

libs/components/src/lib/radio-group/radio-group.spec.ts
```diff
@@ -442,4 +443,13 @@ describe('vwc-radio-group', () => {
...
 	});
+
+	describe('a11y', () => {
+		it('should pass html a11y test', async () => {
+			element.label = 'Label';
+			await elementUpdated(element);
+
+			expect(await axe(element)).toHaveNoViolations();
+		});
+	});
```
YonatanKra, line 454: Shouldn't this be in its own `a11y` test file?

libs/components/src/lib/select/select.spec.ts
```diff
@@ -1288,4 +1293,19 @@ describe('vwc-select', () => {
...
+				<option value="2">2</option>
+				<option value="3">3</option>
+			`;
+			element.selectedIndex = 2;
+			element.label = 'Label';
+			await elementUpdated(element);
+
+			expect(await axe(element)).toHaveNoViolations();
+		});
+	});
```
YonatanKra, line 1310: Should this be in its own `a11y` file?

YonatanKra (changes requested):
I requested 2 small specific changes, but reading through the files I think we should do this manually and gradually.
There are tests that require a move ot their own API describe. An example for this from the accordion is this test:
```
		it('should set aria-expanded to false when closed', async () => {
			const button = element.shadowRoot?.querySelector('.heading-button');
			expect(button?.getAttribute('aria-expanded')).toEqual('false');
		});
		it('should set aria-expanded to true when expanded', async () => {
			element.expanded = true;
			const button = element.shadowRoot?.querySelector('.heading-button');
			expect(button?.getAttribute('aria-expanded')).toEqual('true');
		});
```
The above use cases should be under the `open` interface, as they are a result of the user interacting with the component (e.g. the open interface).
We can merge this and start covering the components when we have time (event one/day, which should take around 5 minutes, should get us there fast enough).
WDYT?

TaylorJ76:
> I requested 2 small specific changes, but reading through the files I think we should do this manually and gradually. There are tests that require a move ot their own API describe. An example for this from the accordion is this test:
> 
> ```
> 		it('should set aria-expanded to false when closed', async () => {
> 			const button = element.shadowRoot?.querySelector('.heading-button');
> 			expect(button?.getAttribute('aria-expanded')).toEqual('false');
> 		});
> 		it('should set aria-expanded to true when expanded', async () => {
> 			element.expanded = true;
> 			await elementUpdated(element);
> 			const button = element.shadowRoot?.querySelector('.heading-button');
> 			expect(button?.getAttribute('aria-expanded')).toEqual('true');
> 		});
> ```
> 
> The above use cases should be under the `open` interface, as they are a result of the user interacting with the component (e.g. the open interface). We can merge this and start covering the components when we have time (event one/day, which should take around 5 minutes, should get us there fast enough). WDYT?

Yes. I'm fine with that plan.

PR #2132, opened 2025-02-12
fix(file-picker): set aria-label on base (VIV-2326)

rachelbt:


libs/components/src/lib/file-picker/file-picker.template.ts
```diff
@@ -18,7 +18,7 @@ export const FilePickerTemplate = (context: VividElementDefinitionContext) => {
 			x.setButtonTag(context.tagFor(Button));
 			x.setIconTag(context.tagFor(Icon));
 		}}
-		<div class="base" aria-label="${(x) => x.label}">
+		<div class="base" aria-label="${(x) => x.ariaLabel}">
```
RichardHelm, line 21: Why add it at all? `aria-label` takes effect on the host anyway
```suggestion
		<div class="base">
```

RichardHelm (changes requested):

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -104,13 +103,12 @@ describe('vwc-file-picker', () => {
 			expect(labelElement).toBeNull();
 		});
 
-		it('should set aria-label on base element if label is set', async function () {
-			const labelText = 'label';
-			element.label = labelText;
+		it('should set aria-label on .base if aria-label is set', async () => {
```
RichardHelm, line 106: Well, now we're just adding the aria-label on the host and checking that it's there
I think the test can be removed as well now :)

RichardHelm (changes requested):

PR #2134, opened 2025-02-12
fix(slider, range-slider): handle zero step (VIV-2271)

TaylorJ76:


libs/components/src/lib/range-slider/utils/roundToStepValue.ts
```diff
@@ -18,3 +18,15 @@ export const roundToStepValue = (value: number, step: number) => {
 	const roundedValue = Math.round(value / step) * step;
 	return Math.round(roundedValue * factor) / factor;
 };
+
+export const handleStepValue = (step: number) => (step > 0 ? step : 1);
```
YonatanKra, line 22: IMO this should not be exported. It is an implementation detail.
We can test the `defaultToOneConverter` which uses it, because this is the interface we expose.

libs/components/src/lib/range-slider/utils/roundToStepValue.spec.ts
```diff
@@ -14,3 +14,14 @@ describe('roundToStepValue', () => {
 		expect(roundToStepValue(0.1 + 0.1 + 0.1, 1e-7)).toBe(0.3);
 	});
 });
+
+describe('handleStepValue', () => {
```
YonatanKra, line 18: According to the other comment, this one should test `defaultToOneConverter`, instead and leave `handleStepValue` as a private function. Is `handleStepValue` used anywhere else?

YonatanKra (changes requested):

PR #2136, opened 2025-02-12
fix(tabs): fix connotation not updating when tab is added (VIV-2334)

RichardHelm:


libs/components/src/lib/tabs/tabs.spec.ts
```diff
@@ -511,20 +511,20 @@ describe('vwc-tabs', () => {
 			checkConnotationDoesntExistOnNonActiveTabs();
 		});
 
-		it('should reflect connotation on active tab after tabs changed', async function () {
```
rachelbt, line 514: why did you changed the test and not added a new test?

RichardHelm: The previous test didn't catch the issue because it didn't test anything. When adding a tab at the end nothing changes, so the component not doing anything passed the test
It's only relevant when the new tab becomes the active tab but the `activeid` stays the same that we need to test. If adding/removing a tab changes `activeid` then it's already covered by the `activeid` test

TaylorJ76 (approved):

PR #2145, opened 2025-02-14
chore(docs): updates note docs format (VIV-2117)

TaylorJ76:


libs/components/src/lib/note/README.md
```diff
@@ -1,89 +1,56 @@
...
-	icon="sparkles-solid"
-	headline="announcement note"
-></vwc-note>
-<vwc-note
-	connotation="accent"
-	icon="megaphone-solid"
-	headline="accent note"
-></vwc-note>
+<template>
+	<VNote label="First name" />
```
rachelbt, line 44: note doesn't have label :)

rachelbt (changes requested):
you are missing a Safari UI test :)

PR #2146, opened 2025-02-14
chore(inline-time-picker): support focus and focus traps (VIV-2296)

RichardHelm:


libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -596,5 +596,33 @@ describe('vwc-inline-time-picker', () => {
 				expect(isScrolledIntoView(getPickerItem('hours', '00'))).toBe(true);
 			});
 		});
+
+		describe('focus trap support', () => {
```
YonatanKra, line 600: IMO describes should describe API's.
I know there's more like this in our codebase. It should be refactored.
Anyway, it describes nothing but a single use case so should be removed and its it remain under keyboard interactions.

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -596,5 +596,33 @@ describe('vwc-inline-time-picker', () => {
...
+			const originalIgnoreEvent = TrappedFocus.ignoreEvent;
+			beforeEach(() => {
+				TrappedFocus.ignoreEvent = vi.fn();
+			});
+			afterEach(() => {
+				TrappedFocus.ignoreEvent = originalIgnoreEvent;
+			});
+
+			it('should submit Tab keydown events that move focus internally to be ignored by focus traps', () => {
```
YonatanKra, line 609: What I see in this test is:
Focus on hours, press tab, expect ignoreEvent method to be called once with the Tab event.
Can you please write the description in the form of: `should {doSomething} when {someThing}`?

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.template.ts
```diff
@@ -82,6 +79,26 @@ export const scrollToOption = (
 	scrollIntoView(element, element.parentElement!, position);
 };
 
+const onBaseKeyDown = (x: InlineTimePicker, event: KeyboardEvent) => {
+	if (event.key === 'Tab') {
+		const focusableEls = x.shadowRoot!.querySelectorAll('.picker');
```
YonatanKra, line 84: `focusableElements` is clearer IMO.

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.ts
```diff
@@ -70,6 +70,14 @@ export class InlineTimePicker extends VividElement {
...
 		}
 	}
+
+	override focus(options?: FocusOptions) {
+		// Override focus instead of relying on default behavior to prevent visible focus
+		const firstFocusableElement = this.shadowRoot!.querySelector(
+			'.picker'
+		) as HTMLElement;
+		firstFocusableElement.focus(options);
+	}
```
YonatanKra, line 80: I could not find a test for this one. Was it tested?

YonatanKra (changes requested):
I commented on the changes for `inline-time-picker`.
We all agreed to work in small interations. That means, each PR should have a single purpose and as few files changed as possible.
To do that, we should finish this PR for the `inline-time-picker` focus handling, merge it, and then move on to using `inline-time-picker`.

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
RichardHelm: Well, I think we should use describe to group related behaviour. This is similar to how we have `describe("form association")` in the codebase.

I don't want to move this under "time picker > keyboard navigation" as it has nothing to do with keyboard navigation for picking times

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
RichardHelm: And how will you understand the purpose without this description? I think it will not become clearer by phrasing it like that

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.ts
RichardHelm: Sure, I added a test

RichardHelm:
@YonatanKra Sure, I've split the PR

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
YonatanKra: What scenario does it test? What user action?

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -329,6 +329,17 @@ describe('vwc-inline-time-picker', () => {
 				expect.objectContaining({ bubbles: false })
 			);
 		});
+
+		it('should not be composed', () => {
```
YonatanKra, line 333:
```suggestion
		it('should emit the event with composed set to false', () => {
```

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -595,6 +606,51 @@ describe('vwc-inline-time-picker', () => {
...
+					(element.shadowRoot!.querySelector('#hours') as HTMLElement).focus();
+
+					const event = pressKey('Tab');
+
+					expect(TrappedFocus.ignoreEvent).toHaveBeenCalledTimes(1);
+					expect(TrappedFocus.ignoreEvent).toHaveBeenCalledWith(event);
+				});
+
+				it('should not call focus trap ignoreEvent on tab keydown when focused on terminal picker element', () => {
+					(element.shadowRoot!.querySelector('#hours') as HTMLElement).focus();
```
YonatanKra, line 629: I think we should test the other case in which we do not call `ignoreEvent` when in the last element.

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -595,6 +606,51 @@ describe('vwc-inline-time-picker', () => {
...
+				it('should call focus trap ignoreEvent on tab keydown when focused on non-terminal picker element', () => {
+					(element.shadowRoot!.querySelector('#hours') as HTMLElement).focus();
+
+					const event = pressKey('Tab');
+
+					expect(TrappedFocus.ignoreEvent).toHaveBeenCalledTimes(1);
+					expect(TrappedFocus.ignoreEvent).toHaveBeenCalledWith(event);
+				});
+
+				it('should not call focus trap ignoreEvent on tab keydown when focused on terminal picker element', () => {
```
YonatanKra, line 628:
```suggestion
				it('should prevent call to focus trap ignoreEvent on tab keydown when focused on terminal picker element', () => {
```

YonatanKra (changes requested):
Final small changes :)

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
RichardHelm: Mhm, I know that you believe that tests shouldn't contain the word not in them, but this is just a less clear description

YonatanKra: Clearer to me and describes the actual test code

RichardHelm: We're in the describe for the event so it would need to be `should be emitted with composed set to false`.
It's not a big deal but I think `should not be composed` is definitely more clear so I don't want to change it

YonatanKra: If it's not a big deal, please change it to what my old mind will be able to better understand in 3 weeks :)

RichardHelm:
```suggestion
		it('should be emitted with composed set to false', () => {
```

YonatanKra (approved):
Thanks :)

libs/components/src/lib/time-picker/inline-time-picker/inline-time-picker.spec.ts
```diff
@@ -1,5 +1,7 @@
 import { elementUpdated, fixture } from '@vivid-nx/shared';
-import { InlineTimePicker } from './inline-time-picker.ts';
+import { beforeEach } from 'vitest';
```
YonatanKra, line 2: I approved. Just notice we don't need this line because `vitest` globals is set to `true` in the config.

RichardHelm:
```suggestion
```

PR #2163, opened 2025-02-21
chore(components): support standard mixins (VIV-000)

RichardHelm:
As an example, I have converted `anchored` to use the mixin pattern


libs/components/cem.config.mjs
```diff
@@ -155,5 +161,51 @@ export default {
...
+		/**
+		 * This plugin makes certain features (e.g. @attr/@event/@slot) work with mixins.
+		 * The analyzer has built in support for mixins, but it is flawed: the mixin declaration is created from the inner
+		 * class when the mixin is first encountered. That means that later plugins, like the @attr decorator plugin, will
+		 * not update the mixin. However, the inner class is still present in the moduleDoc and will be updated with the
+		 * attribute. The inner class is later dropped from the manifest as it is not exported. This plugin copies the
+		 * updated declarations from the inner class to the mixin before that happens.
+		 */
+		{
+			name: 'Improve Mixin Support',
```
YonatanKra, line 173: The amount and complexity of the plugins is starting to become a bit of an issue IMO.
Can we set this as a new libary for `cem plugins` and also test them please?
I'll stretch it a bit - maybe they can also be published as a public open source NPM module?

libs/wrapper-gen/src/generator/customElementDeclarations.ts
```diff
@@ -493,7 +502,7 @@ export const getClassNameOfVividComponent = (name: string): string => {
 export const getVividComponentDeclaration = (
```
YonatanKra, line 502: Since we introduced new cases here, let's cover them with tests.

YonatanKra (changes requested):
Nice work! A well needed update with TS changes and new best practices.
Just a few minor standard requests.

libs/components/cem.config.mjs
```diff
@@ -21,6 +26,7 @@ export default {
 		'**/enums.ts',
 		'**/components.ts',
 		'**/helpers/*.ts',
+		'libs/components/src/shared/patterns/affix.ts', // Cannot process @attr that is @internal
```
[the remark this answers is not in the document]
RichardHelm: No, affix is not the converted mixin (anchored). It's just that this file trips up the analyser until we convert affix as well

libs/wrapper-gen/src/generator/customElementDeclarations.ts
RichardHelm: Well, we're going to start covering this code with tests soon anyway, so I'm not going to add them now

libs/wrapper-gen/src/generator/customElementDeclaration.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+					'../../dist/libs/components/',
+					__dirname + '/__fixtures__/accordion/'
+				),
+				...rest
+			)
+		),
+	};
+});
+
+describe('getVividComponentDeclaration', () => {
```
YonatanKra, line 21: I know this is legacy code, but we should make it better on every iteration.
Let's make this clearer by adding more concise example for each case and test for the actual output.
Expecting that there is `some attribute` is too broad and does not describe what this (concerns-bloated exposed function) is doing.
As I mentioned - tests should be described as `should {doSomthing} when {somethingHappens}`

YonatanKra (changes requested):

libs/wrapper-gen/src/generator/customElementDeclaration.spec.ts
RichardHelm: Sure, I changed it to test for the actual values.
I don't want to create individual examples, as it currently insures that the inheritance doesn't destroy the component's data. Also 9 individual examples would just bloat the test.
Nothing happens in the tests so a when clause is not needed.

YonatanKra (approved):
Nice work. Thanks!

PR #2164, opened 2025-02-24
fix(menu-item): handle menu-item getting checked synchronously (VIV-2352)

RichardHelm:


TaylorJ76 (dismissed):

libs/components/src/lib/menu/menu.spec.ts
```diff
@@ -719,6 +719,18 @@ describe('vwc-menu', () => {
 
 			expect(item.checked).toBe(true);
 		});
+
+		it('should handle menu items being synchronously checked after being connected', async () => {
```
YonatanKra, line 723: `handle` is too broad, IMO.
Is this description true?
```suggestion
		it('should gracefully fail changing menu item checked when element disconnected', async () => {
```

YonatanKra (changes requested):
image
Just a small semantic issue :)

YonatanKra (approved):
Thanks :)

PR #2176, opened 2025-02-26
chore(docs): updates pagination docs to new format (VIV-2120)

TaylorJ76:


libs/components/src/lib/pagination/VARIATIONS.md
```diff
@@ -0,0 +1,49 @@
...
+
+```html preview
+<vwc-pagination total="20"></vwc-pagination>
+```
+
+## Selected Index
+
+The `selected-index` attribute sets the currently selected page.
+
+Note that the value is zero-indexed. It is set to `-1` if total pages is `0`.
```
rachelbt, line 13: this can be an info note :)

PR #2183, opened 2025-03-03
feat(rich-text-editor): value property and attribute (VIV-2378)

YonatanKra:
It should accept and return the HTML value inside the editor


rachelbt (commented):
👍 
You need to fix your linting :)

rachelbt (approved):

PR #2184, opened 2025-03-03
chore(docs): update menu to new format (VIV-2102)

TaylorJ76:


libs/components/src/lib/menu/VARIATIONS.md
```diff
@@ -0,0 +1,254 @@
...
+
+</vwc-note>
+
+## Role
+
+The `role` attribute on **Menu Item** sets its role. It's default value is `menuitem`. See the [API Reference](/component/menu/code/#menu-item) for all possibile values.
+
+```html preview 330px
+<vwc-menu open aria-label="Example menu">
+	<vwc-menu-item role="menuitem" text="Menu item 1"></vwc-menu-item>
```
rachelbt, line 96: maybe it will be clearer if we will change the text to:  role menuitem 1, role menuitem 2.... role menuitemcheckbox 1 and more

libs/components/src/lib/menu/README.md
```diff
@@ -1,43 +1,81 @@
...
+## Usage
 
-Menu, commonly known as a "context menu", is an element that is displayed upon user interaction. It is typically used to provide a list of actions available in the current context for a user to choose from.
+<vwc-tabs gutters="none">
+<vwc-tab label="Web component"></vwc-tab>
+<vwc-tab-panel>
 
 ```js
-<script type="module">import '@vonage/vivid/menu';</script>
+import '@vonage/vivid/layout';
```
rachelbt, line 8:
```suggestion
import '@vonage/vivid/menu';
import '@vonage/vivid/menu-item';
```

libs/components/src/lib/menu/README.md
```diff
@@ -164,19 +163,20 @@ In vivid version 4.12.0, popover attribute was added to menu, using the power of
 </div>
 ```
 
-### Anchor
+## Anchor
 
-It's best to use the [`anchor` slot](#anchor-1) to set the anchor, but you can also use the `anchor` member.
+<vwc-note connotation="warning" icon="warning-line" headline="Prefer using the anchot slot">
```
rachelbt, line 168: its a bit confusing having 2 notes of warning. Maybe this one can be info?

libs/components/src/lib/menu/README.md
```diff
@@ -262,16 +268,108 @@ Use the `action-items` slot to add action items to the bottom of the menu.
...
+			<vwc-menu-item text="Menu item 3.2"></vwc-menu-item>
+			<vwc-menu-item text="Menu item 3.3"></vwc-menu-item>
+		</vwc-menu>
+	</vwc-menu-item>
+</vwc-menu>
+```
+
+<vwc-note headline="Focus Management" connotation="information" icon="info-line">
+
+When the Menu opens or `.focus()` is called, focus moves to the first Menu Item by default. If there is a child with the `autofocus` attribute, it will be focused instead.
```
rachelbt, line 359: since when is this?
Why are we presenting the focus -it looks strange. Its not an input field.

libs/components/src/lib/menu/README.md
```diff
@@ -331,150 +429,124 @@ Use the `--menu-block-size` variable to set the menu's block size.
...
+| **open**             | `boolean`                                                                                                                                                               | Sets the open state of the Menu                                       |
+| **placement**        | `left-start`, `left-center` `left-end`, `right-start`, `right-center`, `right-end`, `top-start`, `top-center`, `top-end`, `bottom-start`, `bottom-center`, `bottom-end` | Sets the desired position of the Menu relative to it's anchor element |
+| **position-stategy** | `fixed` (deafult), `absolute`                                                                                                                                           | Sets the position strategy                                            |
+| **trigger**          | `none`, `legacy` (default), `auto`                                                                                                                                      | Sets trigger method of Menu                                           |
```
rachelbt, line 473: `anchor` is missing

libs/components/src/lib/menu/README.md
```diff
@@ -331,150 +429,124 @@ Use the `--menu-block-size` variable to set the menu's block size.
...
+#### Slots
 
-When the Menu opens or `.focus()` is called, focus moves to the first Menu Item by default. If there is a child with the `autofocus` attribute, it will be focused instead.
+<div class="table-wrapper">
 
-## Use Cases
+| Name        | Description                                                             |
+| ----------- | ----------------------------------------------------------------------- |
+| **meta**    | For Menu Item components or HTML elements with the `role` or `menuitem` |
+| **submenu** | To add Menus as sub-menus                                               |
```
rachelbt, line 539: trailing-meta is missing

libs/components/src/lib/menu/README.md
```diff
@@ -331,150 +429,124 @@ Use the `--menu-block-size` variable to set the menu's block size.
...
-			role="presentation"
-			text="Logout"
-			icon="quit-line"
-		></vwc-menu-item>
-	</a>
-</vwc-menu>
-```
+| Name                | Type                       | Bubbles | Composed | Description                                                                                                                        |
+| ------------------- | -------------------------- | ------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| **expanded-change** | `CustomEvent<HTMLElement>` | No      | Yes      | Fires a custom 'expanded-change' event when the expanded state changes                                                             |
```
rachelbt, line 549: these two are not in the current documentation
and this is missing: https://vivid.deno.dev/components/menu-item/#click

rachelbt (changes requested):
for some reasons the deno-dev didn't get the Spezia font.

I'm not so sure about the mix of menu & menu-item, I find it really confusing here. The face that the long list of slot made you add the name of the component shows its complicated. Maybe we shouldn't do them together.

libs/components/src/lib/menu/README.md
TaylorJ76: Since this task: https://jira.vonage.com/browse/VIV-2196

libs/components/src/lib/menu-item/VARIATIONS.md
```diff
@@ -0,0 +1,219 @@
...
+
+```html preview 100px
+<vwc-menu open aria-label="Example menu">
+	<vwc-menu-item icon="file-pdf-line" text="Export to PDF"></vwc-menu-item>
+</vwc-menu>
+```
+
+<vwc-note connotation="warning" icon="warning-line" headline="Accessibility note">
+
+Icon only Menu Items need an accessible label. An `aria-label` or `title` must be provided to ensure that the user can understand the Menu Item's purpose.
```
rachelbt, line 44: This note should be removed:
- we do not have a use case for icon only in menu-item (like we do have on nav).
-  icon by itself can't have aria-label and anyway the icon in menu-item has aria-hidden="true".
- The icon has no meaning here but decoration, no need to have description for it - its already there in the text.
- Even if someone would want to add aria-label on the icon - they can't because they can't reach the icon in the shadow Dom :)

rachelbt (commented):
I think its so much clearer when they are not together.
Left only one last comment :)

PR #2191, opened 2025-03-05
feat(combobox, searchable-select): popup closes when chevron icon is clicked (VIV-2367)

TaylorJ76:


libs/components/src/lib/searchable-select/searchable-select.ts
```diff
@@ -1036,7 +1037,16 @@ export class SearchableSelect extends FormAssociatedSearchableSelect {
 		}
 
 		if (!e.defaultPrevented) {
-			// Unless something was clicked on that handled the event, e.g. tag remove button, focus the input when clicking anywhere
+			const capturedChevron = (e.target as HTMLElement).closest(
+				`vwc-icon.chevron`
```
RichardHelm, line 1041: Will not work with dynamic prefix. IMO it should intercept click on the chevron and preventDefault

libs/components/src/lib/combobox/combobox.ts
```diff
@@ -338,6 +338,18 @@ export class Combobox extends FormAssociatedCombobox {
 		return true;
 	}
 
+	/**
+	 * Handle closing the listbox when the combobox is open and the chevron icon is clicked.
+	 *
+	 * @param e - the mouse event
+	 * @internal
+	 */
+	chevronIconClickHandler(e: MouseEvent): void {
```
RichardHelm, line 347:
```suggestion
	_chevronIconClickHandler(e: MouseEvent): void {
```

libs/components/src/lib/searchable-select/searchable-select.ts
RichardHelm, line 1: Can we add the cursor style from the combobox as well? I think it can go in a shared mixin in chevron.scss

RichardHelm (changes requested):

RichardHelm (approved):

PR #2196, opened 2025-03-09
feat(rich-text-editor): basic visual interface (VIV-2135)

YonatanKra:
chore(rich-text-editor): add prose mirror inside a facade


libs/components/src/lib/rich-text-editor/ui.test.ts
```diff
@@ -22,6 +22,8 @@ test('should show the component', async ({ page }: { page: Page }) => {
 
 	await page.waitForLoadState('networkidle');
 
+	await page.pause();
```
rachelbt, line 25: Why do you need this?

YonatanKra: I don't necessarily need this, but right now development is done using the ui tests mechanism, so I leave it here.
It will be removed when we finalize the ui tests. I'll add this task to when I set up the documentation.

rachelbt: just don't forget to remove it 😉

rachelbt (approved):

PR #2200, opened 2025-03-09
feat(rich-text-editor): can set selection from the component (VIV-2386)

YonatanKra:
Can now use `element.selectionStart = 5` and `element.selectionEnd = 10`.


libs/components/src/lib/rich-text-editor/rich-text-editor.ts
```diff
@@ -33,10 +33,36 @@ export class RichTextEditor extends VividElement {
 		return this.shadowRoot!.querySelector('#editor') as HTMLElement;
 	}
 
-	get selection(): RichTextEditorSelection | undefined {
-		return this.#editor?.selection();
+	@attr({ converter: nullableNumberConverter, attribute: 'selection-start' })
+	selectionStart: number | null = null;
+	selectionStartChanged() {
```
RichardHelm, line 38: It's needs the despised `@internal` or it will be public
```suggestion
	selectionStartChanged() {
```

RichardHelm (approved):
Just the `@internals` is missing. But it can be added in later PRs if easier

PR #2210, opened 2025-03-12
chore(components): check in metadata (VIV-2288)

RichardHelm:


YonatanKra (approved):
LGTM. 
Is there a way to raise a warning locally? Via a precommit hook or something?
Could be a nice addition to alert someone that one's change caused the issue before pusing to github. Do you think it is worth a new issue?

RichardHelm:
@YonatanKra Mhm, I think it will be too slow because it needs to build the components. That's why I added the updating to the CI.
I think for now we should see if it will be an issue first, and then figure something out

PR #2211, opened 2025-03-12
chore(rich-text-editor): user selection updates the component (VIV-2386)

YonatanKra:
In the image below, see `selectionStart` and `selectionEnd`.
image


rachelbt (approved):
I know that there's no style for selected but maybe there's a need either to add this as a comment that we will not forget or actually add a ui-test for selected string. WDYT?

PR #2229, opened 2025-03-17
chore(rich-text-editor): expose input event from the component (VIV-2318)

YonatanKra:


libs/components/src/lib/rich-text-editor/facades/vivid-prose-mirror.facade.spec.ts
```diff
@@ -279,6 +279,23 @@ describe('ProseMirrorFacade', () => {
...
+			await useOriginalEditorView();
+			const element = document.createElement('div');
+			facadeInstance.init(element);
+			facadeInstance.replaceContent(
+				'<p>This is a pretty long text for a sample, but it should work</p>'
+			);
+
+			const spy = vi.fn();
+			facadeInstance.addEventListener('input', spy);
+			getOutputElement(element).dispatchEvent(new Event('input'));
```
RichardHelm, line 293: Can you emit with the same flags as in reality (bubbles and composed).
Because the event will already bubble and trigger on the RTE on its own right?

RichardHelm (changes requested):

libs/components/src/lib/rich-text-editor/facades/vivid-prose-mirror.facade.spec.ts
YonatanKra: Nice catch! Thanks!

PR #2235, opened 2025-03-18
fix(alert): not announced correctly in screen reader (VIV-1970)

TaylorJ76:
When the dismiss button is present, it receives focus when the alert is opened.
This is beneficial to keyboard only and screen reader users, because they don't have to tab around to find the button.


libs/components/src/lib/alert/alert.scss
```diff
@@ -118,3 +118,7 @@ $alert-transition-delay: --transition-delay;
 		padding-inline-start: 8px;
 	}
 }
+
+[role='alert']:empty {
```
rachelbt, line 122: this is a bit risky - because if in the IFE there's a line break or a space it can be rendered and it will not be empty

TaylorJ76: Thanks. It wasn't needed in the end.

libs/components/src/lib/alert/alert.template.ts
```diff
@@ -44,17 +45,19 @@ export const AlertTemplate = (context: VividElementDefinitionContext) => {
...
+				role="alert" 
+				aria-hidden="${(x) => (x.open ? 'false' : 'true')}" 
+				${(x) => (!x.open ? 'hidden' : '')}
+			>
 				${renderIcon(context)}
 				<div class="alert-text">
 					${when(
 						(x) => x.headline,
-						html`<header class="headline">${(x) => x.headline}</header>`
+						html`<div class="headline">${(x) => x.headline}</div>`
```
rachelbt, line 60: isn't it an issue for readability of the alert?

libs/components/src/lib/alert/alert.template.ts
```diff
@@ -66,3 +69,11 @@ export const AlertTemplate = (context: VividElementDefinitionContext) => {
...
 	`;
 };
+
+// export const AlertTemplate = (context: VividElementDefinitionContext) => {
+// 	return html<Alert>`
+// 		<div role="alert">
+// 			${(x) => x.open ? x.text : null}
+// 		</div>
+// 	`;
+// };
```
rachelbt, line 79:
```suggestion
```

libs/components/src/lib/alert/alert.ts
```diff
@@ -134,13 +134,11 @@ export class Alert extends VividElement {
 		this.#setupTimeout();
 		if (newValue) {
 			this.style.display = 'contents';
-			const alertText = this.shadowRoot!.querySelector(
-				'.alert-text'
+			const closeBtn = this.shadowRoot!.querySelector(
+				'.dismiss-button'
 			) as HTMLElement;
-			if (this.removable && alertText) {
-				alertText.setAttribute('tabindex', '0');
```
rachelbt, line 141: the tabindex is not needed anymore?

libs/components/src/lib/alert/alert.template.ts
```diff
@@ -44,17 +45,19 @@ export const AlertTemplate = (context: VividElementDefinitionContext) => {
 
 	return html<Alert>`
 	<${elevationTag} class="elevation" dp='8' exportparts="vvd-theme-alternate">
-		<div
-			class="${getControlClasses}"
-			role="${(x) => (x.removable ? 'alertdialog' : 'alert')}"
```
rachelbt, line 49: no need for `alertdialog`?

TaylorJ76: No. It's not an alertdialog (like a confirmation dialog), it is a closable alert.
I checked how other design system's tackled it and this seemed to be the most common.
Also, it works as expected.

PR #2243, opened 2025-03-25
chore(rich-text-editor): add text size button to menubar (VIV-2464)

YonatanKra:


libs/components/src/lib/rich-text-editor/menubar/menubar.scss
```diff
@@ -1 +1,12 @@
...
+	--vvd-typography-base: var(--vvd-typography-heading-4);
+}
+
+vwc-menu-item[value='subtitle'] {
+	--vvd-typography-base: var(--vvd-typography-base-extended);
+}
+
+vwc-menu-item[value='base'] {
+	--vvd-typography-base: var(--vvd-typography-base);
+}
```
rachelbt, line 12: this is a bit riskey - what if one day we will change --vvd-typography-base?
Maybe its better to add style like we did in text-filed for dial-pad:

/* Customize for menu-bar*/
:host([internal-part]) {
.text-primary {
	font: var(--new variable-for-text-size, #{constants.$vvd-typography-base});
}
}

YonatanKra: Looks like YAGNI to me. 
And if we change `typography-base`, this will probably change to it :)

libs/components/src/lib/rich-text-editor/menubar/menubar.template.ts
```diff
@@ -1,11 +1,73 @@
...
+	eventName: string,
+	payload: any
+): any {
+	menuBar.$emit(eventName, payload, { bubbles: false, composed: false });
+	return true;
+}
+
+const MENU_BAR_ITEMS: { [key: string]: ViewTemplate<any, any> } = {
+	textSize: html`
+		<vwc-menu
```
rachelbt, line 18: this should be ${menuTag} and not vwc-menu.
Same for the button and menu-item

libs/components/src/lib/rich-text-editor/menubar/menubar.template.ts
```diff
@@ -1,11 +1,73 @@
...
+				value="subtitle"
+				@click="${(_, { parent }) =>
+					notifyMenuBarChange(parent, 'text-size-selected', 'subtitle')}"
+			></vwc-menu-item>
+			<vwc-menu-item
+				text="Body"
+				value="body"
+				@click="${(_, { parent }) =>
+					notifyMenuBarChange(parent, 'text-size-selected', 'body')}"
+			></vwc-menu-item>
```
rachelbt, line 48:
```suggestion
			<vwc-menu-item
				text="Title"
				connotation="cta"
				value="title"
				@click="${(_, { parent }) =>
					notifyMenuBarChange(parent, 'text-size-selected', 'title')}"
			></vwc-menu-item>
			<vwc-menu-item
				text="Subtitle"
				connotation="cta"
				value="subtitle"
				@click="${(_, { parent }) =>
					notifyMenuBarChange(parent, 'text-size-selected', 'subtitle')}"
			></vwc-menu-item>
			<vwc-menu-item
				text="Body"
				connotation="cta"
				value="body"
				@click="${(_, { parent }) =>
					notifyMenuBarChange(parent, 'text-size-selected', 'body')}"
			></vwc-menu-item>
```

rachelbt: only added connotation

rachelbt (commented):
want to point out that in the design there's a selected state but role menu do not have it:
image

PR #2251, opened 2025-03-27
chore(rich-text-editor): text decoration buttons active in the menubar (VIV-2370)

YonatanKra:
Notice that the `monospace` icon is currently not present and will appear once it will be added to the icon repo.


rachelbt (approved):
👍

PR #2256, opened 2025-03-31
chore(ci): typecheck test code (VIV-2311)

RichardHelm:


libs/components/src/lib/date-picker/ui.test.ts
```diff
@@ -5,6 +5,7 @@ import {
 	loadTemplate,
 	useFakeTime,
 } from '../../visual-tests/visual-tests-utils.js';
+import type { DatePicker } from './date-picker.ts';
```
TaylorJ76, line 8: Does this still need the `.ts`?

libs/components/src/lib/date-time-picker/ui.test.ts
```diff
@@ -5,6 +5,7 @@ import {
 	loadTemplate,
 	useFakeTime,
 } from '../../visual-tests/visual-tests-utils.js';
+import type { DateTimePicker } from './date-time-picker.ts';
```
TaylorJ76, line 8: Does this still need the `.ts`?

TaylorJ76 (changes requested):

RichardHelm:
@TaylorJ76 They are are not needed (or not needed) so I've removed them

I also had to change some enums that were missed in the enum change PR

TaylorJ76 (approved):

PR #2269, opened 2025-04-04
feat(radio-group): adds helper and error text (VIV-2398)

TaylorJ76:


libs/components/src/lib/radio-group/VARIATIONS.md
```diff
@@ -27,6 +27,40 @@ Use the `label` to set the **Radio Group** label.
...
+	label="Pick a number"
+	name="number"
+>
+	<vwc-radio label="Number one" value="1"></vwc-radio>
+	<vwc-radio label="Number two" value="2"></vwc-radio>
+	<vwc-radio label="Number three" value="3"></vwc-radio>
+</vwc-radio-group>
+```
+
+### Error Text
```
RichardHelm, line 48: How would one actually implement a required radio-group in a form? And in that case shouldn't the error be displayed automatically on the radio-group?

SpawnAtis: I have added this part

libs/components/src/lib/radio-group/radio-group.template.ts
```diff
@@ -1,38 +1,44 @@
...
-
 			<div
-				class="positioning-region ${(x) =>
+				${delegateAria({
+					role: 'radiogroup',
+					ariaDisabled: (x) => x.disabled,
+					ariaReadOnly: (x) => x.readOnly,
+					ariaOrientation: (x) => x.orientation,
+				})}
+				aria-labelledby="label"
```
SpawnAtis, line 26: @RichardHelm Should I  move all this aria* props inside of `delegateAria` too ?

SpawnAtis (approved):

SpawnAtis (approved):

PR #2284, opened 2025-04-14
feat(file-picker): add rejectedFiles property (VIV-2556)

SpawnAtis:


TaylorJ76:
Nice work @SpawnAtis!
I noticed that we currently have no examples of how to use the the `files` property.
Could you add an example and also an example of the new `rejectedFiles` property?

TaylorJ76 (changes requested):
Nice work @SpawnAtis!
I noticed that we currently have no examples of how to use the the files property.
Could you add an example and also an example of the new rejectedFiles property?

SpawnAtis:
> Nice work @SpawnAtis! I noticed that we currently have no examples of how to use the the files property. Could you add an example and also an example of the new rejectedFiles property?

sure

libs/components/src/lib/file-picker/README.md
```diff
@@ -161,6 +161,69 @@ Note: localisation will need to be handled at the application level.
 
 When setting `block-size` or `max-block-size` on the file-picker the list of the added files will have auto scroll.
 
+## Files & Rejected files
```
TaylorJ76, line 164:
```suggestion
## Files & Rejected Files
```

TaylorJ76: I know it's picky, but all our headings should be in title case.
Here are the [documentation writing guidelines](https://vivid.deno.dev/resources/documentation/)

TaylorJ76 (changes requested):

TaylorJ76 (approved):

PR #2285, opened 2025-04-14
fix(tabs): scroll shadow shown incorrectly when tab is removed (VIV-2279)

TaylorJ76:


RichardHelm:
I can't reproduce the issue in our docs, when trying the original code in Vue the issue is still there:
image

RichardHelm (changes requested):
See comment

TaylorJ76 (commented):
Thanks Richard. Looks good to me.

PR #2288, opened 2025-04-16
fix(alert): refine the correct spacing between dismiss icon and divider according to the current design (VIV-2561)

SpawnAtis:


libs/components/src/lib/alert/partials/variables.scss
```diff
@@ -1,2 +1,3 @@
 $alert-min-inline-size: --alert-min-inline-size;
 $alert-max-inline-size: --alert-max-inline-size;
+$dismiss-button-divider-spacing: --_alert-dismiss-button-divider-spacing;
```
TaylorJ76, line 3: I don't think this is something we want to expose to the consumer.
I think the fix can just be changing `8px` to `16px` in the `alert.scss`.

TaylorJ76 (changes requested):

libs/components/src/lib/alert/partials/variables.scss
SpawnAtis: Ok, I will change it. As I understood,  this` '--var'` and  `'--_var'`  is used  for   "Public" or "Private" variables separation.  Is it correct ?

TaylorJ76: Ah yes. Sorry, I missed the underscore. But even so, I don't think it needs to be a variable as it is only used in one place.

TaylorJ76 (approved):

PR #2290, opened 2025-04-17
fix(pickers): clear value on invalid input (VIV-2400)

RichardHelm:


konradsienkowski (approved):
Looks good to me 👍 Tested it locally based on unit test cases & example reported on Slack - behaves as expected.

PR #2298, opened 2025-04-23
chore(docs): add article/component name in title tag (VIV-2567)

konradsienkowski:


apps/docs/content/_layouts/component.njk
```diff
@@ -1,4 +1,5 @@
 {% extends './page.njk' %}
+{% set title = component.title %}
```
RichardHelm, line 2: Alternatively you could use `eleventyComputed` in `component-[variations/etc.].njk` to set the title and have a different title for each tab. E.g. "Button - Variations - Vivid" for the variations tab

RichardHelm (dismissed):
Nice! Just one suggestion

apps/docs/content/_layouts/component.njk
konradsienkowski: Thanks! I've been thinking about it, but I wasn't sure if we want to include tab name in the title too. I also thought about using a different separator in such case e.g. "Accordion → Variations – Vivid". @TaylorJ76 @SpawnAtis what do you think?

SpawnAtis: I think, we could use the pattern like `Button Component | Vivid` to follow  "title case" from the [guideline](https://vivid.deno.dev/resources/documentation/)

SpawnAtis (approved):

PR #2302, opened 2025-04-25
fix(menu): error when no menu item children (VIV-2568)

konradsienkowski:


libs/components/src/lib/menu/menu.ts
```diff
@@ -252,36 +250,32 @@ export class Menu extends Anchored(DelegatesAria(VividElement)) {
 		return this.isMenuItemElement(el);
 	};
 
-	private setFocus(focusIndex: number, adjustment: number): void {
+	private setFocus(focusIndex: number): void {
 		if (this.menuItems === undefined) {
 			return;
 		}
 
 		while (focusIndex >= 0 && focusIndex < this.menuItems.length) {
```
RichardHelm, line 258: This shouldn't be a loop anymore, right?
```suggestion
		if (focusIndex >= 0 && focusIndex < this.menuItems.length) {
```

RichardHelm (changes requested):

libs/components/src/lib/menu/menu.ts
konradsienkowski: > This shouldn't be a loop anymore, right?

Yes, you're right. Updated in the next commit

RichardHelm (approved):

PR #2309, opened 2025-04-30
fix(file-picker):  set error validity  with  rejected files (VIV-2434)

SpawnAtis:


libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -659,6 +693,7 @@ describe('form associated vwc-file-picker', function () {
 			formWrapper,
 		});
 		addFiles([await generateFile('london.png', 1)]);
+		expect(new FormData(formElement).get('file')).toBeTruthy();
```
RichardHelm, line 696: What does this do?
```suggestion
```

libs/components/src/lib/file-picker/USE-CASES.md
```diff
@@ -1,28 +1,62 @@
...
 		<div class="buttons">
-			<vwc-button label="Reset" appearance="outlined" type="reset"></vwc-button>
+			<vwc-button label="Reset" type="reset" appearance="outlined"></vwc-button>
 			<vwc-button label="Submit" appearance="filled" type="submit"></vwc-button>
 		</div>
 	</vwc-layout>
 </form>
 
+<script>
+	window.addEventListener('load', function () {
```
RichardHelm, line 25: I think we should keep this example intact.
Maybe you can add it to the rejectedFiles example in README instead

RichardHelm (changes requested):
Nice work! It seems to work as expected, but the tests are failing

libs/components/src/lib/file-picker/file-picker.spec.ts
SpawnAtis: Thx for the notice. I have removed it

libs/components/src/lib/file-picker/USE-CASES.md
SpawnAtis: I moved it into README

libs/components/src/lib/file-picker/ui.test.ts-snapshots/snapshots-file-picker-Desktop-Chromium-linux.png
RichardHelm, line 1: Nice! But it has lost the focus now. Can you add it back please? We want to include it in screenshots

libs/components/src/lib/file-picker/README.md
```diff
@@ -220,6 +246,8 @@ Use `files` or `rejectedFiles` to access the list of files that have been added
 
 		simulateDrop(dummyFile1);
 		simulateDrop(dummyFile2);
+
+		// document.querySelector('vwc-button[type="submit"]').click();
```
RichardHelm, line 250: Let's remove the commented out code
```suggestion
```

libs/components/src/lib/file-picker/file-picker.spec.ts
```diff
@@ -622,6 +635,67 @@ describe('form associated vwc-file-picker', function () {
...
+			componentTagName: COMPONENT_TAG,
+			formWrapper,
+		});
+		element.accept = '.txt';
+		element.maxFileSize = 1;
+
+		const submitHandler = vi.fn();
+		// const fakeEvent = { preventDefault: vi.fn() };
+		formElement.addEventListener('submit', submitHandler);
+		// expect(fakeEvent.preventDefault).toHaveBeenCalledTimes(1);
```
RichardHelm, line 661: Here as well
```suggestion
		formElement.addEventListener('submit', submitHandler);
```

RichardHelm (changes requested):
Thanks! It looks good now I just left some tiny comments

libs/components/src/lib/file-picker/ui.test.ts-snapshots/snapshots-file-picker-Desktop-Chromium-linux.png
SpawnAtis: I added it back

RichardHelm (approved):

PR #2316, opened 2025-05-06
feat(searchable-select): add max-items attribute (VIV-2404)

konradsienkowski:


SpawnAtis:
Could u add a browser test for this prop ?

konradsienkowski:
> Could you add a browser test for this prop ?

That's a good point, thanks! Added it to the ui test

RichardHelm (changes requested):
Thanks, looks good! I've left some comments but they are pretty minor

konradsienkowski:
@RichardHelm thanks for the review! Applied your suggestions in the latest commit

RichardHelm (approved):

RichardHelm (approved):

PR #2319, opened 2025-05-07
chore(rich-text-editor): add the attachments slot with combined scrolling (VIV-2584)

YonatanKra:
* Adds the `attachments` slot
* Expose `getValue` from the facade to get the actual HTML value
* Expose the `scrollToAttachments` method on the element
* Add example to docs


libs/components/src/lib/rich-text-editor/facades/vivid-prose-mirror.facade.spec.ts
```diff
@@ -982,4 +982,36 @@ describe('ProseMirrorFacade', () => {
 			);
 		});
 	});
+
+	describe('getContent', () => {
```
TaylorJ76, line 986: Isn't this testing `getValue`?

TaylorJ76 (changes requested):

PR #2322, opened 2025-05-08
fix(dialog): subtitle misalignment when non-dismissible (VIV-2612)

TaylorJ76:


libs/components/src/lib/dialog/dialog.template.ts
```diff
@@ -71,8 +71,14 @@ export const DialogTemplate = (context: VividElementDefinitionContext) => {
 							<slot name="graphic">
 								${when((x) => x.icon, icon(iconTag))}
 							</slot>
-							${when((x) => x.headline, headline())}
-							${when((x) => x.subtitle, subtitle())}
+							${when(
+								(x) => x.headline || x.subtitle,
+								() =>
```
konradsienkowski, line 76: In other components we're usually using a template string directly in the second argument of `when`, here an arrow function is used to return it

konradsienkowski (dismissed):

konradsienkowski (approved):

PR #2324, opened 2025-05-09
fix(select): announce label in multiple version when using screen reader (VIV-2613)

konradsienkowski:


libs/components/src/lib/select/VARIATIONS.md
```diff
@@ -5,8 +5,7 @@
 The `label` attribute provides a short description of the purpose of the Select.
 
 <vwc-note connotation="information" icon="accessibility-line" headline="Accessibility Tip">
-
-If a visible label can't be used, provide one using the <nobr><code>aria-label</code></nobr> attribute. This ensures screen readers announce the purpose of the element, making it accessible to all users.
+<p>If a visible label can't be used, provide one using the <nobr><code>aria-label</code></nobr> attribute. This ensures screen readers announce the purpose of the element, making it accessible to all users. </p>
```
RichardHelm, line 8: Actually, we should be able to just use markdown if it is separated from html by a new line
```suggestion

If a visible label can't be used, provide one using the `aria-label` attribute. This ensures screen readers announce the purpose of the element, making it accessible to all users.

```

RichardHelm (approved):
Nice!

RichardHelm (approved):

PR #2325, opened 2025-05-09
fix(data-grid-cell): add button role announcement for screen readers (VIV-2606)

jhajduk-vonage:


libs/components/src/lib/data-grid/data-grid-cell.template.ts
```diff
@@ -58,7 +58,10 @@ export const DataGridCellTemplate = (
 			@click="${(x) => x._handleInteraction()}"
 			@keydown="${(x, c) => handleKeyDown(x, c.event as KeyboardEvent)}"
 		>
-			<div class="base">
+			<div
+				class="base"
+				role="${(x) => (shouldShowSortIcons(x) ? 'button' : undefined)}"
```
RichardHelm, line 63: I'm not sure that is what we want, because the the whole header cell becomes a button now
So maybe the role should be just on the sort icon and it should have a label of "Sort"?

RichardHelm (changes requested):

libs/components/src/lib/data-grid/data-grid-cell.template.ts
jhajduk-vonage: I took my inspiration from these two sources:
1. https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/ 
2. https://mulder21c.github.io/aria-practices/examples/table/sortable-table.html

In both of these examples the header cell element got its `aria-sort` attributes, and their children are either native buttons, or roles set to button.

RichardHelm: Mhm, in our data grid you could theoretically have buttons inside the header as well (like filter), which would be inside the button now.
Some data-grids have sort at a separate button like that: https://mui.com/x/react-data-grid/sorting/
But I'm not sure what the best way is, so I'll approve it because at least it's better than before

RichardHelm (approved):

libs/components/src/lib/data-grid/data-grid-cell.template.ts
jakub-hajduk: Awesome, thanks :).

Also - I understand your point. Thanks for the example.

I've tried it first, but it felt kinda unnatural for me, to use the tab button to reach the sorting feature, while using roving tabindex for the rest of the table.

I found this solution to be the cleanest, without adding "use tab key to use sorting button" type labels and descriptions and then dealing with i18n.

PR #2326, opened 2025-05-09
feat(eslint-plugin): add rule to forbid IDREF ARIA attributes (VIV-2390)

SpawnAtis:


libs/eslint-plugin/README.md
RichardHelm, line 1: Mhm, this information is already present in our docs, so I would remove it from here

libs/eslint-plugin/src/rules/no-idref-aria-attribute.md
RichardHelm, line 1: You need to add this file to `eslintRules.json` so that it appears in the docs. Which, by the way, are here in case you weren't aware: https://vivid-fwspw8ez0h97.deno.dev/guides/eslint-plugin/

While your docs are very thorough, I would align it more with the other rules so it doesn't stand out on that page

RichardHelm (changes requested):

RichardHelm (approved):

PR #2331, opened 2025-05-12
chore(dial-pad): fix new docs not being displayed (VIV-000)

RichardHelm:


konradsienkowski (dismissed):
Looks good to me 👍  Btw. we have a "Use cases" article in the `dial-pad` directory, but it seems that its content is copied from the `dialog`: https://github.com/Vonage/vivid-3/blob/main/libs/components/src/lib/dial-pad/USE-CASES.md

konradsienkowski (approved):

PR #2337, opened 2025-05-14
fix(tooltip, toggletip): semantic improvements (VIV-2631)

RichardHelm:


libs/components/src/lib/toggletip/toggletip.template.ts
```diff
@@ -33,10 +33,7 @@ export const ToggletipTemplate = (context: VividElementDefinitionContext) => {
 			exportparts="vvd-theme-alternate"
 		>
 			<div class="content-wrapper">
-				${when(
-					(x) => x.headline,
-					html`<header class="headline">${(x) => x.headline}</header>`
-				)}
+				${when((x) => x.headline, html`<h2 class="headline">${(x) => x.headline}</h2>`)}
 				<slot></slot>
 				<footer class="action-items"><slot name="action-items"></slot></footer>
```
TaylorJ76, line 38: Also need to "Remove any landmark semantics from the tooltips as they are not regions", in this case the `footer` element.

```suggestion
				<div class="action-items"><slot name="action-items"></slot></div>
```

TaylorJ76 (changes requested):

TaylorJ76 (approved):

PR #2339, opened 2025-05-14
chore(menu-item, option): remove aria-checked selectors (VIV-2601)

SpawnAtis:


RichardHelm (approved):

RichardHelm:
@SpawnAtis Thanks! I just updated the title as I don't think we want this to appear in the changelog

konradsienkowski (approved):

PR #2342, opened 2025-05-16
feat(dial-pad): adds focus method and autofocus property (VIV-2635)  

SpawnAtis:


libs/components/src/lib/dial-pad/definition.ts
```diff
@@ -17,6 +17,9 @@ export const dialPadDefinition = defineVividComponent(
 	[buttonDefinition, textFieldDefinition, iconDefinition],
 	{
 		styles,
+		shadowOptions: {
+			delegatesFocus: true,
+		},
```
RichardHelm, line 22: So this is makes a pretty big difference, because it will focus the first focusable element when clicking anywhere in the component. I don't think that this component should delegateFocus because it makes little sense that clicking on the bottom of the component will focus the text field

libs/components/src/lib/dial-pad/dial-pad.spec.ts
```diff
@@ -9,22 +10,41 @@ const COMPONENT_TAG = 'vwc-dial-pad';
 describe('vwc-dial-pad', () => {
 	let element: DialPad;
 
-	function getTextField() {
-		return getBaseElement(element).querySelector('.phone-field') as TextField;
+	function getTextField(component: HTMLElement) {
+		return getBaseElement(component).querySelector('.phone-field') as TextField;
```
RichardHelm, line 14: Sorry, but is there any reason for all these unrelated changes in the file?

libs/components/src/shared/foundation/button/button.ts
```diff
@@ -161,6 +161,12 @@ export class VividFoundationButton extends DelegatesAria(FormAssociatedButton) {
 		this.proxy.setAttribute('type', this.type);
 		this.handleUnsupportedDelegatesFocus();
 
+		if (this.autofocus) {
+			DOM.queueUpdate(() => {
+				this.focus();
+			});
+		}
+
```
RichardHelm, line 169: I don't think this is needed. It's also not correct because this will focus any button with autofocus when it is added to the page, while autofocus only does it on page load.
```suggestion
```

RichardHelm (changes requested):

libs/components/src/lib/dial-pad/dial-pad.spec.ts
SpawnAtis: I have added some helpers like `getInput` and `getActiveElement`, so I refactored the rest of them in the same way

libs/components/src/shared/foundation/button/button.ts
SpawnAtis: Actually, its works without it, but ,as I showed it on one of the daily standup, I put it here to be 100% sure that it will work in every browser. I have checked `text-field` and there is the same behavior - [Click](https://github.com/Vonage/vivid-3/blob/340905c0b3eee1706268d01f0c34f05c0733d766/libs/components/src/lib/text-field/text-field.ts#L417). So, what do u think ?

libs/components/src/lib/dial-pad/ui.test.ts
```diff
@@ -13,7 +13,7 @@ const components = ['dial-pad'];
 test('should show the component', async ({ page }: { page: Page }) => {
 	const template = `
 		<div style="margin: 5px;">
-			<vwc-dial-pad></vwc-dial-pad>
+			<vwc-dial-pad autofocus></vwc-dial-pad>
```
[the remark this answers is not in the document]
SpawnAtis: Yeah, I have added a separate test with screenshots for it.

libs/components/src/shared/foundation/button/button.ts
RichardHelm: Alright, let's keep it then

RichardHelm (approved):

PR #2343, opened 2025-05-16
fix(divider): hide divider from screen readers (VIV-2624)

jhajduk-vonage:


jhajduk-vonage:
In this PR I have added aria-hidden atrribute to the host element and it's not read by the SRs. 
However IMO, since divider is not seen by screen readers, it doesn't make sense to setup any of the aria properties (like role and orientation), and they should be removed.
Since removing any properties are breaking changes, I first marked them as `@deprecated`. There should be additional task created for removing these in the future.

libs/components/src/lib/divider/divider.template.ts
```diff
@@ -13,6 +13,7 @@ export const DividerTemplate = html<Divider>` <span
 	class="${getClasses}"
 	${delegateAria({
 		role: (x) => x.role || DividerRole.separator,
+		ariaHidden: true,
```
RichardHelm, line 16: You can just set the default role to presentation instead.

But that only changes the API and doesn't fix any accessibility issues because there aren't any accessibility issues to fix, so I don't think it should be done either.

RichardHelm (changes requested):

libs/components/src/lib/divider/divider.template.ts
jhajduk-vonage: Changing default role to `presentation` also breaks styling in `action-group`, since:
```
/* Separator */
::slotted([role='separator']) {
```

RichardHelm: I see. Well, I know that's what the ticket says but adding aria-hidden to all dividers is wrong because there are non-decorative dividers that should not be hidden.
Again, this component doesn't have any accessibility issues and I don't think anything should be done

jhajduk-vonage: Okay, got you now 😅.

I'll follow your suggestion - I'll make the `presentation` role as the default one, remove the deprecation tags and deprecation info from the docs.

Thanks for the input!

jhajduk-vonage: Opinion from applause is that all the [separators should remain hidden](https://vonage-external.slack.com/archives/C08L4GXK3EF/p1747404301216219). 
I updated the [jira ticket with the comment](https://jira.vonage.com/browse/VIV-2624?focusedId=7893029&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-7893029)

TaylorJ76 (approved):

PR #2350, opened 2025-05-20
feat(components): icon size adjustment (VIV-2633)

TaylorJ76:
See the story for the details of the adjustments.
As TextField and Feedback message icons were adjusted, it affects a whole bunch of components. Hence the number of visual snapshots that needed updating.


libs/components/src/shared/patterns/form-elements/message.scss
konradsienkowski, line 1: It's not being applied in the feedback messages, which are styled in `libs/components/src/shared/feedback/feedback-message.scss`. If we want to adjust the icon size there too, we need to update this file

libs/components/src/lib/menu/ui.test.ts-snapshots/snapshots-absolute-menu-Desktop-Firefox-linux.png
konradsienkowski, line 1: Looks like a bug, not sure if it's intended

konradsienkowski (commented):
Looks good, I've left a few comments

libs/components/src/lib/card/card.scss
```diff
@@ -133,7 +134,7 @@
 }
 
 .icon {
-	font-size: 20px;
+	font-size: #{icon-size-variables.$icon-size-normal};
```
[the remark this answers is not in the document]
TaylorJ76: Yes, Ayala confirmed only dialog icon is to be 28px. Card icon is to stay as is.

TaylorJ76:
More icon size changes to come later. Will revisit then,

PR #2355, opened 2025-05-21
chore(docs): improve example names (VIV-000)

RichardHelm:
This changes the file names for code examples from something like `/frames/cbd-code-block-<random number>.html` to `/frames/components-accordion-1.html` to make them stable between eleventy runs


TaylorJ76 (approved):

TaylorJ76 (commented):
Nice. Just need to sort linting issue.

PR #2357, opened 2025-05-22
chore(components): deprecate icon prop (VIV-2643)

SpawnAtis:


libs/components/src/shared/patterns/affix.ts
```diff
@@ -13,7 +13,7 @@ export const AffixIcon = <T extends Constructor<VividElement>>(Base: T) => {
 	class AffixIconElement extends Base {
```
jhajduk-vonage, line 13: Correct me if I'm wrong, but This whole mixin will also become obsolete. It'd make sense to also mark it as `@deprecated`. This way it will be easier to track when it'll have to be removed.

libs/components/src/lib/fab/README.md
```diff
@@ -45,7 +47,7 @@ registerFab('your-prefix');
 
 ### Icon Slot
 
-Use the `icon` slot to customise icons. If set, the icon attribute is ignored.
+Use the `icon` slot to customise icons. If set, the icon _(deprecated)_ attribute is ignored.
 
 ```html preview
 <vwc-fab icon="thumbs-up-line" size="normal" label="ask-vivid">
```
jhajduk-vonage, line 53: This `icon` attribute should probably be removed.

libs/components/src/lib/nav-item/README.md
jhajduk-vonage, line 1: Usages of `icon` attribute in this file should be removed. And if not, then It'd be good to add informaiton in the docs, that it is used here for a specific reason.

libs/components/src/lib/tabs/README.md
```diff
@@ -131,9 +132,15 @@ When using Tabs inside flex structure, setting `--tabs-block-size: 100%` with `f
...
+		</vwc-tab>
+		<vwc-tab label="Playlist">
+			<vwc-icon slot="icon" name="playlist-line"></vwc-icon>
+		</vwc-tab>
+		<vwc-tab label="Favourites">
+			<vwc-icon slot="icon" name="star-line"></vwc-icon>
+		</vwc-tab>
 		<vwc-tab-panel>
 			<vwc-empty-state
 				icon="error-solid"
```
jhajduk-vonage, line 146: I suppose the `vwc-empty-state` component should preserve this property, since it's not in the changed files, right?

jhajduk-vonage (changes requested):

TaylorJ76 (changes requested):

libs/components/src/shared/patterns/affix.ts
SpawnAtis: It is used to add slotted behavior for icon prop too. So , I do not think, that we will have to remove it in the near future.

libs/components/src/lib/alert/alert.ts
```diff
@@ -96,7 +96,7 @@ export class Alert extends AffixIcon(Localized(VividElement)) {
 
 	/**
 	 * Sets an appropriate icon / icon color for the connotation.
-	 *
+	 * @deprecated Use the icon slot instead. This attribute will be removed in a future version.
```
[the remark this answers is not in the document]
SpawnAtis: Thx for the notice

jhajduk-vonage (approved):

TaylorJ76 (approved):

PR #2366, opened 2025-05-26
fix(data-grid-cell): add visually-hidden information when cell is selected (VIV-2623)

jhajduk-vonage:


jhajduk-vonage:
yes. I'll add it.

PR #2371, opened 2025-05-28
chore(components): refactor Anchor mixin to Linkable using new pattern (VIV-2503)

konradsienkowski:


RichardHelm (dismissed):
Nice work! While it should be fine as is, I left some suggestions

PR #2373, opened 2025-05-29
fix(tab): Replace aria-selected attribute with active (VIV-2597)

konradsienkowski:


konradsienkowski:
> Could ,u, pls add property desc into docs ? Check `libs/components/src/lib/tabs/README.md`. And should we add a browser test for it ?

Hi @SpawnAtis, thanks for the review! I renamed `selected` attribute to `active` to make it consistent with `activeId` after talking about it with Richard. It's now also a hidden, internal use-only observable. This doesn't affect UI or styles so I don't think we have to update browser tests here

libs/components/src/lib/tabs/tabs.ts
```diff
@@ -239,8 +240,9 @@ export class Tabs extends VividElement {
 			const panel = this.tabpanels[index];
 
 			const isActiveTab = tab.id === this.activeid;
-
-			tab.setAttribute('aria-selected', isActiveTab ? 'true' : 'false');
+			if (tab instanceof Tab) {
+				isActiveTab ? (tab.active = true) : (tab.active = false);
```
RichardHelm, line 244:
```suggestion
				tab.active = isActiveTab;
```

SpawnAtis, line 244: May be just `tab.active =  isActiveTab` ?

RichardHelm (approved):
Thanks, looks good!

SpawnAtis (dismissed):

PR #2375, opened 2025-05-29
chore(components): refactor FormAssociated and CheckableFormAssociated mixins (VIV-2502)

RichardHelm:


libs/components/src/shared/foundation/form-associated/form-associated.ts
```diff
@@ -327,15 +201,10 @@ export function FormAssociated<T extends ConstructableFormAssociated>(
 		 * HTML Attribute: name
 		 */
 		//@ts-expect-error: Type is incorrectly non-optional
-		name: string;
+		@attr name: string;
```
konradsienkowski, line 204: Maybe we could remove `//@ts-expect-error: Type is incorrectly non-optional` and use `@attr name!: string;`?

libs/components/src/lib/radio/radio.ts
```diff
@@ -197,6 +196,76 @@ export class Radio extends FormAssociatedRadio {
...
+		const siblings = this.parentElement?.querySelectorAll(
+			`${this.tagName.toLocaleLowerCase()}[name="${this.name}"]`
+		);
+		if (siblings) {
+			return Array.from(siblings) as unknown as Radio[];
+		}
+		return [];
+	}
+
+	get #radioGroup(): RadioGroup | null {
```
konradsienkowski, line 215: These getters seem to be a good use case for the helpers introduced by @jhajduk-vonage here: https://github.com/Vonage/vivid-3/pull/2372/

konradsienkowski (approved):
Great work!

libs/components/src/shared/foundation/form-associated/form-associated.ts
RichardHelm: It would get rid of the error but the type is still wrong, because it can be undefined. In native inputs it defaults to empty string instead, so that would make more sense but might break some code

PR #2377, opened 2025-05-29
chore(components): refactor form elements mixins (VIV-2501)

RichardHelm:


libs/components/src/shared/patterns/form-elements/form-elements.ts
```diff
@@ -86,14 +95,50 @@ export function formElements<
...
-export function errorText<
-	T extends { new (...args: any[]): Record<string, any> }
->(constructor: T) {
-	class ErrorText extends constructor {
+export type ElementWithSuccessText = MixinType<typeof WithSuccessText>;
+
+/**
+ * Mixin for elements that can display error text.
+ */
+export const WithErrorText = <T extends Constructor<FormElementElement>>(
```
konradsienkowski, line 127: While keeping all three mixins in one file matches the existing structure, I'd consider splitting them into separate files, e.g. with-error-text.ts, with-success-text.ts, form-element.ts and adding index.ts to ease importing into the components

konradsienkowski (approved):
Great work 👍 Left one suggestion about the file structure to consider

PR #2379, opened 2025-05-30
feat(alert,banner): Add accessible labels to connotation icons (VIV-2654)

SpawnAtis:


libs/components/src/lib/alert/alert.ts
```diff
@@ -5,22 +5,16 @@ import { AffixIcon } from '../../shared/patterns/affix';
...
-
-const connotationIconMap = new Map([
-	[Connotation.Accent, 'megaphone-line'],
-	[Connotation.Information, 'info-line'],
-	[Connotation.Success, 'check-circle-line'],
-	[Connotation.Warning, 'warning-line'],
-	[Connotation.Announcement, 'sparkles-line'],
-	[Connotation.Alert, 'error-line'],
-]);
+export const CONNOTATION_ICON_MAP = {
```
konradsienkowski, line 8: Why was it changed to the upper snake case? I haven't seen that convention used in our code before

libs/components/src/lib/alert/alert.ts
```diff
@@ -5,22 +5,16 @@ import { AffixIcon } from '../../shared/patterns/affix';
...
+	[Connotation.Accent]: 'megaphone-line',
+	[Connotation.Information]: 'info-line',
+	[Connotation.Success]: 'check-circle-line',
+	[Connotation.Warning]: 'warning-line',
+	[Connotation.Announcement]: 'sparkles-line',
+	[Connotation.Alert]: 'error-line',
+} as const;
+
+export type AlertConnotation = keyof typeof CONNOTATION_ICON_MAP;
```
konradsienkowski, line 17: This seems to be the reason of failing build process:
```
\[error] ../vue-wrappers/src/generated/components/VAlert.types.ts: SyntaxError: Type expected. (100:23)
[error]    98 |  * Sets an appropriate icon / icon color for the connotation.
[error]    99 |  */
[error] > 100 |         connotation?: ,
```
Unless there is a specific reason why it connotationIconMap was changed from the Map to const, I'd leave it as it was before

konradsienkowski (commented):
Nice work, I've left two comments

libs/components/src/lib/alert/alert.ts
SpawnAtis: As it a constant, I changed it to the the upper snake case, and removed duplication in *.spec.ts both in `alert` and `banner`.

libs/components/src/lib/alert/alert.ts
SpawnAtis: The problem is `  keyof typeof CONNOTATION_ICON_MAP`.  `vue-wrapper` generator does not handle such syntax for now. So, i reverted it to the type union.

konradsienkowski (approved):

PR #2380, opened 2025-05-30
fix(nav-item, nav-disclosure): Replace 'aria-current' with synced 'current' attribute (VIV-2595)

konradsienkowski:


libs/components/src/lib/nav-disclosure/nav-disclosure.ts
```diff
@@ -70,6 +71,20 @@ export class NavDisclosure extends AffixIcon(DelegatesAria(VividElement)) {
 	 */
 	@attr({ mode: 'boolean' }) open = false;
 
+	/**
+	 * Sets the nav-disclosure as a current
```
RichardHelm, line 75: This does not seem very clear. On second thought, the attribute should perhaps be named `active` or `selected` instead, as `current` on its own is not very descriptive.

libs/components/src/lib/nav-item/nav-item.ts
```diff
@@ -61,4 +61,18 @@ export class NavItem extends AffixIcon(Linkable(VividElement)) {
...
+	 * @public
+	 * @remarks
+	 * HTML Attribute: current
+	 */
+	@replaces<boolean>({
+		deprecatedPropertyName: 'ariaCurrent',
+		fromDeprecated: (v) => v,
+	})
+	@attr({ attribute: 'current', mode: 'boolean' })
+	current = false;
```
RichardHelm, line 77: The same applies to nav item

RichardHelm (changes requested):
Looks great! I left a few suggestions

libs/components/src/lib/nav-disclosure/nav-disclosure.template.ts
```diff
@@ -9,15 +9,17 @@ import type { VividElementDefinitionContext } from '../../shared/design-system/d
...
+	connotation,
+	current,
+	open,
+}: NavDisclosure) =>
 	classNames(
 		'control',
 		[`appearance-${appearance}`, Boolean(appearance)],
-		[`connotation-${connotation}`, Boolean(connotation)]
+		[`connotation-${connotation}`, Boolean(connotation)],
+		['current', Boolean(current) && !open]
```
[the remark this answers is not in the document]
konradsienkowski: Isn't it the same line? 🤔

libs/components/src/lib/nav-disclosure/nav-disclosure.ts
RichardHelm: Ok, let's stick with it then 👍

RichardHelm (approved):

PR #2381, opened 2025-05-30
feat(menu-item): replace role with control-type attribute (VIV-2602)

RichardHelm:
- The name `control-type` is taken from `selectable-box`
- Role `presentation` also affects behaviour. To avoid the need to specify it, menu-item now automatically becomes presentational when not a direct child of menu


libs/components/src/lib/menu-item/README.md
```diff
@@ -187,23 +187,53 @@ This can be changed by setting a css-variable to a fixed number of lines or `aut
 </vwc-menu>
 ```
 
+## Presentational
+
+If the Menu Item is not a direct descendant of a Menu, it will automatically become presentational. That means it used only for visual purposes and no longer semantically or functionally act as a Menu Item.
+
+You must wrap it in an element with `role="menuitem"` to provide the menu item functionality.
+
```
RichardHelm, line 195:
```suggestion
You must wrap it in an element with `role="menuitem"` to provide the menu item functionality.

<vwc-note connotation="information">
	<vwc-icon slot="icon" name="info-line" label="Note:"></vwc-icon>

Previously you needed to specify `role="presentation"` in this case. This is no longer necessary as the component will handle it automatically.

</vwc-note>
```

libs/components/src/lib/menu/menu.a11y.spec.ts
```diff
@@ -17,8 +18,10 @@ describe('a11y: vwc-menu', () => {
 		element.open = true;
 		element.ariaLabel = 'A11y label';
 		element.innerHTML = `
-      <div role="menuitem" id="id1">Menu Item 1</div>
-      <div role="menuitem" id="id2">Menu Item 2</div>
+      <vwc-menu-item text="Menu Item 1"></vwc-menu-item>
+      <vwc-menu-item control-type="checkbox" text="Menu Item Checkbox"></vwc-menu-item>
+      <vwc-didiver></vwc-didiver>
```
konradsienkowski, line 23: There seems to be a typo:
```suggestion
      <vwc-divider></vwc-divider>
```

libs/components/src/lib/menu-item/VARIATIONS.md
```diff
@@ -49,47 +49,36 @@ The `icon` prop is deprecated (as of 05/25) and directly replaced with `icon` sl
...
 </vwc-menu>
 ```
 
+<vwc-note connotation="information">
+	<vwc-icon slot="icon" name="info-line" label="Note:"></vwc-icon>
+
+Previously, the `role` attribute was used to set the control type. This is now deprecated (as of 05/25) and replaced with the `control-type` attribute. Using `role` is still functional, but will be removed in a future major release.
+
+</vwc-note>
+
```
konradsienkowski, line 78: In other components docs we used "warning" connotation for prop deprecation note
```suggestion
<vwc-note connotation="warning" headline="Deprecated Prop: role">
	<vwc-icon slot="icon" name="warning-line" label="Warning:"></vwc-icon>

Previously, the `role` attribute was used to set the control type. This is now deprecated (as of 05/25) and replaced with the `control-type` attribute. Using `role` is still functional, but will be removed in a future major release.

</vwc-note>

```

apps/docs/content/_data/components.json
```diff
@@ -292,7 +292,6 @@
 		"hideGuidelines": "false",
```
konradsienkowski, line 292: As `libs/components/src/lib/menu-item/GUIDELINES.md` is empty now, it's worth setting hideGuidelines to true:
```suggestion
		"hideGuidelines": "true",
```

konradsienkowski (changes requested):
Nice work 👍 I've left a few comments

PR #2383, opened 2025-06-02
chore: add DOM traversal helper functions (VIV-2681)

jhajduk-vonage:


jhajduk-vonage:
Still unit tests needs to be added. Will back to that as I'll sort out current tasks.

RichardHelm (approved):
Nice!

PR #2393, opened 2025-06-03
fix(components): Use data-vvd-component instead of role attribute (VIV-2604)

konradsienkowski:


RichardHelm:
Not a review, but did you check the css as well? E.g. action-group uses role to style dividers: `::slotted([role='separator']) {`

jhajduk-vonage (dismissed):

konradsienkowski:
> Not a review, but did you check the css as well? E.g. action-group uses role to style dividers: `::slotted([role='separator']) {`

Good spot, thanks! I have added an extra attribute: `vvd-role` for the components that we need to refer to in CSS.

jhajduk-vonage:
Maybe it could be done globally?
For example in `VivdElement` base class we could add the controller/behaviour that on element connected adds a data-attribute with component name (`[data-vvd-component="button"]`, or `[data-vvd-button]`) ?
This way we would have reliable and consistent way of referring to the elements in CSS.

jhajduk-vonage (approved):

libs/components/src/lib/combobox/combobox.ts
```diff
@@ -288,8 +288,9 @@ export class Combobox extends WithFeedback(
 		}
 
 		if (this.open) {
+			const optionTag = ListboxOption.registeredTagName;
 			const capturedOption = (e.target as HTMLElement).closest(
-				`option,[role=option]`
+				`option, ${optionTag}`
```
RichardHelm, line 293: For good measure I would keep `[role=option]` in there as well, as the component wants to work with non-vivid option children.
The same goes for the other selectors
```suggestion
				`option,[role=option],[data-vvd-component=option]`
```

RichardHelm (changes requested):

RichardHelm (approved):

jhajduk-vonage (approved):

PR #2411, opened 2025-06-12
fix(date-picker, date-range-picker, date-time-picker): fix accessibility issues (VIV-2691)

RichardHelm:


konradsienkowski (approved):
Nice work 👍 The only thing that we could improve is usage of ordinal numbers (e.g. 15th instead of 15) for the day announcements

PR #2414, opened 2025-06-13
feat(dialog): add scrollable-body support to dialog (VIV-2638)

konradsienkowski:


libs/components/src/lib/dialog/README.md
```diff
@@ -252,6 +252,103 @@ Often used to indicate which button the user pressed to close it.
 </script>
 ```
 
+## Scrollable Body
```
TaylorJ76, line 255: I consider this a visual variation, so I think it belongs on the Variations page

konradsienkowski: @TaylorJ76 I wasn't sure about it, thanks - updated in the following commit

TaylorJ76 (approved):

PR #2427, opened 2025-06-22
fix(date-time-picker): fix accessibility issues (VIV-2692)

SpawnAtis:


konradsienkowski (changes requested):
Header has now the outline focus style that is inconsistent with other elements. We should change it to the purple one using focus mixin:
Screenshot 2025-06-23 at 09 11 14

Besides that, it looks good 👍

konradsienkowski (approved):

konradsienkowski (approved):

konradsienkowski (approved):

PR #2429, opened 2025-06-24
fix(tag): improve keyboard focus when removable (VIV-2720)

jhajduk-vonage:


libs/components/src/lib/tag/tag.locale.ts
```diff
@@ -0,0 +1,3 @@
+export interface TagLocale {
+	remove: string;
+}
```
konradsienkowski, line 3: Since we're no longer focusing the whole Tag, my suggestion would be to use function here to return localized label and announce it as "Remove Removable tag 1" etc. We're using this approach in a few other localized components, like `SearchableSelect`
```suggestion
export interface TagLocale {
	remove: /* istanbul ignore next */ (label: string) => string;
}
```

libs/components/src/lib/tag-group/ui.test.ts-snapshots/snapshots-tag-group-Desktop-Chromium-linux.png
konradsienkowski, line 1: It looks like the selectable tag background color have changed, so I just want to confirm if that was intentional

libs/components/src/lib/tag/tag.locale.ts
konradsienkowski: Btw. In other locale files, we just use 'locale.ts' filename

libs/components/src/lib/tag-group/ui.test.ts-snapshots/snapshots-tag-group-Desktop-Chromium-linux.png
jhajduk-vonage: Nope, that was totally unintentional.

libs/components/src/lib/tag/tag.locale.ts
jhajduk-vonage: Okay, thanks!

libs/components/src/lib/tag-group/ui.test.ts-snapshots/snapshots-tag-group-Desktop-Chromium-linux.png
jhajduk-vonage: Turns out that the tests might not picked previous changes until today, since they were not meeting the change threshold.

libs/components/src/lib/tag/tag.template.ts
```diff
@@ -32,8 +32,8 @@ const getClasses = ({
 function renderDismissButton(iconTag: string) {
 	return html<Tag>`
 	<span
-		aria-hidden="true"
 		class="dismiss-button"
+		aria-label="${(x) => `${x.label} - ${x.locale.tag.remove}`}"
```
konradsienkowski, line 36: Since you changed the tag.remove to function accepting label as argument, this part should be changed to:
```suggestion
		aria-label="${(x) => x.locale.tag.remove(x.label) }"
```

konradsienkowski (changes requested):

PR #2443, opened 2025-07-08
fix(combobox): move focus indicator to the options when using keyboard navigation (VIV-2731)

konradsienkowski:


konradsienkowski:
Closing that as the issue was resolved in https://github.com/Vonage/vivid-3/pull/2447 without modifying focus behavior

PR #2447, opened 2025-07-09
fix(combobox): Fix contrast accessibility issue (VIV-2731)

konradsienkowski:


libs/components/src/lib/option/option.template.ts
```diff
@@ -33,6 +33,7 @@ export const ListboxOptionTemplate = (
 				ariaSelected: (x) => x.selected,
 				ariaDisabled: (x) => x.disabled,
 			})}
+			?selected="${(x) => x.selected}"
```
RichardHelm, line 36: Option already has a selected attribute (selectedAttribute, which is the default selected state) so this will clash with it.
Maybe make `.selected` an `@attr` with name `current-selected` to keep it aligned with other components (e.g. `current-value`)

RichardHelm (changes requested):

libs/components/src/lib/option/option.ts
```diff
@@ -58,16 +58,18 @@ export class ListboxOption extends HostSemantics(
...
 	protected defaultSelectedChanged(): void {
 		if (!this.dirtySelected) {
 			this.selected = this.defaultSelected;
-
-			if (this.proxy instanceof HTMLOptionElement) {
-				this.proxy.selected = this.defaultSelected;
-			}
+		}
+		if (this.proxy instanceof HTMLOptionElement) {
+			this.proxy.selected = this.defaultSelected;
```
RichardHelm, line 71: But now it sets `this.proxy.selected` even when `dirtySelected`
So I think it needs:
```suggestion
			if(!this.dirtySelected) {
				this.proxy.selected = this.defaultSelected;
			}
```

RichardHelm (changes requested):

PR #2456, opened 2025-07-16
fix(tooltip): tooltip anchor a11y (VIV-2748)

TaylorJ76:
As per a11y test feedback:

~~Removes focus behavior.~~
~~Tooltip can now be toggled using a click (to accommodate keyboard users).~~
Improves a11y semantics on the Anchor element. (aria-haspopup, aria-expanded).

I agreed with the feedback about the removal of focus. We will talk it through with applause.

### DIFF
--- a/apps/docs/content/whats-new/blog/rethinking-design-token-architecture.md
+++ b/apps/docs/content/whats-new/blog/rethinking-design-token-architecture.md
@@ -93,3 +93,5 @@ We evaluated the current state of token usage in Vivid:
 ## Final Thoughts
 
 Design tokens are foundational—but they shouldn’t be overwhelming. The more levels we introduce, the more we need to maintain, document, and teach. When in doubt, opt for simplicity.
+
+Watch this space for our upcoming proposal document for the work we have planned in this area.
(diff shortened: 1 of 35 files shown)


RichardHelm:
I don't understand this change. Without the focus trigger, the tooltip is no longer keyboard accessible.

konradsienkowski:
1. Mouse behavior seems a bit weird, as hovering over the anchor element makes tooltip open, but clicking on the anchor element while hovering hides it/toggles its visibility. Maybe instead of using @click event it'd be better to implement the new behavior only for the keyboard navigation?
2. When focusing and using enter on one tooltip, and repeating the same action on the next, we end up with multiple tooltips stacked:
Screenshot 2025-07-16 at 12 03 18
3. The information that the anchor element must be focusable remains in the docs: https://vivid-pw8atgwq5fa3.deno.dev/components/tooltip/accessibility/#anchor-element-accessbility

RichardHelm:
But clicking the button already triggers it's regular action. So the tooltip is useless if it is only displayed at that point.

konradsienkowski (approved):

PR #2468, opened 2025-07-25
feat(tag): add support for subtle-light appearance (VIV-2741) 

SpawnAtis:


konradsienkowski (dismissed):
Looks good to me, left one comment about `USE-CASES.md`.
Btw. the branch naming you're using seems to be incompatible with Deno deployments. Why don't you use the current convention (`VIV-XXXX-component-name-branch-name`)?

konradsienkowski (approved):

PR #2474, opened 2025-07-28
chore(selectable-box)!: remove clickable prop (VIV-2675)

konradsienkowski:


RichardHelm (approved):
Looks good! You should add a `!` in the title as it is a breaking change

PR #2477, opened 2025-07-29
fix(time-picker): fix time picker focus on button click (VIV-2759)

konradsienkowski:


libs/components/src/lib/time-picker/time-picker.ts
```diff
@@ -168,11 +168,12 @@ export class TimePicker extends TimeSelectionPicker(
 	/**
 	 * @internal
 	 */
-	override _onPickerButtonClick() {
-		super._onPickerButtonClick();
-		if (this._popupOpen) {
+	override _onPickerButtonClick(event: Event) {
+		super._onPickerButtonClick(event);
+
+		if (this._popupOpen && event instanceof MouseEvent && event.detail === 0) {
```
RichardHelm, line 174: Mhm. Focus still needs to move into the dialog, it just shouldn't be visible unless interacted with a keyboard.
So if you click the picker button, the focus should not be visible. But afterwards, pressing any button (even something like A) should reveal the focus inside the dialog.
The decision when focus needs to be visible should ideally be up to the browser, and it already correctly determines that it should not be (element does not match `:focus-visible` selector), but it is shown anyway.
So maybe this can just be a CSS fix?

RichardHelm (changes requested):

RichardHelm (approved):

PR #2491, opened 2025-08-08
chore(layout): split UI tests into the smaller parts (VIV-2776)

konradsienkowski:


libs/components/src/lib/layout/ui.test.ts
```diff
@@ -7,244 +7,336 @@ import {
...
+	const template = `
+		<div style="margin: 5px;">
+			<vwc-layout>
+				<vwc-card
+					headline="Lorem ipsum"
+					text="Lorem ipsum dolor sit amet, consectetur adipiscing elit."
+				>
+					<img
+						slot="media"
+						src="https://picsum.photos/id/1015/300/200"
```
RichardHelm, line 22: Maybe we should host the images locally instead. But actually they're not relevant to testing the layout component, so I think we can just replace them with a div with background color as a placeholder

libs/components/src/lib/layout/ui.test.ts
```diff
@@ -7,244 +7,336 @@ import {
...
+
+	const testWrapper = await page.$('#wrapper');
+
+	await page.waitForLoadState('networkidle');
+
+	expect(await testWrapper?.screenshot()).toMatchSnapshot(
+		'snapshots/layout.png',
+		{
+			maxDiffPixelRatio: 0.02,
+		}
```
RichardHelm, line 107: Don't know why we have this in the first place. Something like 0.02 is the default anyway (but it would be better if it were 0)
```suggestion
```

RichardHelm (changes requested):
Thanks! This looks good, I added some comments that we could address if we're already touching it

RichardHelm (approved):

PR #2505, opened 2025-08-21
feat(toggletip): adds localised aria-label to anchor element (VIV-2787)

TaylorJ76:


konradsienkowski (commented):
Looks good, I've put comments with an idea for making these labels a bit more human-readable

konradsienkowski (approved):

konradsienkowski (approved):

PR #2507, opened 2025-08-21
feat(components): add contextual-help component and mixin (VIV-2755)

konradsienkowski:


jhajduk-vonage (dismissed):

RichardHelm:
As discussed in the standup, I would prefer to have a single slot that consumers use to slot in the contextual help directly.
While both approaches will work fine, I think we should avoid components "forwarding" the API of a child component as it makes them more complicated and coupled than needed. (Although in this case it is only 2 slots instead of 1.)

Also the contextual help makes sense as a standalone component, as this pattern is already used a lot throughout Vonage. (But very inconsistently and often not accessible with a tooltip on a icon.)

We can add a note that only contextual help is allowed in the slot or even enforce it, e.g. with CSS that sets `display: none` on any other slotted content.

libs/components/src/shared/patterns/form-elements/with-contextual-help.ts
```diff
@@ -0,0 +1,32 @@
...
+) => {
+	/**
+	 * @public
+	 * @slot contextual-help - Slot for the contextual-help component, displayed next to the label.
+	 */
+	class ElementWithContextualHelp extends Base {
+		/**
+		 * @internal
+		 */
+		@observable contextualHelpSlottedContent?: HTMLElement[];
```
RichardHelm, line 19:
```suggestion
		@observable _contextualHelpSlottedContent?: HTMLElement[];
```

libs/components/src/shared/patterns/form-elements/with-contextual-help.ts
```diff
@@ -0,0 +1,32 @@
...
+
+		/**
+		 * @internal
+		 */
+		get hasContextualHelp(): boolean {
```
RichardHelm, line 24:
```suggestion
		get _hasContextualHelp(): boolean {
```

PR #2520, opened 2025-09-04
feat(simple-color-picker): add new component (VIV-2790)

konradsienkowski:


libs/components/src/lib/simple-color-picker/README.md
```diff
@@ -0,0 +1,261 @@
...
+registerSimpleColorPicker('your-prefix');
+```
+
+```html preview
+<script type="module">
+	import { registerSimpleColorPicker } from '@vonage/vivid';
+	registerSimpleColorPicker('your-prefix');
+</script>
+
+<your-prefix-simple-color-picker></your-prefix-simple-color-picker>
```
TaylorJ76, line 25: Could you show an example with a button and some swatches?

TaylorJ76 (changes requested):

PR #2585, opened 2025-10-31
feat(dial-pad): add long press on '0' button to render '+' (VIV-2876)

SpawnAtis:


libs/components/src/lib/dial-pad/dial-pad.template.ts
```diff
@@ -175,6 +180,11 @@ function renderDigits(
 							c.parent.autofocus && c.parent.noInput && c.index === 0}"
 					  aria-label="${(x, c) => c.parent.locale.dialPad[x.ariaLabel]}"
 					  ?disabled="${(_, c) => c.parent.disabled}"
+					  @pointerdown="${(x, c) =>
+							c.parent._startLongPress(x.value, c.event as PointerEvent)}"
+					  @pointerup="${(_, c) => c.parent._endLongPress()}"
+					  @pointercancel="${(_, c) => c.parent._cancelLongPress()}"
```
RichardHelm: I think pointercancel will always fire pointerleave as well
```suggestion
```

libs/components/src/lib/dial-pad/dial-pad.template.ts
```diff
@@ -175,6 +180,11 @@ function renderDigits(
...
+					  @pointerleave="${(_, c) => c.parent._cancelLongPress()}"
 					  @click="${onDigitClick}">
```
RichardHelm: What about long-pressing with the keyboard? It should probably long-press with space as well
Maybe it would be easier to track the time when the pointer or keyboard press started and determine in the click handler if it was a long press. Just note that pointer/keyup is fired before click

RichardHelm (changes requested):

libs/components/src/lib/dial-pad/dial-pad.template.ts
SpawnAtis: Hi. `pointercancel` can fire without `pointerleave` (e.g., pointer canceled while still over the element)

libs/components/src/lib/dial-pad/dial-pad.ts
```diff
@@ -147,6 +147,57 @@ export class DialPad extends Localized(VividElement) {
...
+	 * @internal
+	 */
+	_startLongPress(digit: string, event: PointerEvent) {
+		if (this.disabled || this.callActive || digit !== '0') return;
+
+		this._clearLongPressTimer();
+		const target = event.currentTarget as HTMLElement | null;
+
+		this._longPressTimeoutId = window.setTimeout(() => {
+			this._suppressNextClick = true;
```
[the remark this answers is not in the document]
SpawnAtis: I've added a fallback in `_endLongPress()` that resets the flag after a delay if it's still set.  
If click fires normally,  it resets the flag first (no issue). If click doesn't fire,  the timeout resets it as a fallback

PR #2595, opened 2025-11-12
chore(docs): audio player vue examples (VIV-2909) 

TaylorJ76:


konradsienkowski (changes requested):
It looks like this PR is reverting October 2025 updates - I don't think this is intentional. The rest looks good!

TaylorJ76:
> It looks like this PR is reverting October 2025 updates - I don't think this is intentional. The rest looks good!

Not sure what went on there. Fixed now.

konradsienkowski (approved):

PR #2612, opened 2025-11-24
docs(design-tokens): add design-tokens docs (VIV-2859)

jhajduk-vonage:


apps/docs/content/design-tokens/what-design-tokens-are.md
```diff
@@ -0,0 +1,53 @@
+---
+title: What Design Tokens Are
+order: 10
+---
+
+# What Design Tokens Are
+
+Design tokens are the shared language of visual decisions in your design system. They capture choices such as colors, typography, spacing, elevation, and radii in a format that is consistent, portable, and easy to apply across tools and platforms. Instead of storing raw values like `#0077ff` or `16px`, tokens store the _meaning_ behind those values \- so teams can work with intent rather than memorizing numbers.
```
RichardHelm:
```suggestion
Design tokens are the shared language of visual decisions in our design system. They capture choices such as colors, typography, spacing, elevation, and radii in a format that is consistent, portable, and easy to apply across tools and platforms. Instead of storing raw values like `#0077ff` or `16px`, tokens store the _meaning_ behind those values \- so teams can work with intent rather than memorizing numbers.
```

apps/docs/content/design-tokens/what-design-tokens-are.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Core Vocabulary
+
+A few terms help anchor how tokens work:
+
+- **Token** - the conceptual design decision, described by name and meaning.
+- **Variable** - the technical form of a token in platforms such as CSS, Flutter or Figma.
+- **Theme** - a collection of token values that define a specific visual context (e.g., light mode, condensed spacing).
+- **Consumer** - any UI element or component that uses tokens.
+- **User** - the person selecting tokens in design or code.
```
RichardHelm: While it doesn't belong in the vocabulary, I think it would be useful to clarify specifics about our system. E.g. who creates themes or whether people people should override variables. Currently people do that to theme applications for Cpaas

apps/docs/content/design-tokens/reference/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Token Types (Reference Pages)
+
+Design tokens cover several dimensions of the interface - color, typography, spacing, shape, size, and elevation. Each dimension has its own purpose and its own set of semantic scales. To keep things simple and predictable, each token category is documented on a dedicated page.
+
+## Categories
+
+### Colors
+
+Covers semantic color families, usage guidance for backgrounds, solids, borders, and text, and the full neutral, accent, status, and accent-alt scales.
```
RichardHelm: The accent color in Vivid is neutral so it's not clear to me what's going on here. Is it changing? Or are we redefining what accent means?

jhajduk-vonage: Yes, it's going to change. With new tokens, the "accent" is visually accented which would refer to visual hierarchy.

apps/docs/content/design-tokens/reference/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+[Colors](/design-tokens/reference/colors/)
```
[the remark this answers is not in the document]
TaylorJ76: I'm happy with `vivid-` too :)

PR #2618, opened 2025-12-01
fix(icon): icons loading twice (VIV-2901)

TaylorJ76:


libs/components/src/lib/fab/fab.a11y.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('a11y: vwc-fab', () => {
 		const icon = 'home-line';
 		element.icon = icon;
 		element.iconTrailing = true;
+		element.setAttribute('aria-label', 'Home');
```
RichardHelm: Don't think this should be here
```suggestion
```

RichardHelm: Ok I see now

PR #2689, opened 2026-01-14
feat(popover): new component (VIV-3037)

konradsienkowski:


libs/components/src/lib/popover/popover.ts
```diff
@@ -0,0 +1,367 @@
...
+ * @slot anchor - Slot for the trigger element.
+ * @slot footer - Use the footer slot in order to add action buttons or other contents to the bottom of the dialog.
+ * @event {CustomEvent} open - Fired when the popover opens.
+ * @event {CustomEvent} close - Fired when the popover closes.
+ */
+export class Popover extends Localized(DelegatesAria(VividElement)) {
+	/**
+	 * @internal
+	 */
+	popoverEl!: HTMLElement;
```
jhajduk-vonage: I am not sure about the notation at this point, but I believe that internal properties suppose to be annotated with `private` and one of `_` and `#` ?

jhajduk-vonage (approved):

libs/components/src/lib/popover/popover.ts
```diff
@@ -0,0 +1,385 @@
...
+import { DelegatesAria } from '../../shared/aria/delegates-aria';
+
+/**
+ * @public
+ * @component popover
+ * @slot - Default slot for the popover content.
+ * @slot anchor - Slot for the trigger element.
+ * @slot footer - Use the footer slot in order to add action buttons or other contents to the bottom of the dialog.
+ * @event {CustomEvent} open - Fired when the popover opens.
+ * @event {CustomEvent} close - Fired when the popover closes.
```
RichardHelm: I don't think these are events are fired

apps/docs/content/_data/components.json
```diff
@@ -599,5 +599,14 @@
 		"code": "./libs/components/src/lib/contextual-help/README.md",
 		"hideGuidelines": "true",
 		"hideAccessibility": "true"
+	},
+	{
+		"title": "Popover",
```
RichardHelm: The component should be in alpha

PR #2691, opened 2026-01-15
chore(data-table): implement basic components (VIV-3070) 

SpawnAtis:


TaylorJ76 (changes requested):
I think we need to consider how a `table-header-cell` (which is the equivalent to a `th` html  element) responds to being used as **row heading cell** inside a body row.
[See this example](https://www.w3.org/WAI/tutorials/tables/one-header/#table-with-header-cells-in-the-first-column-only)

Currently it has the border color of the a header cell. 
I think it should also set the `scope` attribute to either `row` or column`.

Either the component should be clever enough to know its own scope **(preferred)**,
or we need 2 separate components

TaylorJ76 (changes requested):

PR #2722, opened 2026-01-29
fix(tag): prevent visible focus on remove button (VIV-3049)

konradsienkowski:


libs/components/src/lib/tag/tag.template.ts
```diff
@@ -66,6 +66,7 @@ export const tagTemplate = (context: VividElementDefinitionContext) => {
 					ariaPressed: (x) => x.selected,
 				})}
 				?disabled="${(x) => x.disabled}"
+				@mousedown="${(x, c) => c.event.preventDefault()}"
```
RichardHelm: Well, this prevents the focus from being visible by preventing the focus in the first place. But that's not right, e.g. when clicking on a regular button the button is focused but has no visible focus ring

RichardHelm (changes requested):

libs/components/src/lib/tag/tag.template.ts
konradsienkowski: You're right, I have overengineered it. Replaced it with a simple CSS fix that should do the work

RichardHelm (approved):

PR #2723, opened 2026-01-30
chore(table): build initial set of reusable partials (VIV-3057)

SpawnAtis:


libs/components/src/lib/country-indicator/countries-data.ts
jhajduk-vonage: This is huge file with 4200+ lines, that contains data we are mostly not using. I don't think this is needed.

libs/components/src/lib/country-indicator/ACCESSIBILITY.md
```diff
@@ -0,0 +1,5 @@
+## Implementation
+
+- The component displays the **country code** (e.g. "UK", "US") as the visible text when **country-code** is set; that text is available to assistive technologies.
+- The default flag (from flag-icons) is **decorative** and has `aria-hidden="true"` so it is not announced.
+- If you use the **icon** slot with a custom image: use `alt=""` when the flag is decorative (e.g. when the country code or context already conveys the country). Use a short `alt` (e.g. "UK flag") when the image conveys information that is not available in the text.
```
jhajduk-vonage: I don't think flag needs ever to be announced. Country code or name should be totally enough.

libs/components/src/lib/country-indicator/country-indicator.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class CountryIndicator extends VividElement {
+	/**
+	 * ISO 3166-1 alpha-2 country code (e.g. "GB", "UK", "US"). When set, the component shows the matching flag icon and the country code (e.g. "UK") automatically.
+	 *
+	 * @public
+	 * @remarks
+	 * HTML Attribute: country-code
+	 */
+	@attr({ attribute: 'country-code' }) countryCode?: string;
```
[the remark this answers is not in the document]
jhajduk-vonage: Also the type could be narrowed to available options so that it provides autocomplete.

TaylorJ76 (changes requested):
As stated in the JIRA task description and acceptance criteria, partial docs should be just one page (not split into Usage, code, a11y etc) and should be grouped under a section titled "Partials".

libs/components/src/lib/country-indicator/countries-data.ts
SpawnAtis: I added a comment noting that this was copied from a country-related library, so if anything changes in the future (e.g., the ISO 3166-1 alpha-2 standard), it will be easy to verify and update.

PR #2766, opened 2026-02-16
feat(select): add clear button (VIV-3111)

konradsienkowski:


TaylorJ76 (changes requested):
On a screenreader (at least on VoiceOver), when I focus on the select when the is currently a selected value, the clear button's label is announced, eg. "Title - Miss - Clear selection". The "clear selection" should only be announced when focus is moved to the Clear button.

TaylorJ76 (approved):

PR #2768, opened 2026-02-19
chore(icons): release icons from vivid monorepo (VIV-3103)

jhajduk-vonage:


.github/workflows/release-please.yml
```diff
@@ -109,6 +110,22 @@ jobs:
...
+  vivid-icons-publish:
+    needs: [should-release, release-please]
+    if: ${{ needs.should-release.outputs.should_release == 'true' }}
+    uses: ./.github/actions/upload-icons
+    with:
+      access-key: ${{ secrets.ICONS_S3_ACCESS_KEY }}
+      secret-access-key: ${{ secrets.ICONS_S3_SECRET_ACCESS_KEY }}
+      bucket: ${{ vars.ICONS_BUCKET }}
+      base-folder: ${{ vars.ICONS_BASE_FOLDER }}
+      version: ${{ needs.release-please.outputs.vivid_components_version }}
```
RichardHelm: Unless we want to change how icons are versioned, you will need to add them to release-please-config and use their version here

jhajduk-vonage: Yeah, you're right. Will do it this way.

release-please-config.json
```diff
@@ -26,7 +38,15 @@
 		{
 			"type": "linked-versions",
 			"groupName": "vivid",
-			"components": ["vivid", "vivid-design-tokens", "vivid-vue", "test-utils"]
+			"components": [
+				"vivid",
+				"vivid-design-tokens",
+				"vivid-icons",
+				"vivid-icons-flags",
+				"vivid-icons-brands",
```
RichardHelm: This will sync their version with vivid. You probably want to make another group that syncs the version of the three icon packages instead

PR #2770, opened 2026-02-20
feat(rich-text-editor):  add Shift + Enter  hotkey to insert a new block (VIV-3110)

SpawnAtis:


libs/components/src/lib/rich-text-editor/rte/features/internal/core.ts
```diff
@@ -102,26 +123,8 @@ export class RteCoreImpl extends RteFeatureImpl {
...
-								type: defaultTextblockForMatch(
-									$from
-										.node($from.depth - 1)
-										.contentMatchAt($from.indexAfter($from.depth - 1))
-								)!,
-								attrs: rte.textblockAttrs.extractFromNode(node),
-							};
-						})
-					),
+					Enter: enterKeyChainCommands,
```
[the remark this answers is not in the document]
SpawnAtis: @RichardHelm  I think Enter should always insert a new block, except in very specific cases like a chat app. In that sense, Shift+Enter should remain just an optional shortcut.

At the same time, customers should have an API that allows them to override keyboard shortcuts in the RTE editor. For example, they could override shortcuts through the config when registering the `RteBase() `plugin:
```
RteBase({
  addKeyboardShortcuts() {
    return {
      Enter: () => true, // Prevents default behavior
    };
  },
})
```
 or alternatively, create a small custom plugin to handle their own shortcut behavior.
```
import { RteBase } from '@vivid/rte-vivid';

// Example: Disable Enter key completely
const OverrideEnter = RteBase.create({
  addKeyboardShortcuts() {
    return {
      Enter: () => true, // Prevents default behavior
    };
  },
});
```
But if u want to go with `keyboardShortcuts` prop, I can change it.  What do u think ?

libs/components/src/lib/rich-text-editor/rte/features/keyboard-shortcuts.ts
```diff
@@ -0,0 +1,55 @@
...
+	type PluginContribution,
+	RteFeatureImpl,
+} from '../feature';
+import type { RteInstanceImpl } from '../instance';
+
+/**
+ * A keyboard shortcut handler. Return `true` to consume the key (prevent default behavior), `false` to let other handlers run.
+ * You can use a no-arg function to only prevent default: `() => true`.
+ */
+export type KeyboardShortcutHandler = Command | (() => boolean);
```
RichardHelm: I think we should pass an RteInstance since that it how we want consumers to interact with the RTE instead of the raw ProseMirror API

```suggestion
export type KeyboardShortcutHandler = (instance: RteInstance) => boolean;
```

libs/components/src/lib/rich-text-editor/README.md
```diff
@@ -1114,6 +1122,103 @@ const instance = config.instantiateEditor({
...
+**Configuration options:**
+
+- `id: string`(required):
+  Unique identifier for this feature instance.
+- `options.shortcuts` (record):</br>
+  Map of key name to `KeyboardShortcutHandler`. Keys use ProseMirror key names (e.g. `"Enter"`, `"Shift-Enter"`, `"Mod-b"`).
+
+`KeyboardShortcutHandler`:
+
+- A ProseMirror `Command` `(state, dispatch) => boolean`, or a no-arg function `() => true` to only prevent default. Return `true` to consume the key (prevent default); `false` to let other features handle it.
```
RichardHelm: There is no difference, you can always prove a function with fewer args in TS.
```suggestion
- A ProseMirror `Command` `(state, dispatch) => boolean`. Return `true` to consume the key (prevent default); `false` to let other features handle it.
```

PR #2818, opened 2026-03-18
chore(repo): migrate to changesets (VIV-000)

RichardHelm:


scripts/workflow-tests/stubs/github-api.ts
```diff
@@ -39,46 +33,30 @@ function hashInt(value: string): number {
...
 function save(state: any, key: string, rs: any) {
 	state.repos[key] = rs;
 	writeState(state);
 }
 
+function refUrl(key: string, ref: string): string {
+	return `https://api.github.com/repos/${key}/git/${ref}`;
+}
+
 Bun.serve({
```
jhajduk-vonage: I'm not a fan of introducing another runtime to the stack. I think this would be equally simple to build these stubs using for example hono so we could stay with node.

jhajduk-vonage (dismissed):

PR #2844, opened 2026-04-08
feat: add keyboard shortcuts prototype (VIV-3210)

RichardHelm:


libs/components/src/lib/kbd-key/manifest.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - shortcut
+  - keybinding
+  - kbd
+documentation:
+  code: ./README.md
+  accessibility: ./ACCESSIBILITY.md
+  useCases: ./USE-CASES.md
+  variations: ./VARIATIONS.md
+links:
+  figma: https://figma.com/...
```
jhajduk-vonage: Real working links should be provided here.

PR #2858, opened 2026-04-15
feat(country-group): add country-group component (APIVIV-11) 

SpawnAtis:


jhajduk-vonage:
I'm not sure how to address these changes since they contain things that we explicitly agreed not to include to.

However I wanted to ask, whether you considered `IntersectionObserver` instead `ResizeObserver` ? It's purpose is to observe whether elements are in the boundaries of the other elements. I feel like this would be more efficient way of determining which countries should be placed in popover.

Here's the PoC I've prepared: https://stackblitz.com/edit/vitejs-vite-hngnedct

libs/components/src/lib/country-group/country-group.ts
```diff
@@ -0,0 +1,379 @@
...
+		this.#updateAriaLabel();
+		this.#requestFit();
+	};
+
+	#ensureObserver(): void {
+		if (this.#observer) {
+			return;
+		}
+
+		if (typeof IntersectionObserver === 'undefined') {
```
jhajduk-vonage: Why would `IntersectionObserver` would be undefined? It is available in latest two versions of all modern browsers.

jhajduk-vonage (changes requested):

libs/components/src/lib/country-group/manifest.yaml
```diff
@@ -0,0 +1,23 @@
...
+documentation:
+  legacy: ./README.md
+links:
+  figma: https://figma.com/...
+elements:
+  - tag: country-group
+    design: https://figma.com/...
+  - tag: country
+    design: https://figma.com/...
+  - tag: badge
```
jhajduk-vonage: Badge as well as popup are internal dependencies, not elements facing user. They don't need to be a part of public API.

jhajduk-vonage (changes requested):

SpawnAtis:
@jhajduk-vonage Sorry, the mock for `IntersectionObserver ` not `ResizeObserver`

jhajduk-vonage:
Look, here's quick and dirty example of what I mean: https://github.com/Vonage/vivid-3/blob/APIVIV-11-coutry-group-component-jakubs-example/libs/components/src/lib/country-group/country-group.ts

There's no sentinels, queues or microtasks. Everything is handled by `IntersectionObserver` without additional heavy computations, it recalculates visible elements on resize thanks to default IntersectionObserver behavior. 

And it takes 84 lines (without comments) comparing to yours 348 lines. And I'm pretty sure it could optimized even more.

jhajduk-vonage (approved):

PR #2865, opened 2026-04-16
docs(general): fix missing components in the navigation (APIVIV-43)

jhajduk-vonage:


konradsienkowski (changes requested):
It fixes missing components in the navigation, but also exposes there **underlying** components that shouldn't be included there - like Popup, Visually Hidden etc.

jhajduk-vonage:
Right, good catch.

konradsienkowski (approved):

